PR #69, opened 2026-06-06
Spec: paragraph interruption is the default; rebrand post-Djot to post-Markdown

dereuromark:
## What

Make Markdown-style paragraph interruption the always-on default and reposition Carve as **post-Markdown** rather than post-Djot. Carve now reads as "the best of Markdown and Djot merged": Markdown's familiar block behavior on Djot's linear-time, unambiguous parser core.

## Paragraph interruption (section 10)

A **visible** block (heading, thematic break, block quote, bullet/task list, valid table row, fenced code or `:::` with a closer ahead) interrupts an open paragraph with no blank line before it - the CommonMark rule. This is the key block-level divergence from Djot, where an open paragraph runs until a blank line (verified against the reference `djot` binary). Ordered lists stay on the blank-line rule (an ordered marker is too common in prose).

## Rebrand

- `post-Djot` -> `post-Markdown` across the hero, VitePress config metas/copyright, package.json, og-image, the demo, and the grammar header. Djot heritage stays credited in the prose; only the positioning label changes.
- Home-page feature card `Post-Djot Rigor` -> `Linear-Time Rigor` (the claim that survives unchanged).

## Honesty pass on the parser contract

The fence/`:::` **closer lookahead** is the one block rule that consults later lines, so it is not lookahead-free. `technical-rationale.md` now documents it as a bounded forward scan memoized in the block pre-pass (still O(n), still no backtracking) and explains the usability payoff: a stray fence in prose cannot swallow the rest of the document.

## Correctness audit (vs the real `djot` binary and the conformance corpus)

- Grammar section 10 wrongly listed `+ ` as an interrupting bullet; `+` is the continuation marker, not a Carve bullet. Removed.
- Reframed section 10 / edge-cases section 17 as a divergence **from Djot** instead of from Carve's own past; dropped change-over-time wording.
- `divergence-from-djot.md` was silent on the biggest block divergence; added the paragraph-interruption section and a porting note.
- `comparison.md`: added a paragraph-interruption row (Markdown/MDX/Carve yes, Djot no).

Merged current `main` (compact lists, continuation marker, fence label, lowercase heading ids); corpus regenerated from `examples.md`. Conformance: 211 pass, 0 fail.


dereuromark:
While this would be handy for some cases, it also creates quite a few issues.
We might want to rather go with https://github.com/markup-carve/carve/pull/74 instead.

PR #76, opened 2026-06-07
Spec: a bare list marker is paragraph text, not a list

dereuromark:
Clarifies an under-documented list rule (impls already agree).

## Rule
A bullet or ordered marker is a list item only when followed by a space:
- a **bare** marker line (`-`/`*`/`1.` with no space, no content) is **paragraph text**, not a list;
- a marker + space + empty content (`- `) is a valid **empty item**.

Carve is stricter than CommonMark, which treats a bare `-` line as an empty item. Requiring the space keeps a lone dash (a common prose dash / placeholder) from silently becoming a list.

## Changes
- Grammar: NORMATIVE note on `unordered_item` / marker-requires-space.
- Examples + corpus: pinned `05-lists-2` (bare `-` -> paragraph). Both impls already produce this byte-identically; no parser change.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -437,6 +437,22 @@ Span content is parsed recursively, and an inline link still wins over a span.
 
 :::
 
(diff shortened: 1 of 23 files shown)


dereuromark:
Superseded by the combined fence-singletoken-empty-items PR: the empty-item rule documented here relied on a trailing space (bug); reworking as rule A2 (content-less marker is not a list) plus the fence single-token/attributes cleanup.

PR #78, opened 2026-06-07
Remove the nested-list short-form carve-out (require a blank line to nest)

dereuromark:
**Draft for discussion** (raised on carve#74). Removes carve's last paragraph-interruption special case — the list-marker carve-out that let an indented sublist nest with **no blank line**.

## Change

Before, carve uniquely allowed `- a` + newline + `  - b` to nest a sublist with no blank line (its sole deviation from djot). After, the rule is uniform: **no** visible block — list marker included — interrupts a paragraph without a blank line.

- `- a` + newline + `  - b` → one item, text `a\n- b` (lazy continuation), matching djot.
- A nested list needs a blank line: `- a` + blank + `  - b`.

## Why now (in light of #74)

Before #74 the carve-out was *needed*: the blank-line form loosened the list, so the no-blank shortcut was the only way to get a tight nested sublist. With #74 (compact list blocks) the blank-line form stays tight, so the carve-out's reason-for-being is gone. Removing it restores one uniform rule (every sub-block needs a blank line) and aligns the no-blank case with djot.

**Cost:** `- a` + newline + `  - b` (no blank) now renders flat instead of nested — diverges from Markdown muscle memory. That's the deliberate trade-off this PR is for deciding.

## Blast radius (small)

- `grammar.ebnf` §10: carve-out replaced with the uniform no-exception rule.
- `examples.md`: 3 nesting examples switched to the blank-line form (**HTML unchanged** — the blank-line form already renders tight in the vendored lib); 2 examples that documented the carve-out now show the flat output + explain the rule.
- 5 corpus pairs updated; `npm test` 206/0.

## Lockstep PRs

- Reference impl: markup-carve/carve-js#84 (one-line guard removal + unit tests).
- PHP impl: markup-carve/carve-php#71 (strict-mode list-parser change; the opt-in `blocksInterruptParagraphs` mode keeps no-blank nesting by design).
- carve-js has no interrupt mode, so its removal is unconditional.

## Vendored lib note

`carve-lib` is **surgically patched** (parse module only) to isolate this from unrelated carve-js↔carve drift (a full re-vendor pulls in frontmatter/heading-id changes). It also carries the carve-js#86 lazy-continuation dedent fix as a surgical patch. Re-vendor from merged carve-js main (with #84 + #86) on landing.

## Resolved

The carve-js lazy-continuation dedent quirk (a stray leading space at odd indents, where djot + carve-php dedent fully) was fixed in markup-carve/carve-js#86 (merged). The flat-continuation example uses the original 3-space input again; all three impls + djot agree on the full-dedent output `a\n- b`.

Do not merge before the impl PRs; lockstep.


dereuromark:
Closing: superseded by the interrupt-default / post-Markdown direction (carve#69). That makes block openers interrupt with no blank line and keeps the no-blank-line sublist nesting, the opposite of requiring a blank line to nest.

PR #116, opened 2026-06-12
Spec: ::: line-block (verse) with preserved whitespace and hard breaks

dereuromark:
## What

Adds the `::: line-block` (verse) block: a container that preserves the author's line layout.

```carve
::: line-block
Roses are red,
Violets are blue.
:::
```
```html
<div class="line-block">
  <p>Roses are red,<br>
Violets are blue.</p>
</div>
```

- each soft line break becomes a hard break (`<br>`)
- a blank line starts a new stanza (`<p>`)
- per-line leading whitespace is preserved (each leading space serializes as `&nbsp;`)
- renders as a generic `<div class="line-block">`
- the `:::` trigger avoids the pipe/table ambiguity of the Pandoc `|` form

## Contents

- `resources/grammar.ebnf`: the normative line-block production
- `docs/examples.md`: the **Line blocks** section (generates the corpus pairs)
- `tests/corpus/88-line-blocks*`: 6 conformance pairs (basic, indentation, stanzas, hard breaks)
- `tests/spec/88-line-blocks.test`: the mirrored fenced-example form
- re-vendored `docs/.vitepress/carve-lib` (synced from the merged reference impl)

## Lockstep

The reference implementation landed first in carve-js (#136); this is the matching spec + corpus pin. `npm run test:conformance` is green (265 pass, 0 fail) and `corpus:build` produces no diff. Merges cleanly with `main`.


dereuromark:
Closing: the line-block (verse) feature is already on main via #109 (same 88-line-blocks corpus pairs, grammar, and examples section). This branch is a stale parallel copy, so the merge-base diff is +1/-60 — it would revert the newer 84-block-attribute-lines-6/7 corpus pairs and re-add nothing. Opened in error.

PR #117, opened 2026-06-12
Spec: pin the typed colon-fence opener trailing attribute (resolve divergence)

dereuromark:
## What

Pins the last open attribute divergence: a trailing attribute block on a **typed** colon-fence opener.

```carve
::: warning {#w data-k=v}
Mind the gap.
:::
```
```html
<aside class="admonition warning" id="w" data-k="v">
  <p>Mind the gap.</p>
</aside>
```

The block applies to the admonition / div wrapper: extra classes accumulate after the type class (`::: note {.x}` -> `class="admonition note x"`), an id or key-value attaches to the element, the block may abut the type or quoted title, and a preceding block-attribute line merges (leading classes first, the opener winning on id/key conflict). Grammar `admonition_open` already carries the `[attributes]` slot.

## Changes

- `docs/examples.md`: two new Admonitions examples (trailing id/key-value, and the leading-block-attribute merge) -> corpus pairs `13-admonitions-6/7`
- `MAINTAINING.md`: moves the row from **Open (tracked)** to **Resolved**

## Resolution

carve-php already applied this; carve-js gained it in #135 (merged). Both impls now produce identical output, and the vendored carve-lib applies it, so the pin is green: `npm run test:conformance` passes (269 pass, 0 fail) and `corpus:build` is clean. Both new pairs are byte-identical on carve-php too.


dereuromark:
Closing: this pins a divergence from strict djot. Canonical djot treats a trailing attribute block on a ::: fence opener (`::: note {.x}`) as a paragraph, not an attributed div/admonition — attributes attach only via a preceding {...} line. Per the strict-djot intent for the corpus, this should not be pinned.

PR #130, opened 2026-06-12
MAINTAINING: track two carve-rs gaps found by the registry sweep

dereuromark:
The carve-rs registry sweep (carve-rs#41) ran every corpus pair against the parser and registered all 60 additionally-passing families. It also surfaced two genuine non-conformances with pinned canonical behavior, now recorded as Open rows:

- **Arrow vs highlight opener**: `d => e; x != y` in emphasis-opening position - carve-rs opens a `<mark>` span between the two `=`s; spec PART 8 / PART 9 §8 say the smart-typography pattern consumes the delimiter (`⇒` / `≠`). js + php conform. *(38-smart-typography-arrows-and-symbols)*
- **Line blocks**: carve-rs does not implement the `::: |` container; the opener parses as a paragraph. js + php conform. *(88-line-blocks)*

Both families stay unregistered in carve-rs until fixed, so its suite remains green while the rows track the work.

### DIFF
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -99,7 +99,10 @@ Decided canonical behavior, pinned in the corpus, where at least one
 implementation still diverges (lockstep order: carve-js first, then the carve
 pin; the corpus has no xfail). A row moves to *Resolved* once all impls agree.
 
-_None currently._
+| Input | Canonical | Impl(s) to change | Pin |
+|-------|-----------|-------------------|-----|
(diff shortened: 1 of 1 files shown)


dereuromark:
Superseded: both tracked carve-rs gaps are fixed in code (carve-rs PR for line blocks + => arrow precedence) rather than left as tracked divergences, so these MAINTAINING rows aren't needed.

PR #156, opened 2026-06-17
Block separation: nothing interrupts a paragraph (one rule, fully Djot-aligned)

dereuromark:
# Block separation: nothing interrupts a paragraph (one rule, fully Djot-aligned)

Make carve's paragraph boundary a **single rule with zero carve-outs**: an open paragraph is interrupted by **nothing**. Every block, every list marker, and every invisible construct folds into the paragraph as lazy continuation when it follows a prose line with no blank line; a **blank line is required to start any block after prose**. This is exactly Djot ("paragraphs can never be interrupted by other block-level elements").

> Spec-only proposal. The `tests/spec` submodule in the impl repos tracks this branch. **Draft:** the corpus conformance job is expected red — no implementation ships full folding yet (the carve-php #178 draft is the impl half). It goes green once an impl adopts it.

## The diff, by construct

After a prose line with **no blank line**:

| input (`prose` ⏎ `X`) | main today | this PR |
|---|---|---|
| `# H` heading | `<p>prose</p>` + `<h1>` | `<p>prose␤# H</p>` |
| `> q` block quote | `<p>prose</p>` + `<blockquote>` | `<p>prose␤&gt; q</p>` |
| `\| a \| b \|` table | `<p>prose</p>` + `<table>` | `<p>prose␤\| a \| b \|</p>` |
| ` ``` ` fenced code | `<p>prose</p>` + `<pre>` | `<p>prose␤<code>…</code></p>` |
| `---` thematic break | `<p>prose</p>` + `<hr>` | `<p>prose␤—…</p>` (folds; smart em-dash) |
| `:::note` admonition | `<p>prose</p>` + `<aside>` | `<p>prose␤:::note…</p>` |
| `- item` / `1.` list | folds *(already, since the symmetric-list change)* | folds *(unchanged)* |
| `[r]: /u` ref def | consumed (registered) | `<p>prose␤[r]: /u</p>` (literal; not registered) |
| `%% c` comment | consumed (dropped) | `<p>prose␤%% c</p>` (literal) |
| `{.x}` attribute line | floats forward | `<p>prose␤{.x}</p>` (literal) |

A **blank line** before any of them starts the block as normal. Net: the only difference from main is that **visible blocks and invisible constructs now fold like list markers already do** — one uniform rule instead of "lists fold, everything else interrupts."

## Why

- **Footgun removal (the real win).** The dangerous failure is the false positive: hard-wrapped prose where a line happens to begin `#`, `>`, `|`, `-` silently becomes structure. Folding makes that impossible; a blank line is the always-available, obvious fix. Interruption has no defense against accidental blocks.
- **One rule, not a pile.** Replaces the per-construct interruption table (which interrupts, which doesn't, lists vs visible vs invisible) with a single sentence. carve's own history — symmetric-list, container-interruption, Family-D — is the cost of litigating "which construct interrupts what"; folding deletes the question.
- **Djot alignment.** carve = "djot minus footguns." Interruption is the inherited footgun; folding *removes* a divergence rather than adding one (see `docs/divergence-from-djot.md` §7, now "no divergence").
- **carve can afford the strictness.** It ships the escape hatches Markdown lacks — the `+` continuation marker (§17) and `::: >` fenced blockquote — so block-after-prose / block-in-quote stays easy without per-line markers. The strictness is opt-in load (`+`/`:::`), not a mandatory footgun.

### Rule count

| model | interruption rules |
|---|---|
| CommonMark / Markdown | ~8–10, irregular (setext, ordered-start-1, indented-code, HTML-block type 7, …) |
| carve main today | ~3 (block-openers interrupt; lists fold; invisible interrupt) |
| **this PR** | **1** (nothing interrupts; blank line before any block) |
| pure Djot | 1 |

## Not §10 interruption (unchanged)

Two things look like exceptions but are separate mechanisms that never act on an open paragraph:
- **Captions** (`^ …`, §4) attach to a preceding *captionable* block; after a plain paragraph they fold.
- **Nested sublists** (§24): an indented marker inside an open list item opens a sublist with no blank line.

And **headings stay bounded**: a heading is not an open paragraph, so a block-opener or list marker still ends it (`# H` ⏎ `- item` is a heading + list). The heading also keeps Djot's same-`#` continuation rule (inherited from main).

## Scope

`resources/grammar.ebnf` (§10 + the lazy-continuation / paragraph notes), `docs/examples.md` (every interruption example flipped to folding), `docs/divergence-from-djot.md` (§7), and the regenerated corpus. Current with main — inherits the blockquote list-marker fold and the same-`#` heading rule, so the diff here is purely the paragraph-folding step.


dereuromark:
This would be cleaner, absolutely.
But usability wise people will not like to use it as such. Closing for now.

PR #165, opened 2026-06-19
ci: validate embedded Mermaid diagrams (incl. .crv) in CI

dereuromark:
Wires a Mermaid-diagram lint into CI so a broken diagram fails the PR instead of silently rendering as an error in the browser (via [mermaid-lint](https://github.com/jasonworden/mermaid-lint)).

Carve renders Mermaid through its Tier-3 extension but treats the diagram body as an opaque code fence - it does not parse Mermaid grammar, so a broken diagram sails through and only fails later in the browser. This lint closes that gap for our docs.

## Native `--ext crv` (no wrapper)

Carve's diagrams also live inside **`.crv`** documents (the playground demo `docs/.vitepress/examples/demo.crv`). The original POC shipped a custom `scripts/lint-mermaid.mjs` because the CLI only read `.md`/`.mdx`/`.markdown`/`.mmd`. mermaid-lint has since shipped native `.crv` / arbitrary-extension discovery (jasonworden/mermaid-lint#35), so the wrapper is gone - CI now calls the CLI directly with `--ext crv`.

## What's included

- `lint:mermaid` npm script: `mermaid-lint --ext crv --quiet`.
- `mermaid-lint/cli` `^0.35.0` devDep (reproducible + npm-cacheable in CI).
- A `mermaid diagrams valid` job in `ci.yml`.

## Verified locally

- Passes on current docs: validates the 2 diagrams in `demo.crv` (a `graph TD` + a `sequenceDiagram`), exit 0.
- Fails correctly on a broken diagram: exit 1 with the parser error and source map.

## Note

The demo's diagrams currently render as a plain code block in the playground (the Mermaid extension isn't enabled there / the page doesn't init mermaid.js), so they aren't shown as live diagrams yet. Making them render live is a separate site change; this PR guards their validity.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,3 +72,28 @@ jobs:
       # Build the VitePress site so a broken doc (e.g. unescaped Vue
       # interpolation) fails the PR instead of only the Pages deploy.
       - run: npm run docs:build
+
+  mermaid-lint:
+    name: mermaid diagrams valid
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 20
+          cache: npm
+
+      - name: Prefer HTTPS for Git dependencies
+        run: |
+          git config --global url."https://github.com/".insteadOf ssh://git@github.com/
+          git config --global url."https://github.com/".insteadOf git@github.com:
+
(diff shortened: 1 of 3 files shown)


jasonworden:
https://github.com/jasonworden/mermaid-lint/issues/35 shipped carve support in `mermaid-lint` and many more features have shipped since! 🧜‍♂️

PR #179, opened 2026-06-20
docs(extensions): catalog FencedRender (Tier-3) + §6 reference

dereuromark:
Documents the generic fenced-render factory (issue #167), now shipped in carve-php (#201), carve-js (#221), and carve-rs (#112). Adds it to the Tier-3 catalog list and a §6 reference section (config, text/json content modes, the `</` script guard, always-on attribute hardening, conformance). Mermaid is documented as a preset of it.


dereuromark:
Superseded: main already documents FencedRender in the Tier-3 catalog (list entry + a full paragraph in the same §1 style as Details/MathBlock), added in parallel. This PR's longer list entry + §6 section would duplicate it.

PR #263, opened 2026-07-14
Finalize the 0.1.0 changelog entry for the first spec release

dereuromark:
The spec repo has never been released: no tag, no published GitHub release, and the `[0.1.0]` changelog heading still carried a `YYYY-MM-DD` placeholder. Since nothing ever shipped from here, the version stays at 0.1.0 and the pending `Unreleased` entry folds into it rather than becoming a follow-up version.

- Date the `[0.1.0]` heading, empty out `Unreleased`.
- Fold in the braced-only superscript/subscript decision: the Tier-1 emphasis bullet no longer advertises bare `^text^` / `,text,`, and sub/sup get their own bullet carrying the rationale.
- Record the symbol work: `@mention` / `#tag` / `:name:` share one left-boundary rule at Tier 1, and the old Tier-2 "emoji glyph map" bullet becomes the general symbol map.

A matching draft release is staged and stays a draft until it is published.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,18 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
-
-- **BREAKING: no bare superscript/subscript delimiters.** `^text^` and
-  `,text,` are now literal text; superscript and subscript are written with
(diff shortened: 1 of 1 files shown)


dereuromark:
Superseded: the same changelog commit was merged as part of #262 (shared-checkout mixup). main already carries the finalized 0.1.0 entry.

PR #291, opened 2026-07-19
spec: file inclusion / transclusion (PART 9 §19, reserved {{ }} directive)

dereuromark:
Authors the normative specification for Carve file inclusion / transclusion, the reserved `{{ … }}` processor directive. **Spec-only:** a conformant core still performs no I/O and leaves directives literal unless a host opts into an include processor.

## Language contract

- bare or quoted paths, optional `#section`, `@lines:N-M`, `@shift:N` and `@shift:auto`
- block includes merge blocks; inline includes accept inline-only content
- directives are recognized across contiguous literal-shaped inline runs and stay inert in verbatim contexts
- a child is parsed as a WHOLE document, not a fragment: its own frontmatter is its own, and it gets the same inline vocabulary the parent has
- footnote labels and explicit heading ids rename deterministically; automatic ids are out of scope; reference definitions stay file-scoped
- footnotes are answered in the ASSEMBLED document, and definitions an include brings in are collected like the document's own
- dependency reporting covers resolved and attempted targets in deterministic order

## Writing Carve back out

Two obligations that are separable, and an engine satisfied one while failing the other:

- **I12** - the writer preserves a directive verbatim rather than escaping it. Escaping renders as the same literal text, which is why no round-trip invariant catches it and why formatting a document would silently delete every include in it.
- **I15** - a processor targeting Carve source does not run the expansion pass at all. Expansion answers "what does this document say"; the Carve target answers "what does this document consist of".

## Resolver and security contract

- paths resolve relative to the including file under one top-level containment root
- canonicalize first, then contain; an in-root `..` is valid, symlink, traversal and absolute escapes are not
- includes are opt-in and off for untrusted input; remote URLs require an explicit allowlist
- cycle, finite-depth, total-byte and resolver-call limits are mandatory, and a spent whole-render total refuses later directives **without resolving them**
- a processor MAY cap retained warnings, and then MUST keep one per distinct condition and report the suppressed count
- failures warn and stay literal; included content receives the parent's sanitization
- **when the medium is not a filesystem** - a database, object store, CMS or in-memory map - the host owes a stable identity per target, containment expressed as resolve-then-test, its own per-target size limit, and treating the target string as untrusted input

## Conformance corpus

105 shared vectors under `tests/include-conformance/`, covering I1-I15. Two goldens exist because the obvious ones cannot see what they pin:

- `carveTarget` - the Carve output produced **with a resolver configured**. `fmt` is computed from the pre-expansion document by construction, so an engine that expands before calling its writer satisfies it and still returns a flattened book.
- `flattened` - the Carve source of the **expanded** document. HTML cannot show the assembled tree's shape, since footnotes are collected globally at render time whatever order the tree holds them in.

99 of the 105 resolve through an in-memory resolver, so the non-filesystem host is the better-tested path; the six filesystem vectors exist for canonical containment, which is the one part a filesystem owns.

The processor-neutral security subset merged separately in #1057 and stays the source of truth for those cases.

## Files

- `resources/grammar.ebnf` and `resources/spec/10-includes.ebnf`: PART 6 syntax, including the quoted path form and whitespace-run padding
- `docs/includes.md`: reader-facing language and host contract
- `docs/security.md`: include threat model and containment rules
- `tests/include-conformance/`: the vectors, schema and coverage manifest
- `scripts/include-conformance-*.mjs`: generator and reference adapter
- corpus category 462: no-resolver literal behavior

Engines: markup-carve/carve-js#356, markup-carve/carve-php#373, markup-carve/carve-rs#241 - all three pin this branch and go green on their spec-pin gate once it lands. Highlighting: markup-carve/tree-sitter-carve#285, markup-carve/carve-grammars#403, markup-carve/pygments-carve#44, markup-carve/rouge-carve#6, markup-carve/sublime-carve#39, markup-carve/emacs-carve#34.


dereuromark:
Tracked in #292 (implementation + open design questions).

dereuromark:
Security conformance has been split out into #1057. When this draft is rebased, keep its broader 94-vector language suite here, but consume the security-only corpus from main rather than re-declaring path, remote, depth, or byte-budget behavior. #1057 deliberately does not settle this PR’s rendering/formatting/selection/collision decisions.

dereuromark:
Rebase needed - conflicting against `main`.

The overlap is real on two files rather than incidental:

- `docs/examples/edge-cases.md` is the corpus source and has taken four new categories since this branch was last touched (2026-08-09): the language attribute (294-297), the boolean `lang` spelling (298), and the semantic registry (299). A rebase must keep this branch's sections **at the end of the file** - inserting above an existing section renumbers every category after it and breaks every engine allowlist, which `scripts/generate-corpus.mjs` refuses outright.
- `docs/index.md` and `docs/native-features-analysis.md` moved with the registry change (markup-carve/carve#1150) and the corpus counter.

After rebasing, run `npm run corpus:build` and check the numbering did not shift: the new inclusion categories should sit at 300+, not renumber 294-299.

PR #306, opened 2026-07-21
Pin the column-exact fenced-code delimiter rule (closes #298)

### LINKED ISSUE #298: Executable spec: fenced code only recognized at column 0, engines accept 0-3 columns
The executable spec (`scripts/spec/layout.mjs`) only recognizes a fenced code block when the opening fence starts at column 0. `FENCE` and `PURE_FENCE` are anchored:

```js
const FENCE = /^(`{3,}|~{3,})(.*)$/
const PURE_FENCE = /^(`{3,}|~{3,})[ \t]*$/
```

All three engines accept an opener indented 0-3 columns. Verified at document level, no list involved:

````carve
 ```
 c
 ```
````

carve-js, carve-php, carve-rs all render:

```html
<pre><code>c
</code></pre>
```

The executable spec instead falls through to a paragraph and renders the backtick runs as inline code spans:

```html
<p><code>
c
</code></p>
```

Same divergence at 2 and 3 columns of indent. At 4 columns the engines themselves stop agreeing (see below), so the safe consensus rule to implement is **0-3 columns**.

By comparison, the other block openers are strictly column-0 in every engine: an indented `# h`, `---` or `> q` renders as a paragraph in carve-js, carve-php and carve-rs alike, at every indent 1-4. So the fence is the one indent-tolerant opener, not a general rule.

## Why it is not a one-line regex change

`FENCE` has 15+ call sites in `layout.mjs`, several of them paragraph-interruption and lazy-continuation guards (`startsVisibleBlock`, the blockquote scanner, the list-item scanner). Making the pattern indent-tolerant globally would also change what interrupts a paragraph and what folds lazily, which is a much larger behavioral surface than the bug. It needs its own ground-truth pass over: whether an indented fence interrupts a paragraph, what indent the CLOSER may carry relative to the opener, how much of the opener's indent is stripped from the content, and the behavior inside blockquotes and list items.

Content stripping at least is clear from the engines: the opener's indent is removed from the body (` c` renders as `c`, not ` c`).

## Adjacent, unresolved: indent 4

At 4 columns the engines split, so this should be settled at the same time:

````carve
    ```
    c
    ```
````

carve-js and carve-rs emit a code block whose content is `c` followed by a literal ` ``` ` line - i.e. the closer is no longer recognized. carve-php emits just `c`. Neither treats the 4-column opener as an indented-code-block-style construct, so this looks like a closer-matching bug rather than a deliberate rule.

## Impact today

Three inputs in the post-blank list-continuation matrix are still spec-vs-all-engines because of this, after the fix in the companion PR: a fence indented past the content column inside a list item (`- one` + blank + 3-space-indented fence, and the 4-space variants). Nothing else in that matrix is a spec-only defect anymore.

Found while fixing the post-blank continuation rule; kept out of that PR deliberately to avoid a half-probed fence rewrite riding along with a narrow, well-understood fix. Related: issue 295 (cross-engine divergences in the same area).

### PR DESCRIPTION
**Pull request by dereuromark**
Pins the column-exact fenced-code delimiter rule now that all three engines implement it (carve-js #361/#362, carve-rs #244, carve-php #377), and resolves the executable-spec side of #298.

## The rule

A fenced-code delimiter - opener **and** closer - sits exactly at its container's content column (column 0 at the top level). Carve has no indented-code-block construct, so leading spaces carry no disambiguating meaning and the fence is strict like every other block opener (an indented heading, thematic break, or block quote is already plain text). The earlier 0-3 space Markdown tolerance is removed.

## Changes

- **Grammar** (`resources/grammar.ebnf`): a NORMATIVE note on `code_fence_open`/`code_fence_close` stating the column-exact rule and its two consequences (an indented document-level run is prose; a closer indented past its opener is code content).
- **Corpus** (`docs/examples/core.md` → regenerated `tests/corpus`, 427 pairs): two pinned examples - a document-level indented run falls back to inline code spans, and an indented `` ``` `` line survives as sample text inside a fence (the payoff of the strict closer).
- **Vendored engine** (`docs/.vitepress/carve-lib`): synced to current carve-js main so the Playground and the corpus conformance test run the strict engine; the newly vendored `plantuml` FencedRender preset is classified EXCLUDED (needs a client library, like d2/graphviz).

## #298

The executable spec's fence regex was already column-0-anchored - it was ahead of the then-lenient engines. With the engines now strict, the exec spec matches, and `core:check` pins it (427/427, 0 defects). This closes #298.

## Known limitation (not pinned)

The reference-definition prepass is line-based and can spuriously resolve a definition written inside a fence nested at a list item's content column. This is shared engine behavior, tracked separately for the def-collection refactor; it is deliberately **not** pinned in the corpus.

### CODE REVIEW

**Comment by dereuromark** [comment]
## Reconsidering closer strictness - proposing we relax the *closer* (keep the opener strict)

I dug into the failure modes of the column-exact rule with a strict build of `main`, and I think the **closer** half is worth softening before this pins. The opener is fine; the closer has a bad, non-local failure mode.

### What the strict rule actually does (empirical)

The closer must match the **opener's** column exactly. Any deviation fails to close and swallows following content (opener at col 3):

| Closer position | Result |
| --- | --- |
| col 3 (matches opener) | closes cleanly |
| col 4 (over-indented) | `` ``` `` becomes content -> fence never closes -> swallows the next block |
| col 1 (under, still in item) | same - content -> runaway |
| col 0 (flush-left) | dedents out of the list, then the `` ``` `` **opens a new top-level fence** -> swallows the rest of the document |

One forgiving part: opener and closer may both sit at a non-canonical column (both col 2 inside a `1.` item) and still work - so the real rule is **"closer must match opener,"** not "hit the content column."

### Ordered vs unordered

The delimiter column is set by marker width: `- ` -> col 2, `1. ` -> col 3, `10. ` -> col 4. So:

- Unordered is uniform (always col 2).
- Ordered shifts with the number's width, including **3 -> 4 as a list crosses 9 -> 10**. It doesn't break (opener==closer is what matters), but the "correct" indent isn't constant down an ordered list, and copy-pasting a fence between a `- ` list and a `1. ` list silently mismatches opener/closer.

### Why the closer specifically

The two delimiters have very different blast radii:

- **Opener** at the wrong column -> renders as prose / inline code. **Local, self-evident.**
- **Closer** at the wrong column -> **runaway** that swallows everything after it. **Non-local**, far from the cause.

And the single most common Markdown habit - the **flush-left closing fence** - is exactly the case CommonMark's 0-3 tolerance forgives and strict Carve turns into a runaway.

### Proposal: opener strict, closer at or below the opener's column closes

- A `` ``` `` run (same char, length >= opener) at a column **<= the opener's column** closes the fence.
- Only an **over-indented** run (> opener column) stays content.

This:

- turns the flush-left-closer runaway into a **clean close** (the common mistake just works, ecosystem-standard),
- keeps over-indented `` ``` `` as sample content (the embed-by-indentation payoff this PR wants),
- and count-nesting (a longer outer fence) still covers deliberate fence-in-fence embedding.

The cost is losing the single uniform "delimiter sits at the container column" rule - the opener keeps it, the closer relaxes to `<=`. I think trading a non-local runaway footgun for ecosystem-standard forgiveness is the right call for a language that sells readability and no-surprises.

Implementing this now: reference engine first, then re-vendor and adjust this PR's grammar note and corpus pins, then the other two engines.

**Comment by dereuromark** [comment]
## Implementation hand-off: relax the closer (`<=` opener column), fold into the fence work

Decision from the analysis above: **opener stays column-exact; the closer relaxes to "a run at or below the opener's column closes; only an over-indented run is content."** This note is the ready-to-apply spec so it can be folded into the in-flight fence branches rather than landed as a competing change.

### Spec side (this PR / `spec/strict-fence-columns`)

**Grammar note** (`resources/grammar.ebnf`, the `COLUMN-EXACT DELIMITERS` block): keep the opener clause; replace the *closer* clause. Proposed wording:

> A fenced-code **opener** sits exactly at its container's content column (column 0 at the top level). The **closer** is more forgiving: a same-char run of at least the opener's length closes the fence when it sits at **or below** the opener's column; a run indented **past** the opener is not a delimiter but code content (which is what lets an indented ``` line appear as sample text inside a fence). The asymmetry is deliberate: a mis-columned opener fails locally (it renders as prose/inline code), but a strict closer's failure is a non-local runaway - a flush-left closing fence inside a list would otherwise swallow the rest of the document. Relaxing the closer to `<=` turns that runaway into a clean close, matching the common Markdown habit.

**Corpus**: the two current pins (`11-fenced-code-16` opener-strict, `-17` over-indent-as-content) stay valid - both are unchanged by this. **Add one** once the engine is relaxed and re-vendored: a list-nested fence closed by a flush-left ``` closes cleanly (the item ends, following blocks are siblings), instead of a runaway. It cannot be added until the engine implements the relaxation, so it waits on the fold below.

### Engine side (fold into `fix/migration-preserve-list-fence-indent` / the container-column work in #364)

The relaxation is **not** in `parseFence` - that closer check is already `leadingWhitespace <= 3` lenient. The strict-in-list behavior comes from **container body collection**: a list item / definition / blockquote / admonition gathers only lines indented `>= contentColumn` into its body, then re-parses. A flush-left ``` (indent `< contentColumn`) is excluded from the body *before the fence ever sees it*, so the fence runs to the item boundary and the leftover ``` becomes a new top-level opener - the runaway.

Fix: make body collection **fence-aware** - while an unclosed fence is open in the accumulated body, pull a following `<= opener column` same-char run into the body as its closer, rather than stopping the container at that dedent. Applies to every container body loop (`parseBlocks` `indent < baseIndent` boundary at parse.ts ~717, and the list-item / def-description / blockquote / admonition collectors).

Test cases (opener at a list item's col 3):

```
closer col 3  -> closes (unchanged)
closer col 0  -> NOW closes cleanly; item ends; following blocks are siblings   (was: runaway)
closer col 1  -> NOW closes                                                      (was: runaway)
closer col 4  -> stays content (over-indented) - unchanged
```

Plus: top-level fences unchanged; count-nesting (longer outer fence) unchanged; the `fmt` serializer needs no change (it already emits opener==closer).

Cross-impl: same change in carve-php and carve-rs; corpus re-pin + re-vendor after carve-js merges, per the usual lockstep.

I'm not pushing to either branch (both are checked out in other worktrees right now) - this is the spec to fold in. Ping me to implement any part once the fence branch is at a point where it won't collide.

**Comment by dereuromark** [comment]
## Addendum: fence opener on the marker line — the content column jumps (and the attribute width suddenly counts)

A sharp case for both the closer rule and the prepass column calc. When the fence opens **on the marker line**, the content column becomes the fence's **visual start position**, so a long attribute pushes it far right:

```
10.{.superlong} ```php
                $x = 1;
                ```
```

`10.` (3) + `{.superlong}` (12) + ` ` (1) -> the fence opens at **col 16**, and col 16 becomes the item's content column. Verified behavior on strict `main`:

| Closer | Strict (today) | Relaxed (`<=` opener) |
| --- | --- | --- |
| col 16 (under the opener) | closes | closes |
| col 4 (marker-based column) | content -> fence never closes | **closes** |
| col 0 (flush-left) | runaway top-level fence | **closes** |

The inconsistency: the **same item** has content column **4** when the fence is on a separate line (attribute ignored) but **16** when the fence opens on the marker line (attribute counted, because the fence's actual start is the reference). Strict then demands the closer hit col 16 exactly - a deep, non-obvious column driven by the attribute length. The relaxed closer rescues it: `col 4 <= 16` and `col 0 <= 16` both close, only an over-indented run stays content.

Add this as a corpus pin alongside the flush-left case once the engine is relaxed.

## Prepass verification (carve#248 / carve-js#364)

Because the standalone-fence content column ignores the attribute (marker width + 1, so `10.{.superlong}` -> col 4), `detect_prepass_list_marker` must compute the column the same way. A **wide** attribute makes any column-counting mismatch obvious:

```
10.{.superlong} x:

    ```
    [ref]: https://evil.example
    ```

[ref] must stay literal.
```

If the prepass counted `{.superlong}` toward the column (col 16) while the block parser uses col 4, this attributed-wide marker would reopen the spurious-def bug #248 fixes. Worth adding as a prepass test (the existing `-{.c}` case has only a 4-column gap; a long attribute exposes a mis-count).

PR #309, opened 2026-07-21
Add chat-platform flavor tables

dereuromark:
Chat platforms - WhatsApp, Slack, Telegram, Discord, Signal - each accept a small, mutually incompatible subset of Markdown-like markup. Pandoc ships no writer for any of them, so this gap is not reachable through the pandoc bridge.

This adds those targets as **data** rather than code, so adding a platform costs a file in every implementation at once instead of a renderer in each. The design was built and proven in `carve-php-chat` first, across seven flavors, before being lifted here.

## Non-normative, separately versioned

These describe other people's formats, which vendors change on their own schedule (Discord added headings and lists in 2023; Telegram added expandable blockquote in Bot API 7.3). Binding them to the spec version would let a vendor's roadmap force a Carve revision, so they carry their own version in `resources/chat-flavors/manifest.json` and implementations opt in.

## What's here

- `resources/chat-flavors/*.json` - seven flavors plus a versioned manifest
- `docs/chat-flavors.md` - the schema, wired into both sidebar sections
- `tests/corpus-chat/` - `.crv` paired with expected output per flavor
- `tests/chat-flavors.test.mjs` - table validation, wired into `npm test`

## Two families

A flavor declares `output`:

- `markup` - formatting lives in the message string as delimiters (WhatsApp, Slack, Telegram `parse_mode`, Discord)
- `ranges` - the body is plain text and formatting travels beside it as offsets (Signal, Telegram `entities`, Slack Block Kit)

Signal is the clearest case of the second: its docs say Markdown "is not supported at this time and is not planned", yet the message still displays as bold, because the style rides alongside the text.

The offset unit is **declared, not assumed**. Telegram counts in UTF-16 code units, so a character outside the BMP counts as two - in the thumbs-up-plus-bold example the bold span starts at 3 in UTF-16, 5 in UTF-8, 2 in codepoints. Measuring wrongly shifts every range after it.

## Validation without a renderer

The spec repo ships no chat implementation, so the runner checks the tables themselves: vocabulary, complete coverage of the normative node vocabulary from `docs/profiles.md`, and that each flavor records a `verified` date. It was checked by deliberately removing an entry and confirming it fails, not just by a green run.

## Sourcing caveat

Only the Slack table is sourced directly from vendor documentation. Telegram, Discord, WhatsApp and Signal were corroborated across secondary sources - Discord's and Signal's official pages both refuse automated fetch. `docs/chat-flavors.md` states this plainly, and `verified` should be read as "last reviewed", not as a guarantee.


dereuromark:
Rebase needed - conflicting against `main`.

What landed since this branch was last touched (2026-08-03) and touches the same files:

- `docs/index.md` and `docs/.vitepress/config.ts` moved several times with the corpus counter and the new pages (`html-import`, the language attribute section).
- `package.json` moved with two engine-pin bumps, most recently markup-carve/carve#1158.

Nothing here conflicts with the semantic-span work *semantically* - chat flavors are a separate surface - so this should be a mechanical `git rebase origin/main`, taking `main`'s side on the counter in `docs/index.md` and re-adding the sidebar entry after the newer ones.

One thing to re-check while rebasing rather than after: the flavor tables describe which constructs each platform re-renders, and `code`/`mark` stopped being semantic-span names in markup-carve/carve#1150. If a table names them, the row needs a look.

PR #311, opened 2026-07-21
docs: document the img SVG fence extension

dereuromark:
Documents the `img` SVG fence extension (carve-js markup-carve/carve-js#366, carve-php markup-carve/carve-php#382).

Adds `docs/svg-images.md` covering:
- the `img` (alias `image`) fence and why `svg`/`xml` stay unclaimed,
- **sandbox by default** (a browser-sandboxed `data:image/svg+xml` ``),
- **host-gated inline** (`allowInline` + an `{inline}` attribute) for themeable `<svg>`,
- the capability opt-ins (`allowStyle`, `allowLinks`, `allowAnimation`, `allowExternalImages`),
- the security boundary (sandbox is strong; inline uses the hand-rolled sanitizer and is trusted-content only).

Also wires it into the sidebar, lists ImgFence in the Tier-3 extensions table, and cross-references it from the raw-HTML section of the security page as the safe way to put SVG on the page.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -170,6 +170,7 @@ export default defineConfig({
               ],
             },
             { text: 'Diagrams & Charts', link: '/diagrams' },
+            { text: 'SVG Images', link: '/svg-images' },
           ],
         },
         {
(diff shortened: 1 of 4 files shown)


dereuromark:
inline-mode-full

and

inline-theming-orange

vs e.g. sandboxed

sandbox-mode-full

Issue #326, opened 2026-07-24
List-looseness margins: outer-item loosening after a sub-list (rs/php bug), heading+lazy three-way split, multiline-block interior-blank looseness


dereuromark:
Follow-up from the same differential sweep as #325. These are list-looseness margins where the engines disagree and the corpus does not pin the case. Unlike a clear single-outlier bug, these need a canonical ruling because they are subtle and (for the first one) point at carve-js being right and carve-rs/carve-php being wrong — the opposite of what "two engines agree" would suggest.

## A. An outer item loosened by its OWN paragraph after a nested sub-list

Baseline (all three agree — a blank before an item's own second paragraph makes the item loose):

```
- a

  c
```
→ `<li><p>a</p><p>c</p></li>` in rust, js, php.

Now put a sub-list between `a` and the blank:

```
- a
  - b

  c
```

`c` is at the OUTER item's content column (2), so it is the outer item's own second paragraph, separated from `a` by a blank — the same shape as the baseline. Expected: the outer item is loose, both `a` and `c` wrapped.

| engine | outer item |
|--------|-----------|
| carve-js | `<li><p>a</p> <ul>…</ul> <p>c</p></li>` — loose, consistent |
| carve-rs | `<li>a <ul>…</ul> <p>c</p></li>` — `a` bare but `c` wrapped |
| carve-php | `<li>a <ul>…</ul> <p>c</p></li>` — `a` bare but `c` wrapped |

carve-rs / carve-php render the SAME `<li>` with one direct paragraph unwrapped (`a`) and another wrapped (`<p>c</p>`), which is neither a valid tight item nor a valid loose one. carve-js is internally consistent and matches the baseline. This looks like a carve-rs + carve-php bug (the intervening sub-list suppresses loosening of the item's first paragraph only), not a carve-js bug.

Recommendation: rule the outer item **loose** (carve-js behavior), fix carve-rs + carve-php, and pin `- a\n  - b\n\n  c`. Note this is the reverse of the initial read where "rs+php agree" looked canonical.

## B. Flush-left lazy text after a heading that ends a nested sub-list

```
- a
  - b
    # N
lazy
```

[...]

Issue #280, opened 2026-07-18
Inline literal text without escaping or code styling: {=literal} vs {literal} bool-attr


dereuromark:
## Motivation (the real driver: linguistics)

`/.../` is standard phonemic-transcription notation. Carve swapped emphasis to `/italic/`, so a linguist writing a transcription gets emphasis instead of literal text:

```
The word "cat" is /kaet/.
```

renders as

```html
<p>The word "cat" is <em>kaet</em>.</p>
```

This is not a stray edge token - it is a whole notational domain (phonemic `/.../`, and other slash-delimited notations). This exact objection was raised upstream on Djot, which is *why* Djot never adopted `/` for emphasis (it uses `_emphasis_`); Carve deliberately swapped to `/italic/`, so the objection lands on us:

> Unfortunately `/.../` for italics would be a disaster for any linguist (like me) using djot because /.../ has a very specific special meaning in linguistics. Nobody will want to type `\/...\/` all the time! I guess `{/.../}` might work though.

(source: jgm/djot#13, comment 1306200727)

Note `a/b/c` already stays literal (intra-word slashes are not emphasis delimiters), but an isolated `/kaet/` does not.

## Goal

A construct that is
- verbatim inside (no need to escape `/ * _ ~ =`) - the ergonomics bar the linguist raises: no per-character `\/.../\`,
- **not** wrapped in `<code>` (it is prose, not code),
- HTML-escaped on output (safe),
- honored by every renderer (HTML / Markdown / ANSI / plain), not dropped.

## Why the obvious candidates fail

- `` `/kaet/`{=html} `` works today (bare text, no `<code>`, no escaping inside) **but** it is raw/unescaped passthrough (`` `a<b>`{=html} `` injects a real `<b>` tag) and is dropped by every non-HTML renderer.
- `` `/kaet/`{=text} `` reuses the raw-format channel but **hijacks a plausible render-target name**. The `{=FORMAT}` model is "raw content for the FORMAT target: unescaped, emitted only by that target's renderer, dropped elsewhere." `text` / `markdown` / `latex` are all plausible targets, so `{=text}` most naturally means "only in the plain-text renderer, dropped in HTML" - the opposite of what we want. Defining it as "escaped + universal + never-drops" silently inverts all three properties of the channel.

Confirmed current behavior (carve-js):

```
{=html}     => X        (survives, RAW/unescaped)
{=latex}    => (empty)  dropped
{=text}     => (empty)  dropped
[...]

PR #340, opened 2026-07-28
spec: pin the heading-lazy-fold and fence-interior-blank list behaviors

dereuromark:
## What

Pins the two list-continuation behaviors just fixed across the engines, and
refreshes the vendored `carve-lib` to current carve-js so the conformance step
runs the merged fixes.

New examples in **List nesting and looseness** (`docs/examples/edge-cases.md`):

1. A heading folds its trailing flush-left plain text as continuation at any
   nesting depth — even when the heading opens on a deeper sub-list item's
   marker line (markup-carve/carve#326, B).
2. A blank line inside a fenced code block is verbatim content, not an interior
   separator, so it does not loosen the list; the sibling after the fence stays
   tight (markup-carve/carve#326, C).

## Verification

`corpus:build` regenerated (only the two new pairs added); `npm test`
(conformance + fixtures + normativity) and `core:check` pass. The vendored-lib
refresh changes no existing corpus output.


dereuromark:
The two examples pass conformance and the impl unit tests, but they parse-but-diff under the formal Core executable-spec gate (`core:check`) — the ohm-based Core reference does not model heading-lazy-fold across list nesting or fence-interior-blank looseness, and a DEFECT can't be allowlisted (only refusals can). Pinning these needs the formal Core reference updated, which is a separate executable-spec change. The behaviors are already guarded by per-impl tests (carve-rs #280, carve-js #393, carve-php merged), so closing this rather than shipping a red gate.

PR #378, opened 2026-07-29
spec: balance parentheses in a link destination

### LINKED ISSUE #377: carve fmt corrupts a link whose URL contains parentheses, in all three engines
Found while closing out the last case in #369. This one is worse than the case that led me to it: `carve fmt` changes what the document renders as, for a very ordinary URL shape.

## Reproduce

```
[wiki][w]

[w]:  https://en.wikipedia.org/wiki/Foo_(bar)
```

```console
$ carve --carve wiki.crv
[wiki](https://en.wikipedia.org/wiki/Foo_(bar))
```

The reference definition becomes an inline destination, and the closing paren of the URL now ends the destination early:

```html
before fmt   <p><a href="https://en.wikipedia.org/wiki/Foo_(bar)">wiki</a></p>
after fmt    <p><a href="https://en.wikipedia.org/wiki/Foo_(bar">wiki</a>)</p>
```

The href is truncated and a stray `)` leaks into the text.

**All three engines produce exactly this**, byte for byte - carve-js, carve-php and carve-rs. So it is a gap in what the canonical writer can express, not an engine defect, and there is no oracle to copy from.

## Why the corpus missed it

`119-scheme-probe-strips-unicode-whitespace` is the only corpus case with a parenthesised URL in a reference definition, and its URL is `javascript:` - which takes the sanitizer path and comes back percent-encoded (`alert%281%29`) rather than truncated. So the corpus shows a harmless spelling difference where a real URL loses data. Every `to_html(fmt(x))` check passes at 498/498 because no case exercises the shape.

## Why it cannot be fixed in the writer alone

There is no angle-bracket destination form to fall back on. `[a](<url>)` is not supported by the parser:

```
[click](<javascript:alert(1)>)   ->   href = "<javascript:alert(1"
```

So an inline destination cannot carry an unescaped paren, and the AST does not record that a link was written as a reference - the writer has no way to know it should stay one.

## Options, roughly in order of size

1. **Percent-encode `(` and `)` in inline destinations.** Smallest, and the mechanism already exists on the sanitizer path. Renders identically (RFC 3986 treats `%28`/`%29` as equivalent), and `to_html(fmt(x)) == to_html(x)` holds again. The href string changes, so `parse(fmt(x)) == parse(x)` still does not - the URL is equivalent but not identical.
2. **Add an angle-bracket destination form** (`[a](<url>)`) to the grammar and the three parsers, and have the writer use it when the destination contains a paren. Preserves the href byte-exactly and closes both invariants. A real grammar addition with value beyond this bug - it is also the standard way Markdown handles this.
3. **Record reference-ness in the AST** so the writer can keep the reference form. Largest, and it constrains the writer in ways PART 11 §6 arguably already rules out.

My recommendation is **2**, with **1** as an immediate mitigation if a release is near: option 1 stops the corruption today and is a few lines per engine, and option 2 is the version that actually round-trips. They are compatible - 1 can land first and be narrowed once 2 exists.

This wants a maintainer decision because option 2 is normative spec surface, and option 1 changes writer output for every URL containing a paren.

dereuromark:
Closes #377.

A `(...)` destination ended at the first `)`, with no balanced-paren rule and no escapes, so a URL containing a parenthesis could not be written at all:

```
[x](https://en.wikipedia.org/wiki/Foo_(bar))

before   <a href=".../Foo_(bar">x</a>)
after    <a href=".../Foo_(bar)">x</a>
```

Wikipedia and MDN produce these constantly, so this was silent corruption of ordinary documents, and the writer reproduced it: formatting such a document truncated the href and leaked the rest into the text.

### Why balancing, and not the angle-bracket form

An angle-bracket destination was implemented first (carve-js#414, carve-rs#305, carve-php#439). It is dropped in favor of balancing, after checking what the reference implementations actually do - measured against djot 0.3.2 and commonmark.js, not recalled:

| input | djot 0.3.2 | CommonMark |
|---|---|---|
| `[a](http://a/b(c))` | `http://a/b(c)` | `http://a/b(c)` |
| `[a](<http://a/b(c)>)` | `<http://a/b(c)>` - brackets land in the URL | `http://a/b(c)` |
| `[a](http://a/b\(c)` | `http://a/b(c` | `http://a/b(c` |
| `[a](x)y)` | `x`, then literal `y)` | `x`, then literal `y)` |

Djot has no angle form at all. Both solve the parenthesis problem by balancing, and both accept a backslash escape for the unbalanced case. Balancing also fixes the spelling people actually type, where the angle form left `[x](.../Foo_(bar))` still broken, and it avoids carrying two destination spellings for one job.

Djot's rule is not adopted wholesale: djot has no inline titles, folding a quoted run into the URL. A Carve destination still ends at whitespace so a title can follow. Balancing is the part that transfers.

### The rule

- Parentheses nest to any depth; the destination ends at the first `)` with no opener left to pair with.
- An escaped parenthesis, and an escaped backslash, are the only escapes - for the unbalanced case.
- A backslash before anything else stays literal, so URLs carrying backslashes are unaffected and need no doubling.
- Whitespace is still percent-encoded, which is the other thing angle brackets would have carried.

### Corpus

Case 107 changes meaning and is renamed accordingly, from pinning the truncation to pinning the balance, with cases added for depth, the unbalanced tail, and the three escapes. The corpus test fails here until the engines land, since it runs against the vendored engine.

Engine PRs follow in carve-js, carve-rs and carve-php; the three angle-form PRs will be closed.


dereuromark:
On the red `spec-corpus conformance` checks: they are the three renamed 107 cases, and they fail because that job runs the vendored `docs/.vitepress/carve-lib`, which is a published carve-js build that does not balance yet. Re-syncing it against markup-carve/carve-js#415 turns all three green locally.

The sync is deliberately not in this PR. Re-vendoring also picks up everything merged into carve-js since the last sync, and that surfaces four unrelated failures - `04-literal-punctuation` and `07-mention-and-tag-literals`, both on the PART 11 section 2 and section 4 byte assertions. I checked those against a carve-js `main` build with none of this branch's changes and they fail there too, so they are pre-existing drift between the committed roundtrip fixtures and the current writer, not fallout from balancing. They deserve their own change rather than riding along here.

Merge order: carve-js#415 first, then re-sync carve-lib, then this.

PR #414, opened 2026-07-31
spec: refresh PART 12 §4's implementation status

dereuromark:
The paragraph said only carve-js records positions and carve-php has none. **carve-php now serializes conformantly** (markup-carve/carve-php#499) — it records positions behind a parse option and enables them whenever it serializes, which is exactly the arrangement the preceding paragraph permits.

The other two are now stated with what is actually missing rather than as a blanket "none", because the cases are not alike:

- **carve-js** records positions but not on every node. The eight types without one are all *reassembled* rather than sliced from the source — a caption built from a marker line plus continuations, a line block whose leading whitespace was expanded, a table cell scanned out of a row and re-joined, and the numbering nodes synthesized during resolution. They are **absent rather than wrong**, which is the half of the rule that matters. Tracked in markup-carve/carve-js#441.
- **carve-rs** has none at all, and why it is a bigger job than carve-php's is worth recording: several inline variants are tuple-shaped (`Text(String)`, `Code(String, Option<Attrs>)`), so there is nowhere to put a position without changing the variant itself. Tracked in markup-carve/carve-rs#333, which also carries the six trap classes carve-js hit — the part that does not look hard in advance.

### DIFF
--- a/resources/grammar.ebnf
+++ b/resources/grammar.ebnf
@@ -3376,6 +3376,37 @@ EOF = (* end of file *) ;
       spec surface for kind names. A document is
       `{"type":"document","children":[...]}`.
 
+      THE ROOT'S OWN FIELDS ARE FIXED -- NORMATIVE. Beyond `type` and
+      `children` the root carries `srcByteLength` always, and carries
+      `frontmatter` and `footnoteDefs` exactly when the document has them. It
+      carries nothing else.
+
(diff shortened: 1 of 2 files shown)


dereuromark:
Superseded by a version rebuilt on current main. The branch predated markup-carve/carve#415, so it conflicted on text neither of us wrote as part of this change.

PR #444, opened 2026-08-01
docs: make the source migration-clean for Carve (migration prep, do not merge yet)

dereuromark:
## Do not merge yet

This is the docs-source half of the CarvePress migration. Merging it before the site cuts over makes the published VitePress site visibly worse — evidence below.

## What it does

Four classes of construct that VitePress accepts and Carve reads differently. With all four fixed, a plain `markdownToCarve` conversion of `docs/` builds under CarvePress with **no manual fix-ups and no lint findings**.

### 1. Admonition titles must be quoted (18 occurrences, 9 files)

`grammar.ebnf` lines 651-657: the opener is a colon fence, a type, an optional quoted header and an optional label, and *"any other text not matching the header/label shape makes the line an ordinary paragraph, not a fence."*

So `::: tip Some Title` silently loses the admonition **and** its title.

### 2. `::: compare no-render` is not a valid opener (17 occurrences)

An admonition kind is a single token, so that line degrades to a paragraph and takes the whole block with it. The Carve spelling is a preceding attribute line:

```
{.no-render}
::: compare
```

Verified in the built output: 281 compare blocks on `examples/edge-cases`, of which exactly 17 render with the live pane suppressed.

### 3. The home page has no usable title

`docs/index.md` keeps its title in `hero.name`, so a Carve build finds neither a frontmatter `title` nor an H1. Added an explicit `title: Carve`, which VitePress also honours.

### 4. Three headings collide on one slug

`dismissed-syntax.md` has three `## Link Syntax: ...` headings that slugify identically. The second and third get auto-suffixed ids, and any reference to the base slug silently resolves to the first. The later two now carry explicit ids.

`lintCarve` over all 40 converted pages reported exactly one issue — this one.

## Why it cannot merge yet

VitePress renders container info strings verbatim and does not understand attribute lines. Built the site to confirm the first case:

before
```
custom-block-title">One thing you must still configure<
```

after
```
custom-block-title">&quot;One thing you must still configure&quot;<
```

Items 2 and 4 regress similarly: `{.no-render}` and `{#id}` render as literal text under VitePress, and `no-render` stops reaching the custom container that currently honours it. The two syntaxes have no overlapping form, so this has to land together with the switch away from VitePress.

## Verification

- `npm test` — 623 pass.
- `npm run docs:build` — succeeds; the only output differences are the ones described above.
- Full CarvePress build from this branch: **40 pages in 2.8 s**, 638 outline entries, none pointing at a missing anchor, 177 highlighted code blocks, 96 admonitions.

`playground.md` is excluded from the migration: it embeds Vue components, and CarvePress has no component layer.


dereuromark:
https://github.com/markup-carve/carve/pull/1026

Issue #427, opened 2026-08-01
Make section wrappers optional


Brixy:
### Tier

Tier-1 (core language)

### Problem or motivation

Wow. Carve is a game changer! Thank you and good luck.

---

Currently, like djot, a heading and its following content live in a `<section />` wrapper.
The resulting HTML is semantically correct, of course. But the HTML syntax change Markdown -> Carve may cause great problems for existing web projects.

## An example using the `stack` class

[Rationale for the stack (or flow) class](https://every-layout.dev/layouts/stack/)

```
<head>
	<style>
		* {
			margin: 0;
		}
		.stack > * + * {
			margin-block-start: var(--stack-space, 1em);
		}
	</style>
</head>
<body>
	<h1>Page Title</h1>
	<main class="stack">
		<!-- Content here -->
	</main>
</body>
```

Markdown (+ front matter e.g. using an SSG) will work, Carve won't: no margins will be rendered.

```
---
title: Page Title
---

## Page Heading

A paragraph.

Another paragraph.
```

---

This is just a very basic example. But forcing `<section />` wrappers _may_ break complete web projects and make it impossible to switch from Markdown to Carve.

### Proposed syntax / behavior

Make `<section />` wrappers optional.

When optional,

```
{#install .featured}
## Setup
```

should render as

```
<h2 id="install" class="featured">
```

### Alternatives considered

Djot allows filters like this:

```
function removeSections() {
	return {
		section: (el) => el.children,
	};
}

export function djotToHtml(content) {
	const contentDjot = djot.parse(content);
	djot.applyFilter(contentDjot, removeSections);
	return djot.renderHTML(contentDjot);
}
```

dereuromark:
Hah, thx :) Cool that you are looking into this.

I had to look up why Djot did it this way (sections by default) - and in that process I also found your old ticket there with that topic.

Short answer: yes. Drafted in #430.

## Why djot does it

Djot never documented the behavior. `doc/syntax.md` says nothing about section wrapping; it only mentions that identifiers are added automatically. The behavior first surfaced in jgm/djot#213 ("Document implicit section behavior") and is still undocumented today.

The full stated rationale is jgm's reply in your own jgm/djot#221:

> I think we're using `<section>` correctly, judging from the documentation [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section).

[...]

PR #452, opened 2026-08-02
test: check combinations of constructs, and fix the first thing it found

dereuromark:
The corpus pins **constructs**. Nothing pins what happens when two of them meet, and that is where every cross-engine divergence in #427 lived: nested headings were covered, attributes were covered, no case gave a nested heading attributes — so four implementations held four different answers and every suite stayed green.

A pair space is larger than a hand-written case list, so it has to be generated rather than remembered.

`npm run combinatorial:check` walks a curated product of axes (heading level × attribute provenance × container nesting × trailing body = 120 documents), renders each through carve-js, the executable spec, carve-rs and carve-php, and diffs.

No expected-output files: **the oracle is agreement**, which is exactly the property that was missing. A second oracle covers what agreement cannot — structural invariants (no dangling `href="#id"`, no duplicate DOM id, every heading reachable by a fragment) that hold whatever the agreed answer turns out to be, so they fire even when every engine agrees and all of them are wrong.

## What the first run found

Four classes, all in the pair space, none previously visible:

| # | Class | Cases |
|---|---|---|
| 1 | **exec-spec never implemented the implicit `[Heading][]` fallback** — renders it as literal text | 8 |
| 2 | **carve-rs resolves `[Heading][]` into a blockquote**, which PART 11 R1 says is declined | 20 |
| 3 | **carve-php and exec-spec decline `[Heading][]` inside a list item**, which R1 says resolves | 10 |
| 4 | **carve-rs drops a block attribute line inside a list item** — `- {a=b .c}` renders as literal `<p>{a=b .c}</p>` and the attributes vanish | 6 |

Class 1 is the sharpest: every corpus case using `[X][]` also defines `[X]: url`, so the heading fallback — a README headline feature, and the rule I wrote into PART 11 R1 in #430 — has **zero corpus coverage** and the spec's own executable reference does not implement it.

**Left unfixed here.** A divergence is a question, not a verdict, and four of them are four separate decisions. So the script is deliberately **not wired into CI yet** — it would land red. Wire it once these are settled, promoting each answer to a corpus case in `docs/examples/edge-cases.md`.

## Two things it taught about its own design

**Engine provenance is printed.** A CLI engine is whatever its checkout is sitting on. My first run reported *two* divergence classes that were nothing but a stale working tree — carve-php was on a feature branch, 27 commits behind, dirty — and one of those classes was my own already-merged fix simply not being present. The output now names each engine's branch, revision and dirty state, so that costs a glance instead of an investigation.

**The executable spec throws its refusals rather than returning them** (`formal-core-check.mjs:73` catches it the same way). Testing the return value made that branch dead code, which would have filed every out-of-subset document as an engine *error* the first time an axis reached the subset boundary. Caught in review; refusals are now caught, counted and reported separately. Verified live: 200 levels of blockquote returns `REFUSED: block nesting exceeds MAX_NESTING_DEPTH`.

## Also

Extracts the sibling engine paths into `scripts/lib/engine-locations.mjs` so the two differential runners cannot drift about which directory "the php engine" means. `compare:impls` re-verified against clean checkouts.

Render options (`sections`, `sourceLine`) are not an axis yet: neither CLI exposes them and the executable spec implements neither, so there is nothing to compare across. The script says so in its output rather than running quietly narrower.

636 tests pass.


dereuromark:
Pushed the first fix on top, since it is the checker's own repo and the finding was in the executable spec.

**`4a1c511` — the oracle now resolves an implicit heading reference (#453).**

Every corpus case using `[X][]` also defined `[X]: url`, so the fallback branch was unreachable from the corpus and `formal-core-check` reported full conformance without ever running it.

Proof it was a live gap rather than a tidy-up, by reverting just the oracle change against the new corpus case:

| `scripts/spec/html.mjs` | result |
|---|---|
| at HEAD (before fix) | `DEFECTS (parse but diff): 1` |
| with fix | `513/513 conformant, DEFECTS: 0` |

Corpus case 170 pins the whole boundary in one document: a case-folded match, an unmatched label staying literal, a quoted heading declining, a list-item heading resolving, and a real definition beating a same-named heading.

The checker drops from 44 divergences to 36 as a result. The remaining three classes are the two engine bugs (markup-carve/carve-rs#410, markup-carve/carve-php#572) and the undecided attribute-line question (#454); fixes for the two engine bugs are next.

dereuromark:
Rebased onto main. The conflict was in `docs/examples/edge-cases.md`, where main's "Attribute braces on a list-item marker line" (#457) and this branch's "Implicit heading references with no definition" both append a section, interleaved around a shared `::: compare` block. Resolved by taking main's file whole and appending this branch's section after it - the branch's diff is a pure addition, so nothing of either is lost. 127 sections, both present.

Gates on the rebased commit, matching what CI runs: `npm test` 638 pass / 0 fail / 4 skipped, `core:check` 514 corpus inputs with 0 defects and 0 refusals, `docs:build` clean, `lint:mermaid` clean.

## One consequence worth stating before this merges

Corpus case 170 encodes the R1-correct answers, and two engines currently disagree with it. Checked just now against each engine at its own `main`:

| engine | case 170 |
| --- | --- |
| carve-js | matches |
| carve-rs | **differs** - resolves `[Quoted][]` under a blockquote, which R1 declines (class 2 in the description) |
| carve-php | **differs** - declines `[In an item][]` in a list item, which R1 resolves (class 3) |

The description says classes 2 and 3 are left unfixed, and that is the right call for the *script*. The corpus case is different in kind: engines consume it through the `tests/spec` submodule, so the next spec-pin bump in carve-rs and carve-php turns their corpus suites red.

Nothing breaks on merge - the pins are explicit, so it surfaces when someone bumps, which is arguably the mechanism working. But it turns two open questions into two downstream CI failures with a delay, so it should not arrive as a surprise on an automated bump PR. Filing the two engine issues now so the failure has an explanation waiting.

Issue #450, opened 2026-08-02
A bare ::: inside a code fence or comment block closes the container around it


dereuromark:
A colon-fence container collects its body by scanning raw lines for a closer. That scan does not know a fenced code block, a raw block or a comment block is opaque, so a bare `:::` line inside one closes the container it is sitting in - and everything after it reparses in the wrong place.

````
::: note
```text
:::
```
body
:::
after
````

carve-js and carve-rs both produce:

```html
<aside class="admonition note">
  <pre><code class="language-text">
</code></pre>
</aside>
<pre><code>body
:::
after
</code></pre>
```

The container closed at the fence-internal `:::`, so the real closer, `body` and `after` all fell out of the tree and got swept into a stray code block. The correct output keeps the code block and `body` inside the aside, closes at the second `:::`, and leaves `after` at the top level.

## Coverage

| opaque construct holding a bare `:::` | carve-js | carve-rs | carve-php |
| --- | --- | --- | --- |
| backtick code fence | wrong | wrong | **correct** |
| tilde code fence | wrong | wrong | **correct** |
| comment block | wrong | wrong | wrong |

carve-php is the oracle for the two code-fence rows - it already skips those spans. The comment-block row is wrong everywhere, php included: the `%%%` body is scanned for closers just like the rest.

Measured against `origin/main` of each engine, php through a worktree with a real vendor copy.

## Why it survived

The corpus has cases for colon fences and cases for code fences, but none that puts a fence-shaped line inside an opaque span inside a container. Nothing could fail.

It is also invisible from the well-formed side: a document that never writes `:::` inside a code sample renders identically either way. It shows up exactly when someone documents Carve in Carve, which is what the docs and every tutorial do.

## What is owed

1. Corpus cases pinning a bare closer inside a backtick fence, a tilde fence and a comment block, each inside a container - plus a non-bare opener inside a fence, since that variant becomes load-bearing under exact-length closers.
2. Engine fixes in carve-js and carve-rs (code fences), and all three (comment blocks). The body collector must skip an opaque span using the same predicates the block parser uses to recognize it, rather than a second copy that can drift.

## Relationship to the fence-direction work

[...]

Issue #442, opened 2026-08-01
The round-trip corpus never nests containers, so every writer's fence width went unchecked


dereuromark:
`tests/corpus-roundtrip/` holds 17 documents and not one of them contains `::::`. `tests/corpus/` has exactly one nested-container case, `68-nested-containers.crv`, two levels deep with both fences already correct. So the writer invariant `toHtml(fmt(x)) == toHtml(x)` was pinned only over inputs where fence width cannot be got wrong.

It is got wrong. Every engine's canonical writer picks a container's fence width from a one-level lookahead, which breaks at three levels of nesting:

```
::::: a

:::: b

::: c
X
:::

::::

:::::
```

comes back from `fmt` as

```
:::: a
:::: b
::: c
X
:::
::::
::::
```

and the outer two fences are now equal width, so the middle container is no longer nested. Same output from carve-js, carve-rs and carve-php. Engine-side issues: markup-carve/carve-js#496, markup-carve/carve-rs#385, markup-carve/carve-php#551.

## What this repo owes

1. Round-trip corpus cases that nest containers three and four levels deep, mixing admonitions, generic divs and line blocks, so the width rule is exercised past the point where a one-level lookahead can pass.
2. A conformance corpus case for the depth-3 render, so the nesting semantics themselves are pinned and not just the writer.
3. While the area is open: the equal-length and inverted nesting forms are unpinned too. `::: note` / `::: tip` / `:::` / `:::` leaks the inner opener as literal text in Carve and nests in djot, and that divergence is not recorded in `docs/divergence-from-djot.md`. Discussion in #439.

Sequencing: the engine fixes land first, then the corpus cases here, otherwise the new fixtures fail every engine's CI on the submodule bump.

dereuromark:
Blocked on the engine fixes landing, and worth saying why rather than leaving it looking untouched.

`tests/roundtrip.test.mjs` runs against the **pinned** carve-js build, and there is no xfail or pending-engine mechanism in that harness - unlike the HTML corpus, a round-trip case either passes or reddens CI. So adding depth-3 and depth-4 cases before a fixed writer is pinned would knowingly break this repo's CI rather than document the gap.

Sequence that works:

1. markup-carve/carve-js#496 lands (in progress)
2. bump the carve-js pin here
3. add the round-trip cases, which then pass, and the conformance case for the depth-3 render

Doing it in that order means the cases arrive green and stay green, and the ratchet reads as "this is pinned" rather than "this is known-broken".

[...]

Issue #439, opened 2026-08-01
Invert container nesting level syntax


Brixy:
Unlike #427 this proposal is not crucial, but maybe worth considering from a user's point of view:

Nested containers currently work like this -- just like Djot:

```
:::: note
Outer.

::: tip
Nested.
:::
::::
```

It may make sense to invert this: always start with `:::`, add a colon for each 'deeper level'.

### It's easier to add a nesting level

Currently you have to edit this complete structure to add a nesting level; with the inverted structure you can just add another level without touch the rest

```
:::
Outer

::::
Inner

:::::
NEWLY ADDED AFTERWARDS
:::::
::::
:::
```

This feels in line with HTML: you can just add a nested `<div />` without touching the rest.

### It corresponds to the heading levels logic

Simply put: more `#` = deeper level:

```
# H1

## H2

### H3
```

Due to (now optional) container wrappers you really get nested HTML syntax when using headings.

'More indicators = deeper level' may correspond more to the "visual mnemonics" concept.

Moreover, it corresponds to other markup concepts, e.g. Asciidoctor's nested ordered lists:

```
. Step 1
. Step 2
.. Step 2a
.. Step 2b
. Step 3
```

### Easier to be combined with "admonitions

With this proposal an admonition _always_ starts like this:

```
::: note
Heads up — this is important.
:::
```

, no matter how many nesting levels there are or will be.

I guess this is a lot easier for people who use admonitions and discover or need nestability later on.

---

Please also consider [this answer](https://github.com/jgm/djot/discussions/299#discussioncomment-9217211).

dereuromark:
Thanks - this one is better founded than it looks at first glance, and digging into it turned up a real bug in the current scheme. Long answer, because the answer changed my mind partway through.

## What the proposal actually costs

Inversion is not a syntax swap. Exactly one normative rule carries it:

```
colon_fence_close = colon_fence:close where len(close) >= len(open)
                                                       ^^ becomes ==
```

(`resources/grammar.ebnf` colon_fence_close, PART 9 §12 note 2.)

With an exact-length closer matched against the innermost open fence, nesting only requires *different* lengths, so both directions parse:

```
:::: note        :::  note
::: tip          :::: tip
:::              ::::
::::             :::
```

Longer-outer documents keep working as long as their closers match their openers exactly. Longer-inner starts working. Nothing is added to the syntax, and no author has to pick a side.

## What the engines do today

[...]

PR #455, opened 2026-08-02
spec: a colon fence closes on an exact length match (PART 9 §12)

dereuromark:
Implements the resolution on #439. Engine side: markup-carve/carve-js#505.

## One guard

```
colon_fence_close = colon_fence:close where len(close) = len(open)
```

matched against the INNERMOST open fence. Everything in this PR follows from that:

| form | before | after |
| --- | --- | --- |
| equal-length fences | inner opener is literal text | nests |
| wider inner fence | closes the outer block | nests |
| opener with no closer | opener and the rest of the document become text | container, closed at end of input |
| one closer, several openers | closes one, the rest degrade to text | closes one, the rest close at end of input |

Exact matching is direction-neutral, so longer-outer documents keep parsing and longer-inner ones start parsing. The direction becomes a question for the writer alone, and `carve fmt` now emits the inward-widening form.

## The closer is optional now

Not a second idea - the counterweight to the first. Exact matching makes a mistyped closer leave a container open, and under the old rule that turned one wrong character into a document of literal text. Auto-closing at end of input costs the container's extent instead, and matches djot. Lint and the language server should diagnose it; the parser should not. This is deliberately *not* the `%%%` rule from §28, where an opener with no closer opens nothing - a comment block is invisible either way, so failing closed is free there.

## Why the rule changed

Fence width was a quoting relation borrowed from the code fence, being used as a depth count. Under equal-or-greater, a container had to outrank every container anywhere in its subtree, so a canonical writer had to know its own maximum depth before it could emit its opening line. Every engine got that wrong the same way and silently unnested the middle container of a three-level document (#442). Width is now local depth, computable on the way down.

Code fences keep `>=`. There the length axis really is quoting: opaque content that never nests, which must be able to hold a shorter fence.

## What moved

**Ten corpus fixtures.** Four are the degenerate forms pinned in #447 - flipping as a visible diff is exactly what pinning them was for. Four more are paragraph-interruption and list-continuation cases that leaned on an unclosed opener degrading to text. The last three flip for a different reason: `:::note` with no space was accepted by every engine though the grammar always required one, and the old closer lookahead is what hid it.

**The executable spec.** `scripts/spec/layout.mjs` is a second implementation of the same rule and moves with it, so the formal Core grammar and the engines still agree on all 512 inputs rather than the gate quietly measuring the old rule.

**The round-trip pair keeps its widest-outer inputs.** They parse the same under exact matching, so `09` and `10` now prove the writer *normalizes* an old-direction document into the canonical form instead of echoing what it was given.

**Section 13 of `divergence-from-djot.md` shrank.** Carve now agrees with djot on equal-length nesting and on closing at end of input - both were divergences a week ago. What survives: the exact closer, one-closer-closes-one against djot's greedy closer, a shorter fence being an opener rather than content, inward widening working here and not there, and the required space after the fence. Re-measured against djot.js rather than recalled.

## Pin

`package.json` points at the carve-js branch head, since the conformance run here needs an engine that implements the rule while that repo needs a corpus that describes it. Re-point at carve-js main once markup-carve/carve-js#505 lands.

## Related

#450 is the opaque-span bug this turned up - a bare `:::` inside a code fence closing the container around it. Fixed in the carve-js PR because exact closers cannot be correct without it; carve-rs still needs it independently, and all three need the comment-block half.


dereuromark:
Checked the two things this PR said were outstanding.

**The pin is done.** `ff77cee` was the carve-js branch head, and the squash merge of markup-carve/carve-js#505 produced `5e43fc1` - `git branch -r --contains ff77cee` returns nothing, so the old pin referenced a commit on no branch, which installs until the deleted branch is collected and then stops. It now points at `5e43fc1`, moved with `scripts/bump-carve-pin.mjs`. (I had the same fix ready in a worktree and found this already pushed - no change needed.)

Re-verified against the merged engine build rather than the branch one: `npm test` 638 pass / 0 fail / 4 skipped, `core:check` 514 inputs with 0 defects and 0 refusals, `docs:build` and `lint:mermaid` clean.

**The engine side is not done, and it is two engines wide.** carve-js implements the rule; carve-rs and carve-php do not. Measured just now against each engine's own main:

```
:::
::: note
body
:::
:::
```

| engine | result |
| --- | --- |
| carve-js | nests - `<div><aside class="admonition note">` |
| carve-rs | inner opener is literal text, trailing `:::` becomes its own paragraph |
| carve-php | identical to carve-rs |

Same for an unclosed `::: note`: carve-js emits the container, both others emit a paragraph.

Ten corpus fixtures moved here, four of them the degenerate forms pinned in markup-carve/carve#447 precisely so a change would show. Both engines consume the corpus through `tests/spec`, so the next pin bump turns their corpus suites red on those cases - nothing breaks on merge, since the pins are explicit, but the automated bump PR will fail with no explanation attached.

Filed with repros and the rule: markup-carve/carve-rs#421 and markup-carve/carve-php#587.

PR #456, opened 2026-08-02
spec: a container closer pairs with its opener's width (proposal for #439)

dereuromark:
Draft. **Step 2** of the sequencing recorded in markup-carve/carve#439. Step 1 is
markup-carve/carve-js#506.

`colon_fence_close` moves from `len(close) >= len(open)` to `==`, matched against the
**innermost** open fence. That single guard is the whole of the "inversion" proposal:
nesting then needs only *different* lengths, so both directions parse and every existing
well-formed document keeps working.

```
:::: note        :::  note
::: tip          :::: tip
:::              ::::
::::             :::
```

Inversion is a **writer** rule - the canonical direction `carve fmt` emits and the docs
teach - not a parser rule. That split is why this ships without a migration.

**Equal-width nesting needs one thing beyond the guard**, in the automaton as much as in the
engines: the closer scan counts depth. A bare fence is a closer whenever a container of its
width is open; only a *non-bare* fence line of that width opens a nested one. Stopping at
the first bare fence ended the outer container on the inner one's closer.

**Unclosed containers close at end of file** (or at the end of their enclosing scope)
instead of degrading the whole tail of the document to literal text. Exact-width closers
make a mistyped closer likelier, and that failure is too loud to pair with a stricter rule.
This is deliberately *not* the `%%%` rule of §28: a comment's content is invisible, so
auto-closing one would hide the rest of the file, while a container's content still renders.

The executable spec follows in both halves - `findColonCloser` counts depth and returns
end-of-input rather than `-1`. Four call sites that asked "is there a closer ahead" are
**deleted** rather than left as conditions that can no longer be false.

**Twelve fixtures flip**, which is exactly what pinning the degenerate forms in #447 was
for: all four `68-nested-containers` cases plus the generic-div, line-block,
paragraph-interruption and list-item-fence cases that exercised the same rules.

## One disagreement with step 1, found here

For `- ::: note` / `  - para text` (fixture `114-…-5`), this executable spec puts the nested
list **inside** the auto-closed aside - the container runs to the end of its enclosing
scope, which §12 now states explicitly. carve-js#506 closes the aside **immediately**,
leaving it empty with the list as a sibling.

The spec's reading is the intended one, so carve-js#506 needs a follow-up before either
merges. Recording it here rather than letting the two land and disagree.

Not in this PR: carve-php and carve-rs (step 3) and the lint/LSP diagnostic for an unclosed
container (step 4).

### DIFF
--- a/docs/divergence-from-djot.md
+++ b/docs/divergence-from-djot.md
@@ -511,17 +511,23 @@ ASCII before slugging (section 1 above), so `# Don't repeat yourself` gives
 `Don-t-repeat-yourself` whether the heading
 renders with a curly apostrophe or a straight one.
 
-## 13. Container nesting is strict, and an unclosed opener is text
+## 13. A container closer pairs with its opener's width
 
-Both languages spell a container `:::` and close it with a bare colon fence,
-and both require the closer to be at least as long as its opener. Three rules
-around that agreement differ, in both directions.
+Both languages spell a container `:::` and close it with a bare colon fence.
+Djot requires the closer to be at least as long as its opener; Carve requires
+it to be **exactly** as long, matched against the innermost open fence.
 
-**Equal-length fences.** Djot nests them: an inner `:::` inside an outer `:::`
-is a child container. Carve does not - a container has to be strictly wider
-than every container below it. At equal length the inner opener is neither a
-closer (a closer must be bare, and `::: tip` carries a type word) nor an
-opener, so it stays content:
+That one guard makes nesting direction-neutral. Only *different* lengths are
+needed, so both of these nest:
+
+```
+:::: note        :::  note
+::: tip          :::: tip
+:::              ::::
+::::             :::
+```
+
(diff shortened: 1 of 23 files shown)


dereuromark:
Superseded by #455, which implements the same resolution. Its engine half, markup-carve/carve-js#505, has already merged.

The two were written in parallel against #439 and reach the same guard: `len(close) = len(open)` matched against the innermost open fence, with the direction left to the writer. Closing this to keep one spec of record.

Carried over from this branch's framing: the sequencing note (engine first, then the corpus, because neither repo can go green while the other still describes the old rule) is exactly how #455 and carve-js#505 were landed, via mutual branch-head pins.

One difference worth recording rather than losing: this branch proposed an allowlist asserting that corpus fixtures STILL DIFFER from the pinned spec. #455 instead flips the fixtures and moves both pins together, so there is no interim state where the corpus and the engine disagree and no list that can rot. Same goal, fewer moving parts.

Issue #434, opened 2026-08-01
Headings: two ways to continue onto the next line, one of them a silent Markdown footgun


dereuromark:
## Summary

A heading currently has **two** ways to spill onto the next line, and one of them is a silent, unlintable corruption for anyone arriving from Markdown. Proposal: make headings **single-line**, ending at the newline, matching CommonMark.

This closes the half-applied divergence in `docs/divergence-from-djot.md` section 7.

## What the spec says today

`resources/grammar.ebnf:206-236` (MULTI-LINE HEADINGS, normative). Verified against carve-js and djot.js side by side. Carve matches Djot exactly.

| Source | Carve / Djot | CommonMark |
|---|---|---|
| `# Title` + `Some text.` | one `<h1>Title␤Some text.</h1>`, id `Title-Some-text` | `<h1>` + `<p>` |
| `## A` + `## B` | one `<h2>A␤B</h2>`, id `A-B` | two `<h2>` |
| `# A` + `## B` | two headings (different `#` count breaks the fold) | two headings |
| `# X` + `- item` | heading + `<ul>` | same |

Three continuation rules where a Markdown author assumes one:

1. a following line with the **same** `#` count folds in (marker stripped),
2. a following line with **no** marker folds in,
3. a **different** `#` count starts a new heading.

## Why this is a problem

**It is silent.** `lintCarve('# Title\nSome text.\n')` returns `[]`. No warning, no error. The document renders, just wrong: the title text is wrong, the auto-id is wrong (so `</#id>` cross-references and TOC anchors break), and the intended body paragraph disappears into the title. `fmt` then re-serializes the corruption as canonical.

**It is frequent.** Counted the headings in this repo's own `.md` files: **136 of 1854 (7.3%)** are immediately followed by a non-blank line. Every one would corrupt if the file were Carve. That is the optimistic number - this repo is close to markdownlint MD022 clean. Hand-written READMEs and LLM-generated markup are worse, and LLM-authored Carve is a stated target audience (carve-skill).

**It is already documented as a trap and not fixed.** `docs/edge-cases.md:570` calls it "the biggest authoring trap in the heading syntax". A documented trap is still a trap.

## The decisive argument: section 7 is only half applied

`docs/divergence-from-djot.md` section 7 already broke from Djot on exactly this class of bug, with this stated reasoning:

[...]

PR #459, opened 2026-08-02
docs: retire the last references to heading folding

dereuromark:
#451 made headings single-line but left six places still describing the folding it removed. Each is now either false or advertises a rule that no longer exists.

**The one that matters** is `resources/grammar.ebnf`: a definition term is defined as folding a following plain line *"like a heading"* — a comparison that now points at a construct doing the opposite. The comparison is gone.

Whether `::` terms should **also** become single-line is the open question #434 raised as a same-pass follow-up. It is recorded at the rule rather than answered here; **term behavior is unchanged**.

The rest:

| file | was |
|---|---|
| `resources/carve-core.ohm` | listed "multi-line heading folding" in what the executable spec covers |
| `docs/grammar.md` | same, in the coverage sentence |
| `docs/edge-cases.md` | "Plain text folds into a heading too", contrasting it with a blockquote |
| `docs/examples/edge-cases.md` | introduced the nested-lazy cases as folding into the heading — what they actually pin is that the line stays inside the **item**. Their expected output already changed in #451; only the prose was left behind |
| `docs/validation.md` | listed carve-php's `heading-lazy-continuation` rule |

That last row lands with markup-carve/carve-php#589, which removes the rule because the habit it warned about ("Markdown starts a new block here; Carve does not") is no longer a hazard.

No fixture changes: `corpus:build` reproduces all 514 pairs byte-identically.

Closes the wording follow-up in #434.

---

**Rebased onto main (merge, not force-push) — most of this landed via #458 already.**

#458 covered `MAINTAINING.md`, `docs/validation.md`, `resources/carve-core.ohm`
and the `resources/grammar.ebnf` definition-term note. Two conflicts resolved in
main's favor:

- `carve-core.ohm`: both removed "multi-line heading folding" from the coverage
  list; main also says *why* (out of the language, not just this pipeline).
- the definition-term note: this branch left #434's follow-up open, main
  **answers** it — terms keep folding, deliberately. What made the heading's
  fold a defect was having two ways to spill, one silent; a term has one, so it
  already satisfies the principle the heading violated, and a long term stays
  wrappable.

What remains here is the part #458 missed, and it is the reason this PR is still
worth merging:

| file | was |
|---|---|
| `docs/edge-cases.md` | "Plain text folds into a heading too" |
| `docs/examples/edge-cases.md` | introduced the nested-lazy cases as folding into the heading; what they pin is that the line stays inside the **item** |
| `docs/grammar.md` | "headings incl. multi-line folding" in the coverage sentence |

Plus one neither PR caught: the **COLUMN ZERO** note in `grammar.ebnf` said an
indented `#`-line under an open heading "folds into the heading as literal
text". There is nothing to fold into now — it starts its own paragraph.

`docs/validation.md`'s row is already on main; carve-php#589 is closed,
superseded by carve-php#586, which removed the rule.

638/638 tests, `corpus:build` byte-identical.


dereuromark:
Superseded by #458, which fixed the normative wording. The three prose spots it did not reach are reopened as #461.

PR #482, opened 2026-08-03
ci: run the conformance checker, and stop a stale build reading as a current one

### LINKED ISSUE #475: ast:check never runs in CI, and skipping an engine is indistinguishable from passing
`ast:check` is the only thing that measures PART 12 conformance. It does not run in CI, and when it does run it silently skips whatever is not on disk.

```
$ grep -rn "ast:check\|ast-conformance" .github/workflows/
(nothing)
```

`ci.yml` runs `corpus:build` + `git diff --exit-code`, `npm test`, `core:check`, `docs:build` and `lint:mermaid`. The conformance checker is not among them - it needs sibling checkouts (`../carve-rs`, `../carve-rb`, `../carve-php`), which CI does not have, so it was never wired in.

And when a checkout is absent it prints

```
carve-rb: checkout not found, not checked
```

and moves on. A run with no satellites present looks almost exactly like a run where every satellite passed: the only difference is one line of prose in the middle of the output.

## What that has been hiding

Ran locally against all four engines over 524 corpus documents:

| engine | shape parity | missing pos | **schema violations** |
| --- | --- | --- | --- |
| carve-js (reference) | 0 | 4 | 0 |
| carve-rs | 25 | 5 | 0 |
| **carve-rb** | 65 | **229** | **153** |
| carve-php | 65 | 2 | 0 |

carve-rb is emitting **153 schema violations** against the published `ast-schema.json` - including `/children/0/items/0 must have required property 'type'`, a node with no type at all - plus 229 nodes with no position. None of it has ever been reported, because nothing runs the checker that would report it.

That is notable given carve-rb wraps carve-rs, which is clean on schema: the divergence is in the Ruby serialization layer, not the parser underneath.

(The shape-parity column is new, from markup-carve/carve#473. The schema and position columns are pre-existing checks that simply never ran.)

## Two separate problems

1. **The checker does not run.** Wiring it into CI needs the satellite checkouts, which is real work - a matrix job that clones them, or a scheduled run rather than per-PR. Either beats the status quo, where the number can drift arbitrarily far and nobody learns.
2. **Skipping is indistinguishable from passing.** A missing checkout should be reported as `NOT MEASURED` and, once CI has the checkouts, should fail rather than print a line. The current wording is the same shape as the `.todo` problem the corpus runner already fixed for itself: "silently skipped, which is exactly how 14 spec categories once went unvalidated" is a comment in `test/corpus.test.ts` about this same failure mode.

## Suggested order

Fix 2 first - it is a few lines and it makes 1 honest. Then decide where the checker runs, and treat carve-rb's 153 schema violations as its own piece of work.

dereuromark:
Closes #475.

`ast:check` is the only thing that measures PART 12 conformance and it ran in **no workflow
at all**. It needs the satellite checkouts, which is why it was never in `ci.yml` - so it
gets its own scheduled job that clones and **builds** all four engines and sets
`CARVE_REQUIRE_ALL_ENGINES=1`.

Scheduled rather than per-PR because what it catches is cross-repo *drift* - an engine's
main moving away from the published AST while this repo stands still - which is a
daily-cadence problem. Building three engines on every PR would add minutes to a gate that
would nearly always say the same thing.

carve-rb is provisioned too rather than left out with the require flag on. A scheduled job
that can never go green gets muted, which is this same issue wearing different clothes.

The issue's second half - "skipping is indistinguishable from passing" - **is already fixed
on main** (`NOT MEASURED`, the roll-up, and the opt-in exit 1). I left it alone.

## The part that is not what the issue asked for, and matters more

**The checker reads whatever artifact is on disk and reports it as the engine's
conformance, with nothing in the output about its age.** That is how this issue's own table
came to be wrong.

Measuring carve-rs against the build that happened to be sitting in its checkout:

| carve-rs build | findings |
|---|---|
| Aug 1 binary (what the checker found) | **286**, incl. 144 schema violations |
| same checkout, rebuilt | **4** |

The 144 were the pre-node definition-list shape (`{terms, definitions}`) that carve#435
settled and the engine had *already stopped emitting*. carve-rb inherited the identical 144
through its equally stale compiled extension - which is also why its numbers looked so
alarming. Both runs looked exactly the same in the output.

So each built engine now reports its build date, compares it against the newest source file
in its own checkout, and says so when the source is newer:

```
carve-rs (over target/release/carve [built 2026-08-01 13:30, STALE - source is newer,
  rebuild before believing this], 30 documents): 5 findings, 5 distinct
carve-rb (over carve-rs [built 2026-08-01 12:38, STALE - ...], 30 documents): 6 findings

STALE BUILDS: carve-rs, carve-rb - findings above are from an OLD build.
Rebuild those engines and re-run before recording any number from this run.
```

Under the require flag a stale build fails the run. A stale build is worse than the skip
this script already reports: **a skip produces no number, and a number gets believed and
filed.**

## On the issue's figures

Worth re-measuring before acting on them. What I can confirm with fresh builds is
carve-rb's shape: **144 schema violations and 227 missing positions** against the issue's
153/229 - so that engine does have a real problem, and it is in the Ruby serialization
layer, since carve-rs underneath is clean at 4 findings. The carve-rs column, though, was
an artifact.

Spec suite: 680 tests, 0 failures.

### DIFF
--- a/.github/workflows/ast-conformance.yml
+++ b/.github/workflows/ast-conformance.yml
@@ -1,0 +1,118 @@
+name: AST conformance
+
+# `ast:check` is the only thing that measures PART 12 conformance, and until now
+# it ran nowhere (carve#475). It needs the satellite checkouts, which is why it
+# was never wired into ci.yml - so it lives here instead, on a schedule.
+#
+# SCHEDULED rather than per-PR on purpose. What this catches is cross-repo
+# DRIFT: an engine's main moving away from the published AST while this repo's
+# main stands still. That is a daily-cadence problem, and cloning and BUILDING
+# three engines on every PR would add minutes to a gate that would almost always
+# say the same thing. workflow_dispatch is here for when you want it now.
+#
+# CARVE_REQUIRE_ALL_ENGINES=1 turns a missing or unbuildable engine into a red
+# run. That is the whole point of running it here: the checkouts are provisioned
+# deliberately, so one dropping out is a broken job, not a line of prose.
+
+on:
+  schedule:
+    # 04:00 UTC daily, after the downstream bump workflow has had a night to run.
+    - cron: '0 4 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  conformance:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out carve
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
+
+      - run: npm ci
+
+      # Each engine is checked out and BUILT from its own main. A build is not
+      # optional here: the checker reads whatever artifact is on disk, and a
+      # stale one reports the old behavior under the current name - which is
+      # exactly how carve#475's own figures came to be wrong.
+      - name: Check out carve-js
+        uses: actions/checkout@v4
+        with:
+          repository: markup-carve/carve-js
+          path: carve-js
+          submodules: recursive
+
+      - name: Build carve-js
+        working-directory: carve-js
+        run: npm ci && npm run build
+
+      - name: Check out carve-rs
+        uses: actions/checkout@v4
+        with:
+          repository: markup-carve/carve-rs
+          path: carve-rs
+          submodules: recursive
+
+      - name: Build carve-rs
+        working-directory: carve-rs
+        run: cargo build --release
+
+      - name: Check out carve-php
+        uses: actions/checkout@v4
+        with:
+          repository: markup-carve/carve-php
+          path: carve-php
+          submodules: recursive
+
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.3'
+
+      - name: Install carve-php
+        working-directory: carve-php
+        run: composer install --no-interaction --no-progress
+
+      - name: Check out carve-rb
+        uses: actions/checkout@v4
+        with:
+          repository: markup-carve/carve-rb
+          path: carve-rb
+
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.2'
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      # rb-sys drives bindgen, which needs libclang. Same step carve-rb's own CI
+      # runs; without it `rake compile` fails on a fresh runner.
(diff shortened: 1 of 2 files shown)


dereuromark:
Correcting my own PR description, because I made the exact error it is about.

I wrote that carve-rb's figures were "what I can confirm with fresh builds - 144 schema
violations and 227 missing positions". **They are not fresh.** I measured carve-rb through
its compiled extension dated Aug 1 12:38 - the stale artifact this PR exists to flag. Its
numbers are unverified for the same reason carve-rs's were, and I should have applied the
finding to the engine I found it with.

What is actually established:

- carve-rs: 286 findings stale, **4 rebuilt**. Measured both ways, so this one is solid.
- carve-rb: 371 findings from a stale extension. **Unknown** until rebuilt - and since
  carve-rb wraps carve-rs, whose stale build carried all 144 of those schema violations,
  the honest expectation is that most of them evaporate too.

Rebuilding it now and will post the real number. The rest of the PR stands: the mechanism
does not depend on which engine turns out to be dirty.

dereuromark:
carve-rb rebuilt, and the real numbers are in. **The headline figure in #475 was an
artifact of the stale build, exactly as this PR predicted.**

| carve-rb | stale extension (Aug 1) | rebuilt (Aug 3) |
|---|---|---|
| schema violations | **144** | **0** |
| missing positions | 227 | 119 |
| shape divergences | - | 26 |

**Every one of the 153 schema violations the issue reports is gone.** They were carve-rs's
pre-node definition-list shape, inherited through a two-day-old compiled extension - not
anything the Ruby layer does. So the issue's most alarming claim, that carve-rb "is
publishing invalid AST today", is not true of current code.

What is real and remains:

- **119 nodes with no position**, concentrated in `text`, `definition_term`,
  `definition_description`. Since carve-rs is clean on positions, this is the binding
  dropping `pos` on the way out - a genuine gap in the Ruby serialization layer, which is
  where the issue said to look.
- **26 shape divergences** from the reference, several structural (`link`, `admonition`,
  `paragraph` node counts differing), which is carve#474's territory.

Worth noting how easily this went wrong: rebuilding carve-rb needed
`BINDGEN_EXTRA_CLANG_ARGS` for rb-sys to find `stdarg.h`, so the extension had silently
stayed at its Aug 1 state through every attempt to use it. A build that fails to refresh and
an artifact that reports as current are the same hazard from two directions.

The workflow in this PR builds every engine from scratch, so neither can happen there.

Issue #474, opened 2026-08-02
PART 12 requires cross-engine interop but never says whether adjacent text runs are coalesced


dereuromark:
PART 12 §1:

> a consumer written against one implementation MUST be able to read another's output

That is a statement about tree shape, and the spec never says what shape. Nothing defines whether **adjacent text runs must be coalesced**, and the engines disagree across most of the corpus.

```
foo_bar_baz and snake_case stay literal
```

| engine | that paragraph |
| --- | --- |
| carve-js | **1** text node |
| carve-php | **4** text nodes |

Same characters, same rendering, different tree. carve-js merges the runs left over when a delimiter does not form emphasis; carve-php emits one per run.

Measured with the shape-parity check in markup-carve/carve#473: **67 findings across 40 distinct shapes** in carve-php against the reference, nearly all this one cause. `01-emphasis-11` is 1 node vs 6; `01-emphasis-14` is 2 vs 6.

## Why it went unnoticed

`ast:check` validated each engine against `ast-schema.json` and counted missing positions. The schema permits any number of adjacent text nodes, so both trees are valid and the check had nothing to say. The one requirement that speaks to interop had no test behind it - which is also how carve-php#612 and carve-rs#413 reached main.

## The decision

Someone has to state the rule, and it belongs in PART 12 next to §1:

1. **Adjacent text runs MUST be coalesced.** The reference already does; carve-php and (probably) carve-rs would change. It makes the serialized tree canonical for a given document, which is what makes "read another's output" mean anything stronger than "parses".
2. **Adjacent text runs MAY be split**, and consumers must treat a run of text nodes as one. Cheapest, but it pushes the normalization onto every consumer forever, and quietly means two engines can never be compared node-for-node.

I lean to 1: PART 11 already requires a canonical *source* form, and a canonical *tree* form is the same argument one layer up. But it is a real break for at least one engine and wants a maintainer's call, not mine.

Whichever way it goes, `resources/ast-schema.json` cannot express it, so the rule needs prose plus the check in markup-carve/carve#473 - and that check should become a hard gate once the engines agree.

## Related

- markup-carve/carve-php#612 - a continuation cell as three nodes instead of one (fixed in markup-carve/carve-php#613)
- markup-carve/carve-rs#413 - the same cell as five

dereuromark:
The scope here is wider than text runs, and the reason it went unseen is sharper than "nothing checked it".

[...]

PR #483, opened 2026-08-03
spec: adjacent text runs are coalesced, so the serialized tree is canonical

dereuromark:
Decides markup-carve/carve#474 in favour of **option 1: adjacent text runs MUST be
coalesced.**

§1 requires that a consumer written against one implementation can read another's output,
and never said what shape that output has:

```
foo_bar_baz and snake_case stay literal
```

| engine | that paragraph |
|---|---|
| carve-js | **1** text node |
| carve-php | **4** text nodes |

Same characters, same rendering, different tree - and both valid against the schema, which
is why nothing caught it.

## Why normative rather than left to producers

Without the rule the interop requirement is satisfied by *any* tree, and "read another's
output" degrades to "parses". Coalescing is what makes the serialized tree **canonical for
a given document** - the same argument PART 11 already makes one layer down for canonical
source form - and it is what allows a divergence to be *measured* node-for-node instead of
argued about.

## Scope, deliberately narrow

- `text` nodes only, and only adjacency within one parent's child list.
- Does not merge across an intervening node, does not touch other types.
- Says nothing about how a producer represents runs internally - only what it publishes.
- **`escaped_text` stays distinct.** An escape is authored form and §5 keeps the two types
  apart on purpose; merging them would lose the distinction the type exists to carry.

## Why it is prose plus a checker, not schema

JSON Schema cannot forbid two adjacent array entries of the same shape. The shape
comparison in `scripts/ast-conformance.mjs` is what checks it - which is the argument for
that check becoming a gate, and markup-carve/carve#482 is what makes it run at all.

## Work this creates

carve-js already complies (verified: 1 node for the example above). carve-php and carve-rs
are the follow-up, and the measured counts to fix are in #474.

Spec suite: 683 tests, 0 failures.

### DIFF
--- a/docs/ast-json.md
+++ b/docs/ast-json.md
@@ -73,6 +73,34 @@ document into invalid text.
 An implementation that cannot place a node **omits `pos` rather than inventing
 one**, and says so. Absent is a fact a consumer can act on; a wrong span is not.
 
+## Adjacent text runs are coalesced
+
+A serialized node's children hold **no two adjacent `text` nodes**. Where a
+producer's internal tree has a run of them, they join into one before
+serialization, concatenating `value` in order:
+
+```
+foo_bar_baz and snake_case stay literal
+```
+
+is one text node, not four. An implementation that splits wherever a delimiter
+failed to open emphasis is publishing its parser's bookkeeping rather than the
+document.
+
+This is normative (§1a) because without it the interop requirement means very
+little: two engines can publish 1 node and 4 for the same characters, both valid
+against the schema, and "read another's output" degrades to "parses".
+Coalescing is what makes the tree **canonical** for a given document - the same
+argument PART 11 makes one layer down for canonical source form - and it is what
+lets a divergence be measured node-for-node instead of argued about.
+
+`escaped_text` is not `text` and does not merge with it: the two stay distinct
+on the wire because an escape is authored form.
+
+The schema cannot express this - JSON Schema has no way to forbid two adjacent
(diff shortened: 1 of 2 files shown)


dereuromark:
Correcting the PR description: I wrote that **carve-rs already complies**. It mostly does,
but not entirely, and the exception is interesting enough to record before this merges.

Scanning 60 corpus documents for adjacent `text` siblings, carve-rs has exactly one -
`03-links-12`, `[pre <http://h> post](/u)`:

```
rs:  text "pre "   text "http://h"   text " post"
js:  text "pre "   autolink ""       text " post"
```

So it is **two divergences stacked**, not one:

1. **Node type.** carve-js publishes an `autolink` node nested inside a `link`; carve-rs
   unwraps it to text. Unwrapping follows the links-never-nest rule, so carve-rs looks right
   here and carve-js looks like the outlier - a nested autolink is exactly what that rule
   forbids.
2. **Coalescing.** Having unwrapped it, carve-rs then leaves three adjacent text nodes,
   which is what §1a forbids.

Only the second is this PR's business, and the fix is the same one carve-php needs. The
first is a separate question about what an unwrapped autolink *becomes*, and I would rather
raise it than quietly settle it inside a coalescing change - fixing coalescing alone would
make the two engines' node COUNTS match while the types still disagree, which is worse than
a visible difference.

The rule as written stands either way. Flagging so the engine follow-ups do not inherit my
too-strong claim.

Issue #506, opened 2026-08-03
S4 and PART 9's lazy-continuation clause disagree for nested blockquotes, and the engines split along the seam


dereuromark:
Two clauses in the spec give different answers for a lazy continuation under a *nested* blockquote, and the three engines split along that seam.

```
> a
>> b
c
```

| engine | result |
| --- | --- |
| carve-js, carve-php | `c` folds into the INNER quote's paragraph: `<blockquote><p>a</p><blockquote><p>b c</p></blockquote></blockquote>` |
| carve-rs | `c` closes both quotes and starts a sibling paragraph |

Filed against the engine as carve-rs#452, but the disagreement is upstream of any engine.

## The two clauses

**PART 1, S4 (partial match):**

> LAZY CONTINUATION: if the innermost **MATCHED** context holds an OPEN PARAGRAPH and the residue is NOT an interrupting line, L folds into that paragraph and NOTHING closes.

`c` carries no `>`, so it matches **zero** containers - the innermost matched context is the document, and the document holds no open paragraph. Read strictly, nothing folds and the quotes close. That is carve-rs.

**PART 9, LAZY CONTINUATION - NORMATIVE (CommonMark-compatible):**

> After one or more blockquote_lines, a line that does NOT carry the '>' marker still continues the blockquote -- exactly as if it carried '>'.

Unqualified about depth, so `c` continues the quote whose paragraph is open - the inner one. That is carve-js and carve-php.

And "exactly as if it carried `>`" admits a **third** reading: one `>` at depth two would close the inner quote and continue the outer, which no engine does.

## What djot says, and what it cannot say

The clause claims CommonMark compatibility, so djot is worth asking - but it cannot arbitrate this case, because it does not nest here at all:

Input:

```
> a
>> b
c
```

Djot's output:

```html
<blockquote>
  <p>a &gt;&gt; b c</p>
</blockquote>
```

djot requires `> > ` for nesting, so `>> b` is literal text. It does show laziness folding into the open paragraph rather than closing containers, which points the same way as carve-js and carve-php - but on a document Carve parses differently.

**That is a second finding**: all three Carve engines treat `>>` as a nested blockquote and djot does not, and `docs/divergence-from-djot.md` does not mention it. Whether or not this issue changes anything, that page has a missing row.

## What needs deciding

1. Which reading is normative for a lazy line under nested quotes. S4 and the PART 9 clause should not be able to disagree.
[...]

PR #507, opened 2026-08-03
docs: Carve's blockquote marker does not need a space, and Djot's does

dereuromark:
`divergence-from-djot.md` had no row for the blockquote marker, so the page reads as if blockquotes behave the same in both. They do not, in the spelling that matters most:

| input | Carve | Djot |
| --- | --- | --- |
| `>quoted` | `<blockquote><p>quoted</p></blockquote>` | `<p>&gt;quoted</p>` |
| `>> quoted` | nested blockquotes | `<p>&gt;&gt; quoted</p>` |
| `> > quoted` | nested blockquotes | nested blockquotes |

Djot requires a space after `>` and spells nesting `> > `. Carve makes the space optional - CommonMark's rule, and the grammar says so directly:

```ebnf
blockquote_line = '>', [' '], inline_content, newline ;
```

That is the spelling in every Markdown document being ported. A marker that silently produced a paragraph instead of a quote is exactly the kind of migration trap [the migration guide](https://github.com/markup-carve/carve/blob/main/docs/migrate-from-markdown.md) exists to remove - and the page that lists the differences did not list this one.

## How it surfaced

While investigating #506 (PART 1's S4 and PART 9's lazy-continuation clause disagree about a lazy line under nested quotes). That clause claims CommonMark compatibility, so I asked Djot to arbitrate - and Djot parses the test document completely differently, because it does not nest on `>>` at all. The arbitration failed; the finding was the reason it failed.

## Tests

Two cases added to `tests/divergence-claims.test.mjs`, which is what keeps that page honest.

`> > quoted` is deliberately **not** asserted, and the file's absent-sections note now says why: both engines nest it - that is the claim - but they pretty-print the nesting differently (`<blockquote><p>` against `<blockquote> <p>`), and that file compares whitespace-squashed HTML. Asserting it would pin a formatting detail while claiming to pin a semantic one.

Spec suite 705 passing, executable spec clean.

### DIFF
--- a/docs/divergence-from-djot.md
+++ b/docs/divergence-from-djot.md
@@ -200,6 +200,27 @@ punctuation (`typo ,oops, happens` must not become a subscript).
 technical writing especially, biasing toward the literal reading avoids
 surprise lists.
 
+## 5b. The blockquote marker does not require a space
+
+**Djot:** `>` must be followed by a space (or end the line). `>quoted` is a
+paragraph, and `>> quoted` is a paragraph too - nesting is spelled `> > quoted`.
+
+**Carve:** the space is optional, so `>quoted`, `>> quoted` and `>>quoted` are
+all blockquotes, the last two nested. `> > quoted` also works and means the same
+thing.
+
+```carve
+>quoted
+>>nested
+```
+
+**Why.** This follows CommonMark, where the space after `>` is optional and
+`>>text` nests. It is the spelling in every Markdown document being ported, and
+a marker that silently produced a paragraph instead of a quote would be a
+migration trap of exactly the kind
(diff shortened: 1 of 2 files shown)


dereuromark:
Follow-up filed as #509: the same evidence, read as a proposal rather than as documentation.

The finding that changed my view while writing this page: `>` is the **only** block marker in Carve whose following space is optional. `#head`, `-item`, `1.item` and `:::note` are all paragraphs; `>quote` is a blockquote. PART 1's container table already carries the exception in brackets - `marker \`>\` [+ one space]` - while the heading and list productions have the space inline.

This PR stays documentation-only and is correct either way: if #509 lands, the row changes from "Carve differs from Djot" to "Carve used to differ".

PR #540, opened 2026-08-03
spec: restore the normative clauses #525 reverted, and gate the reference

dereuromark:
PR #525 was cut from a branch that predated #518 and merged four days of spec work back out of `resources/grammar.ebnf` on its way in. Net **-254 lines**, none of them mentioned in its description, and every gate stayed green - nothing in the suite asserts that a clause the docs describe still exists.

The symptom is visible from outside the repo: `docs/ast-json.md` cites `(§3a)` and the **published** `ast-schema.json` cites `PART 12 section 3a`, against a grammar that has had no §3a since 3 August.

## What came back out

| clause | what was lost | from |
| --- | --- | --- |
| PART 12 §3a | THE SERIALIZED TREE IS PRE-RESOLVE - deleted whole | #518 |
| PART 12 §1a | the merge is part of `parse(x)` not of serialization; the POSITIONS paragraph | #488 |
| PART 12 §4 | IMPLEMENTATION STATUS, reverted to numbers that were already stale ("869 block nodes and 54 left", carve-js#441 / carve-rs#333 as open) | #498 |
| PART 12 §7 | THIS COVERS EVERY DEFINITION KIND, NOT ONLY FOOTNOTES, with the formatter consequence and the rejected alternative | #518 |
| PART 9 §10 | I1 and I4: a `:::` opener interrupts closer or not | #514 |
| PART 9 §23 | MEDIAL GAPS in a line block | #491 |
| block algorithm | S4 lazy continuation folds into the innermost open paragraph at any depth, plus the same phrase in the blockquote production | #508 |
| math note | the correction that carve-php does NOT drop math attributes - the reverted text asserts a defect no engine has | #500 |

## Measured, not trusted from the diff

Two of the restored clauses make claims about output, so they were checked against the reference engine before restoring.

Input:

```
Text
:::
stuff
```

Output:

```html
<p>Text</p>
<div>
  <p>stuff</p>
</div>
```

That is what the restored I4 says and the opposite of what `main` says. Corpus `41-line-blocks-9` likewise renders the tab arithmetic the restored MEDIAL GAPS clause describes.

The I4 corpus citations are **updated** rather than restored verbatim: #525 inserted a case and shifted the numbering, so the unterminated-fence pin is now `81-paragraph-interruption-19` and the `:::` pin is `-20`.

## #525's own change is kept

The blockquote marker still requires a space, and #527/#530's later rewordings stand. After this PR the only differences from the pre-#525 text are exactly those - `diff` against the pre-#525 grammar shows seven hunks, all of them intended.

## The gate that was missing

`tests/normativity.test.mjs` resolved `PART 9 §N` citations only. PART 12's clause numbers carry a letter suffix (`1a`, `3a`) and were not checked at all, from any file.

It now resolves every PART 12 reference in the docs and in `resources/ast-schema.json`, expanding citation groups so `§1-2` reaches both ends rather than only the first. Run against the grammar as `main` has it today, the new test fails with six dangling references to §3a; against the restored text it passes.

### DIFF
--- a/resources/grammar.ebnf
+++ b/resources/grammar.ebnf
@@ -104,12 +104,23 @@
         content_column -> sublist (PART 9 §24 C3); otherwise paragraph /
         lazy item text.
      S4 PARTIAL MATCH. The unmatched containers are candidates to close.
-        LAZY CONTINUATION: if the innermost MATCHED context holds an OPEN
+        LAZY CONTINUATION: if ANY container in the open stack holds an OPEN
         PARAGRAPH and the residue is NOT an interrupting line (PART 9 §10 --
         list markers and plain text fold; visible openers interrupt), L
-        folds into that paragraph and NOTHING closes. Otherwise close the
-        unmatched containers and re-classify the residue in the surviving
-        context (S3).
+        folds into the INNERMOST such paragraph and NOTHING closes --
+        regardless of how many containers the line failed to match.
+        Otherwise close the unmatched containers and re-classify the residue
+        in the surviving context (S3).
+
+        This said "the innermost MATCHED context holds an OPEN PARAGRAPH",
+        which describes something no implementation does. A lazy line carries
+        no markers, so it matches ZERO containers and the innermost matched
+        context is the document root, which holds no open paragraph -- under
+        that wording `> a` / `b` would close the quote at depth 1. Every engine
+        folds there, and corpus 82-blockquote-lazy-continuation pins it. What
+        the wording did produce was a split at depth 2: carve-rs closed both
+        quotes where carve-js and carve-php folded into the nested one
+        (carve#506, carve-rs#452). Depth is not a parameter of this rule.
      S5 OPENERS. When S3/S4 classify the residue as a container opener (a
         quote marker; a list marker at an admissible column; a `+`
         continuation marker at the marker column, PART 9 §17 L3/L4; a fence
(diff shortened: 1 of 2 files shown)


dereuromark:
Closing rather than merging: everything here is already on `main`, and merging it now would take work back out.

This PR's only commit is `e8bbd90`, the restoration. It was also the base of #537, so when #537 squash-merged as `5e5d194` the restoration landed with it. `main` has carried §3a, §1a's merge clauses, §4's implementation status, §7's every-definition-kind rule, I1/I4, MEDIAL GAPS and the S4 lazy-continuation wording since then.

Measured against `main` as it stands, this branch is behind rather than ahead:

```
 .github/workflows/ast-conformance.yml |  55 ++++++-------
 CHANGELOG.md                          |  23 -------
 docs/ast-json.md                      |  17 +-----
 resources/grammar.ebnf                |  59 ++-----------
 resources/normative-clauses.txt       |  83 ---------------------
 scripts/normative-clauses.mjs         |  70 ------------------
 scripts/spec/ast-shape.mjs            |  77 ---------------------
 tests/ast-shape.test.mjs              |  64 ----------------
 tests/normativity.test.mjs            | 100 +++++-------------------
 ...
```

Merging it would revert #539's clause inventory, #543's `ast:check` fix and #537's §4 clause. The five lines it would add back are the superseded "Until that is decided" wording that #537 replaced.

The guard this PR proposed is on `main` in a stronger form: #539's `resources/normative-clauses.txt` pins every `-- NORMATIVE` clause **with counts**, and its citation test was merged with this one's - the range-aware scanner from here, the wider source list (schema + changelog) from there.

Superseded by #537 and #539.

Issue #561, opened 2026-08-03
Does a column-0 line continue a list item whose last block is a container? Three engines, three answers


dereuromark:
A line at column 0 following a blockquote inside a list item: does it continue the item?

The three engines give three answers, and one of them disagrees with itself.

```
. >
X
```

`. ` is a bare-dot ordered marker whose content is `>`, an empty blockquote. `X` is at column 0, outside the item's content column.

| X | carve-js | carve-rs | carve-php |
|---|----------|----------|-----------|
| `*` | inside the item | outside | inside |
| `-` | inside | outside | inside |
| `x` | inside | outside | inside |
| `|` | inside | outside | **outside** |

carve-php treats `|` differently from every other character, which is a bug on its own terms whatever this question resolves to - filed as markup-carve/carve-php#683.

## What the rule seems to say

A lazy continuation needs an open paragraph - that is what carve-js#557 and carve-php#654 established last week. Here the item's last block is a blockquote, and an empty one, so there is no open paragraph for `X` to continue. On that reading `X` ends the list and starts a new block, which is carve-rs's answer.

Against that: `X` is at column 0, so it is not the item's content by indentation either. Both readings point the same way, and yet two engines attach it.

So either the lazy-continuation rule has a case for "the item's last block is a container" that is not written down, or two engines are wrong. Worth stating explicitly in §11 or §14, because the reading that produces carve-rs's answer is the one the other two fixes were built on.

## Where it is not covered

The corpus has nothing with a container as an item's last block followed by a column-0 line. Every suite is green.

## Found by

`scripts/fuzz-impls.mjs --target=html --seed=101 --count=200`, two of the eight minimal reproducers in that run:

```
. >
*

. >
|
```

Build provenance for that run, per #558:

```
carve-js   2026-08-03 19:08
carve-rs   2026-08-03 19:41
carve-php  2026-08-03 02:32
```

PR #586, opened 2026-08-04
spec: the executable spec follows S4, and the corpus pins it

### LINKED ISSUE #572: A flush-left line after a marker-line quote: three engines, three answers, and one disagrees with itself
A flush-left line under a list item whose FIRST BLOCK is a quote written on the marker line splits the engines three ways, and one of them disagrees with itself.

Measured on today's mains, HTML normalized identically for all three:

**1. The quote has an open paragraph**

```
- > q
lazy
```

| engine | result |
| --- | --- |
| carve-js | `<ul><li><blockquote><p>q lazy</p></blockquote></li></ul>` |
| carve-php | same |
| carve-rs | `<ul><li><blockquote><p>q</p></blockquote></li></ul><p>lazy</p>` |

carve-rs disagrees with ITSELF here. Move the quote off the marker line and it folds like the others:

```
- item
  > q
lazy
```

gives `<li>item <blockquote><p>q lazy</p></blockquote></li>` in all three. The same document at the top level agrees too:

```
> q
lazy
```

is one quoted paragraph everywhere. So the marker-line lead is the only shape where carve-rs ends the list, and PART 1 S4 says a lazy continuation folds into the INNERMOST OPEN PARAGRAPH - which `q` is.

**2. The quote is empty, so nothing is open**

```
- >
lazy
```

| engine | result |
| --- | --- |
| carve-rs | `<ul><li><blockquote></blockquote></li></ul><p>lazy</p>` |
| carve-js | `<ul><li><blockquote></blockquote> lazy </li></ul>` |
| carve-php | same as carve-js |

Here the majority looks wrong to me. S4 requires an OPEN PARAGRAPH, and there is none: the quote is empty and the item's first block is that quote, so the item has no open paragraph either. Nothing is open, so nothing can be continued - carve-rs's answer.

**3. The lazy line is a block opener**

```
. >
|
```

| engine | result |
| --- | --- |
| carve-rs | `<ol><li><blockquote></blockquote></li></ol><p>\|</p>` |
| carve-php | same |
| carve-js | `<ol><li><blockquote></blockquote> \| </li></ol>` |

A `|`-delimited row is in §24 C3's block-opener set, and §10 says a block opener interrupts. carve-js folding it is the odd one out. (A lone `|` is not a valid row, so it lands as a paragraph either way - the question is whether it stays inside the item.)

The same split holds for `*`, `-`, `#`, `1.` and plain text as the following line, with carve-rs alone on one side in every case except `|`, where carve-php joins it.

## Why nothing caught any of this

No corpus document puts a quote on a list marker line and follows the item with a flush-left line. The construct is reachable, spelled in one line, and sits at the junction of three rules that each have corpus coverage on their own - S4 (lazy continuation needs an open paragraph), §24 C3 (what a line below the content column does), and §10 (block openers interrupt).

Found by the differential fuzzer (#510) with `--seed=101`, which generated `. >` followed by `*` on its own.

## What I think the answers are

1. carve-js and carve-php are right and carve-rs has a defect - its own next-line behavior is the tell.
2. carve-rs is right; the other two invent an open paragraph that S4 requires to exist.
3. carve-rs and carve-php are right; a block opener interrupts.

Each needs a corpus case, and 2 and 3 need the spec to say so explicitly, since the majority is on the wrong side of both.

dereuromark:
Closes #572 and the executable-spec half of #561.

PART 1 S4 - **NO OPEN PARAGRAPH, NO LAZY LINE** - landed with an unusual admission in it:

> Three engines gave three answers here and the reference is not among the right ones … The executable spec in `scripts/spec` also keeps it inside, so THIS repository's implementation moves too - recorded rather than glossed, because a rule whose own reference implementation does not follow it is worth flagging when it lands, not when someone next reads it.

This is that move, plus the corpus case that stops it drifting back.

## The change

The rule is about whether a paragraph is OPEN, so the fix is exactly that test: a quote line leaves the fold open only when the quote holds something, stripping nested markers first (`> > q` does, `> >` does not).

A first attempt generalized it - "does this line open a paragraph at all" - and broke corpus 161, where `- ::: note` is literal text on the marker line and *does* leave the item's paragraph open. The corpus caught it in the same run; the narrow version is the one that matches what S4 actually says.

## The corpus cases

```
- >
lazy
```

closes the item, and

```
- > q
lazy
```

folds - the pair the rule turns on. Byte-identical across carve-rs, carve-js and carve-php.

## Landing order

carve-rs already followed the rule. The other two are markup-carve/carve-js#587 (merged) and markup-carve/carve-php#703 (**must merge first**, or the nightly comparison goes red on these two cases).

The comparison is requoted from a run with both applied: 546/546 in all three engines across all five targets, `cross_impl_diffs=0`. The carve-php SHA in the table is that PR's head rather than a main commit, since that is the build that was measured.


dereuromark:
Superseded by #583, which landed the same executable-spec fix with three corpus cases while this was open. I did not see it in flight - two passes at the same clause from different directions.

The two arrived at the same narrow predicate, which is the reassuring part: **does this quote line carry content**, at both the marker-line and later-row sites. The broad version - "does this line open a paragraph at all" - is the one that looks right and breaks corpus 161, where `- ::: note` is literal text on a marker line and so DOES leave the item's paragraph open. Verified against main just now: that case still renders as literal text, and the three shapes the rule turns on all behave.

Nothing here is lost. #583's fixtures cover the same pair (`- >` + `lazy` closes, `- > q` + `lazy` folds) plus the top-level control, and its comparison table is measured from clean clones rather than my worktrees.

The engine half stands on its own: markup-carve/carve-js#587 is merged and markup-carve/carve-php#703 is open - both are what makes #583's fixtures pass in all three engines, and #583's own table records carve-js at 545/547 and carve-php at 544/547 without them.

Issue #631, opened 2026-08-04
Nothing pins that a floating attribute stops at a list-item boundary, though all four agree


dereuromark:
Four implementations agree on this shape and no corpus case pins it:

```
- a

  {.c}
- b
```

| implementation | commit | output |
|---|---|---|
| carve-js | `addf7aa` | `<li><p>a</p></li><li><p>b</p></li>` |
| carve-php | `22fa88b` | the same |
| carve-rs | `2d5d050` | the same |
| executable spec | `b9a01b3` | the same |

The floating attribute does not cross the item boundary: §15 A2a says it skips what renders nothing and attaches to the next VISIBLE block, and every implementation reads the item boundary as ending its scope rather than carrying it into the next item's paragraph.

markup-carve/carve-js#620 filed this as a divergence when carve-php's then-unmerged #747 branch attached the class to `b`. That did not survive to main, so the disagreement is gone - and with it the only thing that would have caught a future regression, because agreement is not a check.

Worth a case in the same family as `87-compact-list-blocks`. I did not add it to markup-carve/carve#628 because that PR is already re-quoting the comparison page for one corpus addition, and a second would mean another full run; keeping them separate keeps each set of numbers honest.

dereuromark:
Pinned by markup-carve/carve#634, now merged - corpus category `a-floating-attribute-stops-at-the-item-boundary`.

Re-measured all four before adding the case; still unanimous, so this is pure regression insurance. Your point was the right one: carve-js#620 filed the shape as a divergence, the divergence resolved itself before merging, and that removed the only thing that would have caught a future regression. Agreement is not a check.

Issue #629, opened 2026-08-04
An indented %%% still renders as text after #624, though the same rule consumes %%


dereuromark:
markup-carve/carve#624 made a comment recognized at any column, because folding it made `%% c` visible - "the one outcome a comment may never have". The branch it added excludes comment FENCES:

```js
if (!nm && !COMMENT_FENCE.test(line) && COMMENT_LINE.test(line) && itemLines.length > 0) {
```

so an indented `%%%` still folds, and still renders as text.

```
- a
 %%%
 hidden
 %%%
 b
```

| | output |
|---|---|
| carve-js | `<li>a` / `hidden` / `b</li>` - the `%%%` lines consumed, `hidden` is text |
| carve-php | `<li>a` / `%%%` / `hidden` / `%%%` / `b</li>` - everything visible |
| executable spec | the same as carve-php |
| carve-rs | `<li>a` / `b</li>` - the whole block swallowed |

## What the productions say

```
comment_line       = [whitespace], "%%", {character}, newline ;
comment_block_open = ("%%%", {'%'}):open, {character - newline} ;
```

`comment_line` allows leading whitespace; `comment_block_open` does not. So an indented `%%%` is not a fence opener - it is a comment LINE whose content happens to start with `%`, and #624's own rule then says it is consumed wherever it sits. That is carve-js's behaviour, and it makes the body between two of them ordinary text.

carve-rs reads the indented `%%%` as a fence and swallows the body, which the production rules out from the other side (markup-carve/carve-rs#551 tracks that half).

## Why this is worth a separate issue

#624 settled the `%%` case and left `%%%` deciding the opposite way - a comment marker that renders as text. The exclusion looks like it was there to avoid carve-rs's swallowing behaviour, but the two are different questions: whether an indented `%%%` OPENS A BLOCK (no, per the production) and whether it is CONSUMED AS A LINE (yes, per #624's own rule).

No corpus case covers it, which is why the three engines could split.

dereuromark:
Measured the whole family across the three renderers today: carve-rs main (b4c894e), carve-js with markup-carve/carve-js#632 applied, and `scripts/spec` at main (3cdd681). The spec is the outlier on every shape where the fence is indented INSIDE a list item, and it agrees with the engines at top level.

Inside a list item, below the content column:

```
- a
 %%% n
 x
 %%%
 tail
```

| | output |
|---|---|
| carve-rs | `<li>a` + `tail` |
[...]

PR #634, opened 2026-08-04
spec: a comment fence is a comment at any column, and a floating attribute stops at the item boundary

dereuromark:
Two spec changes that share one comparison run, so the count docs are re-quoted once rather than twice.

## A comment fence is a comment at any column (markup-carve/carve#629)

markup-carve/carve#624 exempted the `%%` LINE form from §24 C3's below-column fold and left the `%%%` fence folding with everything else, so

```
- a
 %%% n
```

rendered the opener as **visible text** where carve-js and carve-rs drop it. Same rule, one delimiter over.

`COMMENT_LINE`'s `%%` prefix already matched both forms, so the exclusion was the whole of the bug. The fence's body lines are indented with the opener and follow it in, which is what makes the enclosed content invisible too.

Worth recording what the before/after showed: the executable spec used to render the **whole** comment - opener, body and closer - as item text, an answer no implementation agreed with. It now matches carve-rs.

Measuring that across the engines surfaced [carve-js#630](https://github.com/markup-carve/carve-js/issues/630): carve-js drops both delimiters but keeps the body, so a comment renders its own contents.

## A floating attribute stops at the item boundary (markup-carve/carve#631)

§15 A2a floats a pending attribute past what renders nothing and attaches it to the next VISIBLE block. All four implementations read an item boundary as ending that scope, so

```
- a

  {.c}
- b
```

gives neither paragraph the class - and nothing pinned it. Re-measured all four before adding the case; still unanimous, so this is pure regression insurance.

carve-js#620 filed the shape as a divergence when carve-php's then-unmerged branch attached the class to the second item. That never reached main, so the disagreement resolved itself and took with it the only thing that would catch a regression: agreement is not a check.

## Comparison

carve-rs now passes every fixture case. carve-js fails two and carve-php one, each a rule pinned before the engine moved - recorded on the page rather than left to be rediscovered.


dereuromark:
Sequencing note, since corpus `186-a-comment-fence-is-a-comment-at-any-column-too` pins exactly the shape carve-php still disagrees on.

Re-measured after today's engine fixes:

| | `- a` / ` %%% n` / ` x` / ` %%%` / ` tail` |
|---|---|
| carve-js (markup-carve/carve-js#632, merged) | `<li>a` + `tail` - matches this PR's fixture |
| carve-rs (markup-carve/carve-rs#574, merged) | `<li>a` + `tail` - matches |
| **carve-php** | `<li>a` + `%%% n` + `x` + `%%%` + `tail` - all text |

So the fixture is right and carve-php is the one engine that will fail it. Merging this before carve-php implements the rule turns the next spec-corpus bump in that repo red, which is the failure mode markup-carve/carve#443's ordering rule exists for: land the engine change first, then move the pin.

Filed as markup-carve/carve-php#770 (second half). The first half of that issue - an indented fence rendering its BODY under a top-level paragraph - was a separate defect present in all three engines and is fixed in each: carve-js#632, carve-rs#574, carve-php#773. Worth knowing while reading this PR's before/after, because that leak is why the earlier comparison showed carve-js "dropping both delimiters but keeping the body".

The other half of this PR (the floating attribute at an item boundary) has no such dependency - all four already agree.

PR #638, opened 2026-08-04
spec: pin trailing reference-definition attributes with corpus cases

### LINKED ISSUE #604: A2a left reference definitions with no way to carry attributes, and the grammar has no slot for one
§15 A2a settled where a floating attribute goes, and as a side effect removed the only spelling any engine had for attributing a reference definition. Nothing replaced it, and the grammar has no slot to replace it with.

## What A2a decided

Pending attributes float PAST a construct that renders nothing and attach to the next VISIBLE block:

```
{#i}
[f]: u

e
```

is `<p id="i">e</p>`. Right, and not in question here.

## What it cost

carve-php read that same line as attributing the DEFINITION, so every link resolving the label carried the attributes:

```
[Example][ex]

{.external}
[ex]: https://example.com
```

gave `<a href="https://example.com" class="external">`. Under A2a it cannot, and markup-carve/carve-php#719 removes it there. carve-js and carve-rs never had it.

## The replacement, measured

All four - carve-js (c42bbce), carve-rs (3b2d034), carve-php (main), and `scripts/spec` - agree on every row:

| spelling | result |
| --- | --- |
| `[Example][ex]{.external}` at the link | `class="external"` on the link |
| `[ex]: https://e.com "My Title"` | `title="My Title"` on the link |
| `[ex]: https://e.com {.external}` | attributes ignored |

So a TITLE has a definition-level spelling and everything else does not. Attributing a definition means repeating the attribute at every reference: one definition used ten times becomes ten `{.external}`. That is the ergonomic gap - the definition exists precisely so a destination is written once.

## The gap in the grammar

```
reference_definition = '[', reference_label, ']', ':', space, link_destination, [link_title], newline ;
```

`link_title` is there; there is no attribute slot. So the trailing form is not "unimplemented", it is unspecified - which is why all three engines ignore it and none is wrong to.

## What would close it

A trailing `[attributes]` in the production, applied to every link that resolves the label, with link-site attributes winning on conflict:

```
[ex]: https://e.com {.external}
```

Two things worth deciding with it, not after:

1. **Where they live in the AST.** There is no `link_reference_definition` node - PART 12 §3a and §10a both record it as open, on other grounds. Definition-level attributes would be a third reason the same node keeps coming up: today they can only be materialized onto each resolved link, which is what carve-php did.
2. **Precedence.** carve-php's removed implementation merged definition attributes first and let the link's own override, and its tests pinned that. If the slot lands, that ordering is the obvious candidate but should be stated rather than inherited.

## Not urgent, and not nothing

No corpus document uses the shape (checked: zero attribute lines above a definition across the corpus), and the two engines that never had it are unaffected. But carve-php shipped it, so this is a capability the ecosystem had in one engine and now has in none - worth an explicit yes or no rather than silence.

Filed from markup-carve/carve-php#702 item 2. markup-carve/carve-php#719 (the conformance fix) is in draft until this is answered, so carve-php does not remove the feature and then regain it under different syntax.

dereuromark:
Pins the behaviour PART 9R R1 describes for trailing attributes on a link reference definition. Nothing in the corpus said any of it. Fixes markup-carve/carve#604's corpus half.

Rebased onto current `main` - the original branch was `CONFLICTING` against a base that had moved a long way, and rather than splice conflicts through a fenced block I reset to `main` and re-applied the section. All three examples re-verified against carve-js `c3aa467`, carve-rs `ad865aa` and carve-php `aa4a900` before touching anything; all four implementations already agree on each.

## What is pinned

A definition's attributes reach every link that resolves the label:

```
[Example][ex] and [again][ex]

[ex]: https://example.com {.external}
```

```html
<p><a href="https://example.com" class="external">Example</a> and <a href="https://example.com" class="external">again</a></p>
```

A link's own list stacks over the definition's, per PART 9 §15 A3 - definition first, link second, so a repeated key takes the LAST value and classes accumulate:

```
[Example][ex]{.internal #b}

[ex]: /u {.external #a}
```

```html
<p><a href="/u" class="external internal" id="b">Example</a></p>
```

And a `{...}` on its own line ABOVE a definition is a different construct - it floats PAST the definition to the next visible block (§15 A2a). Both can appear at once and do different things:

```
{.a}
[ex]: /u {.b}

[E][ex] and text
```

```html
<p class="a"><a href="/u" class="b">E</a> and text</p>
```

## The comparison snapshot

Moves to 583 pairs. Its cross-engine diff count jumps from 5 to 32, which is one defect surfacing rather than a coverage regression, and the page now says so.

carve-js stopped inflating a nested list (markup-carve/carve-js#653): each level had been indented twice, once by an absolute depth term and again by the parent item's continuation prefix, so `fmt` returned O(depth^3) bytes for an O(depth^2) source. carve-rs and carve-php still do, identically - 130 bytes in, 460 out at depth 10 - so fixing the third engine left those two as outliers on every corpus case containing a nested list. Filed as markup-carve/carve-rs#594, markup-carve/carve-php#792 and markup-carve/carve#662.

All 32 are on the `carve` target. `html`, `markdown`, `plain` and `ansi` agree everywhere, and every engine renders every fixture case: 583 of 583, zero mismatches, zero errors. Idempotence and the round trip hold in all three, so the inflation is a fidelity and size defect, not a §1 violation.

### DIFF
--- a/docs/examples/edge-cases.md
+++ b/docs/examples/edge-cases.md
@@ -7998,3 +7998,61 @@ An indented `>` that reaches no content column is not a container at all: the li
 ```
 
 :::
+
+## Trailing attributes on a link reference definition
+
+A `{...}` block at the end of a definition line attaches to the DEFINITION and
+reaches every link that resolves the label (PART 9R R1). That is what makes a
+definition worth writing once: without it, attributing a destination used ten
+times means repeating the attribute ten times.
+
+::: compare
+
+```carve
+[Example][ex] and [again][ex]
+
+[ex]: https://example.com {.external}
+```
+
+```html
+<p><a href="https://example.com" class="external">Example</a> and <a href="https://example.com" class="external">again</a></p>
+```
+
+:::
+
+The merge is the one stacked attribute lists already use (PART 9 §15 A3): the
+definition's list first, the link's second, so a repeated key takes the LAST
+value and classes ACCUMULATE.
+
+::: compare
+
+```carve
+[Example][ex]{.internal #b}
+
+[ex]: /u {.external #a}
+```
+
+```html
+<p><a href="/u" class="external internal" id="b">Example</a></p>
+```
+
+:::
+
+A `{...}` on its own line ABOVE a definition is a different construct: it floats
+PAST the definition to the next visible block (§15 A2a). Both can appear at
+once, and they do different things.
+
+::: compare
+
+```carve
+{.a}
+[ex]: /u {.b}
+
+[E][ex] and text
+```
+
+```html
+<p class="a"><a href="/u" class="b">E</a> and text</p>
+```
+
+:::
--- a/docs/implementation-comparison.md
+++ b/docs/implementation-comparison.md
@@ -13,45 +13,54 @@ implementation exposes.
 > nothing across rows; the counts are the point, and
 > `tests/implementation-comparison-counts.test.mjs` fails when they stop
 > matching the corpus - which is how this page came to quote 302 pairs against a
-> corpus of 529, and again at 531, 532, 533, 535, 536, 539, 542, 544, 547, 548, 550, 552, 553, 554, 557, 562, 564, 567 and 571.
+> corpus of 529, and again at 531, 532, 533, 535, 536, 539, 542, 544, 547, 548, 550, 552, 553, 554, 557, 562, 564, 567, 571 and 580.
 
 <div class="impl-summary-grid">
   <div class="impl-summary-card">
-    <strong>580 / 580</strong>
+    <strong>583 / 583</strong>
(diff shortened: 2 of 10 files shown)


dereuromark:
Independently reproduced the three shapes on current engine mains, since the case for pinning rests on them already agreeing:

| shape | all three engines |
|---|---|
| `[r]: /u {.a}` used twice | `<a href="/u" class="a">` on BOTH links |
| `[r]: /u {.a #i}` + `[r][]{.b}` | `class="a b" id="i"` - classes accumulate, and the site's own key would win |
| `{#top}` above a definition, then a paragraph | `<p id="top">` - the floater passes the definition |

Byte-identical across carve-js, carve-rs and carve-php. Nothing to flag; the premise holds.

Issue #515, opened 2026-08-03
Unterminated fence inside a ::: body: fence wins or div wins? Engines split 2-1 and nothing pins it


dereuromark:
When a code fence inside a `:::` body has **no closer**, both the fence and the div are unterminated, and something has to give. The engines split 2-1, and the spec does not say.

Input:

````
::: note
```
x
:::
after
````

| engine | resolves in favor of | output |
| --- | --- | --- |
| carve-js | the fence | code block holds `x`, `:::` **and** `after`; the aside closes at EOF |
| carve-rs | the fence | same |
| carve-php | the div | code block holds `x`; the aside closes at the `:::`; `after` is a top-level paragraph |

carve-js and carve-rs:

```html
<aside class="admonition note">
  <pre><code>x
:::
after
</code></pre>
</aside>
```

carve-php:

```html
<aside class="admonition note">
  <pre><code>x
</code></pre>
</aside>
<p>after</p>
```

## This is not the literal-`:::`-in-a-fence case

That one is settled and all three agree. With the fence **closed**:

`````
::: note
````
:::
````
after
:::
`````

every engine renders `:::` as literal code and keeps `after` inside the aside. Being able to show Carve syntax in a code block is not what is at stake here.

## Recommendation: adopt carve-php's rule

An unterminated fence is not an opaque span, so the div's `:::` closes the div.

1. **It is what I4 already says one level up.** PART 9 §10 I4's whole principle is that a fence opener with no closer is not a fence - it cannot interrupt a paragraph. Opacity should follow the same test. carve-php's `appendOpaqueColonFenceSpan` already gates its raw-fence branch on a closer lookahead; markup-carve/carve-php#644 just gave its code-fence branch the same gate, which is where this difference comes from.

2. **It bounds the blast radius.** Under the carve-js/carve-rs rule, one unclosed fence anywhere inside a div consumes the remainder of the document. `after` is text that visibly follows the closing `:::` and ends up inside a container it was never written in.

3. **It costs nothing.** The closed case above is unaffected, and a fence you actually intended to be a code block has a closer.

The counter-argument is that a fence should be opaque unconditionally, and that scanning ahead for a closer before deciding is a second pass. That pass already exists - every engine does the lookahead for I4, and two of the three do it for the raw-fence branch of the div body collector too.

## Whatever is decided, it needs a corpus case

[...]

Issue #494, opened 2026-08-03
§25 does not say how over-cap openers group, and the three engines each chose differently


dereuromark:
PART 9 §25 says what happens past the nesting cap:

> beyond which further openers degrade to literal text rather than recursing … Past the cap every container kind FLATTENS the same way -- the opener becomes literal paragraph text -- rather than crashing.

It does not say how the flattened lines **group**. Three engines, three answers, for `::: note` x 203 followed by `x`:

| engine | tail |
| --- | --- |
| carve-js | one paragraph per over-cap opener - `<p>::: note</p>`, `<p>::: note</p>`, then `<p>::: note\nx</p>` |
| carve-php | ONE paragraph holding all of them, with a trailing newline before `</p>` |
| carve-rs | ONE paragraph holding all of them, no trailing newline |

All three satisfy "the opener becomes literal paragraph text" as written. None of them agrees with another byte for byte.

## Why nobody noticed

No corpus document reaches the cap. Every gate the project runs compares HTML over the corpus, so a path with no corpus case is a path with no comparison - and `compare:impls`, which would have caught it across engines, only started running in CI in #493 and also only reads the corpus.

That is the more interesting half: the flattening rule is one of the few in §25 whose *observable output* is unspecified, and the reason it stayed unspecified is that the only documents anyone measures are ones that never trigger it.

## What needs deciding

1. **Grouping.** Do consecutive over-cap openers form one paragraph or one each? "The opener becomes literal paragraph text" reads singular, which argues for one paragraph per opener (carve-js); but consecutive non-blank lines are one paragraph everywhere else in the language, which argues for one (carve-php, carve-rs). The second is more consistent with the rest of the grammar; the first is what the reference implementation does.
2. **The trailing newline** before `</p>` - carve-php includes it, the other two do not. Almost certainly an accident of how the paragraph is reconstructed rather than a decision, but it is a byte difference between engines and should be settled with the grouping.
3. **A corpus document**, so whatever is decided is measured. Without one this reopens the moment someone touches a degrade path.

Related: carve-rs#449 stops carve-rs discarding the over-cap openers entirely (it emitted 200 asides and nothing else, byte-identical output whether 5 or 7800 openers were past the cap). That PR deliberately does not pick a shape.

dereuromark:
Re-measured against all three engines today (after carve-rs#449 landed, which is what made carve-rs emit these lines at all). The split is unchanged. For `:::: note` x 203 followed by `x`, the over-cap tail:

[...]

PR #641, opened 2026-08-04
spec: an image reference resolves the same definition entry a link does

dereuromark:
Decides markup-carve/carve#604's open follow-up: definition attributes reach an **image** reference, not only a link.

## The rule that could not be stated

PART 9R's symbol table is `label -> (url, title?, attrs?)`, and `![alt][ex]` looks the label up in exactly that table. Measured before this change:

| `![a][ex]` + `[ex]: /i.png …` | carve-js | carve-rs | carve-php |
| --- | --- | --- | --- |
| `"T"` (title) | `title="T"` | `title="T"` | `title="T"` |
| `{.wide}` (attrs) | — | — | `class="wide"` |

Every engine already took `url` and `title` from that entry; `attrs` crossed in one of three. Two of three fields transferring is not a rule - it is where an implementation stopped, and the alternative cannot be written down without appealing to history. The ergonomics agree: a class repeated at every use site is more common for images (sizing, floating, lazy-loading) than for links, which is the case the definition form exists to serve.

So `[ex]: /i.png {.wide}` now gives `alt`, with the reference site's own attributes overriding per key under §15 A3, exactly as on the link side.

## The bigger half: the executable spec had no image references at all

`![a][ex]` matched `bang` followed by `bracketed`, so it rendered as a literal `!` and an `<a href=…>` - **wrong for every image reference**, attributes or not.

It was invisible because **no corpus document contains an image reference**. I checked every `.crv` for `![...][` and the count is zero. A construct with no corpus case has no comparison, so the one implementation this repository owns has been wrong about it for as long as it has existed. Same shape as #494, #515, and the §25 refusal earlier this week.

- `imageRef` production in `resources/carve-core.ohm`, tried before `image` so the bracket form wins over the paren form for the same prefix
- semantics emitting the same ref sentinel a reference link does, flagged so resolution builds an ``
- `resolveImageRef` in `scripts/spec/html.mjs`, taking url, title and attrs, with an unresolved reference staying literal source (PART 12 §3a keeps the construct rather than discarding it)
- the PART 10 standalone-image unwrap re-applied **after** resolution: the inline form is caught at paragraph-render time, but a reference is still a sentinel then and only becomes an `` once refs resolve

## Verified

Against carve-js `main` on five shapes - bare, with a title, with site attributes, mid-paragraph, and unresolved - all byte-identical.

Spec suite: 795 passing, 4 skipped.

## Deliberately not here

No corpus case and no engine change. carve-js and carve-rs still drop definition attributes on an image reference, so a fixture would fail their gates the day it landed. Engines next, corpus last - the ordering §25 already uses.


dereuromark:
Independently reproduced the premise on current engine mains, since the whole argument rests on that table being right:

`![a][ex]` with `[ex]: /i.png {.wide}`

| | output |
|---|---|
| carve-js | `a` |
| carve-rs | `a` |
| carve-php | `a` |

`[a][ex]` with `[ex]: /u {.wide}` - all three give `<a href="/u" class="wide">a</a>`.

So it is exactly one of three on the image side and three of three on the link side, from the same table entry. That does make "two of the three fields transfer" hard to state as a rule without appealing to which implementation stopped where.

After this lands, carve-js and carve-rs both move.

dereuromark:
This is **12 commits behind** and no longer mergeable as-is - `gh pr merge` refuses it, and a local rebase conflicts in `scripts/spec/html.mjs`.

Its green checks describe the base it was opened against, five hours and a dozen merges ago. Several of those touched the same oracle (#645 the lazy frame, #656 collapsed-reference keying, #663 column-0 definitions), so the conflict is real rather than incidental - the reference renderer moved under it.

Two things worth knowing before it is rebased:

- The body cites a `compare:impls` run at **561** corpus pairs. The corpus is at **583** now, so those numbers need regenerating along with the counts in `docs/index.md` and the comparison page - the count guard will catch it, but only after the rebase.
- **The premise still holds.** I re-measured on current engine mains: `![a][ex]` with `[ex]: /i.png {.wide}` gives `class="wide"` in carve-php only, and the link form gives it in all three. So it is still exactly one-of-three on the image side and three-of-three on the link side, which is the argument the PR rests on.

Nothing here needs a decision - it just needs a rebase and a fresh comparison run.

dereuromark:
Rebased onto main (was 17 behind, conflicting) and pushed. Two review findings came out of it, one fixed here and one older than this branch.

## The conflict

`scripts/spec/html.mjs`, in `resolveRefs`. main had added `source` to the destructure for carve#648 (a collapsed reference is keyed by the label AS WRITTEN, not by the rendered text); this branch had added the `img` dispatch to the same line. Both kept.

That is not only textual. `resolveImageRef` keyed on `label ?? alt`, which is the pre-#648 rule. The sentinel carries `alt: alt.sourceString`, so it was already the source string rather than rendered text, and the two paths only differed in that the link side collapses whitespace. Now they match, which is what this branch is for:

```js
const key = label ?? alt.trim().replace(/\s+/g, ' ')
```

## Fixed here: a standalone image reference lost its paragraph's attributes

The unwrap that turns an image-only paragraph into a bare `` runs twice - once at paragraph-render time, and once after reference resolution, since an image reference is still a sentinel on the first pass. Only the first carried the block attributes.

```
{.block}
![alt][x]

[x]: /u
```

| | before | after |
|---|---|---|
| oracle | `<p class="block">alt</p>` | `alt` |
| all three engines | `alt` | unchanged |

The direct form `{.block}` / `alt` already unwrapped, so this branch was introducing an inconsistency between the two spellings of the same construct.

## Not fixed here: an unresolved reference drops its trailing attributes

```
[text][missing]{.wide}
```

All three engines render `<p>[text][missing]{.wide}</p>`. The oracle renders `<p>[text][missing]</p>` - the literal fallback reconstructs the source from `label` and `text` and never had the attribute text to put back. The image path this branch adds inherits the same shape rather than introducing it, so it is out of scope here; filing separately.

One more, also pre-existing and also out of scope: on a standalone image with block attributes the oracle writes `class` FIRST and all three engines write it last. No corpus case pairs the two, which is why nothing fails.

## Gates

895 tests pass, 0 fail. `core:check` reports 583 of 583 conformant, 0 defects, 0 sentinel leaks.

PR #680, opened 2026-08-05
spec: the §3a rationale contradicts §7 about hoisting

dereuromark:
Reworked after carve#683 landed the C3 wording and its loose-list corpus case - those halves are dropped here, and what remains is what that PR did not touch.

## Two clauses in one file that disagree

§3a's rationale said "§7 hoists `footnote` and `abbreviation_def` to the document". §7 says the opposite two pages later: an `abbreviation_def` is never hoisted, because the construct is only recognized at document level. Fixed to say a footnote hoists and an abbreviation definition is already there by construction.

§7 also named the `abbreviation_def` node in two clauses without ever describing it, so `abbr` and `expansion` existed only in `resources/ast-schema.json` - not visible to a reader of the grammar.

## The third container

The rule names three containers a definition does not define in - a block quote, a list item and a div. Quote and item were pinned; the div was not:

```
:::
*[HTML]: Hyper Text

The HTML spec.
:::
```

```html
<div>
  <p>*[HTML]: Hyper Text</p>
  <p>The HTML spec.</p>
</div>
```

A div renders its children unchanged, so it is the case where the line looks most like a document-level definition. All three engines already agree.

## Two tooling fixes the insertion needed

**Sidecar fixtures did not survive a renumber.** Inserting a case shifts every number after it; the regenerated pair moves and a hand-written `NN-slug.fmt` stays behind - an expected output with no input beside it, pinning nothing. Three had orphaned. `tests/corpus-targets.test.mjs` reports it after the fact; `generate-corpus.mjs` now moves them with their case and prints each move.

**`compare:impls` counted mismatches without naming them.** `rust: pass=594/595 mismatch=1` and nothing more, on a twenty-minute run - finding the document meant re-running every fixture by hand. Same defect `ast:check` had before carve#670. It now prints `mismatch: <target> <case>`, which is how the stray `%%` in carve-rs#602 was located before carve-rs#606 fixed it.

## Numbers

The comparison page and the landing page quote a fresh run over the 585 pairs (carve-rs `0c3e043`, carve-js `cb64471`, carve-php `ed77e6a`): 597 of 597 in every engine, zero cross-implementation differences. Per-file times run high because the machine was loaded - the page already says what those numbers are worth.


dereuromark:
Tightened the C3 clause after checking where else the rule lives: §10 I5 already says an invisible construct - a reference definition, a comment, a block-attribute line - INTERRUPTS an open paragraph, and I6 applies that relation inside a list item like anywhere else.

So the old sentence was not merely under-specified, it contradicted a normative clause two sections away. The new text says so and defers to §10 I5 rather than arguing the point again.

Measured the neighbours while I was there, in a loose item so the paragraph boundary is visible:

```
- a
  [r]: /u
  b

- c
```

```html
<ul>
  <li><p>a</p>
    <p>b</p>
  </li>
  <li><p>c</p></li>
</ul>
```

Same for `[^f]: note`, `{.cls}` (which also lands its class on `b`), the `%%` line and the `%%%` fence - all three engines, all agreeing with §10 I5. The one that does NOT split is `*[HTML]: Hyper Text`, which is correct: inside a container that line is not a definition at all, it is paragraph text, so there is nothing to interrupt with. That is the rule the other half of this PR pins with the div case.

dereuromark:
Reworked and reopened, because carve#683 landed the C3 wording and its loose-list corpus case while this was open - the same two shapes, measured the same way.

What that PR did not touch is now in a branch of its own: the §3a rationale contradicting §7 about hoisting an `abbreviation_def`, the `abbreviation_def` node §7 names but never describes, the div corpus case (quote and list item were pinned, the third container was not), the generator moving sidecar fixtures with their case on a renumber, and `compare:impls` naming the cases an engine mismatches.

Issue #662, opened 2026-08-04
carve fmt inflates a nested list cubically in two engines, and the roundtrip fixture pins that form


dereuromark:
`carve fmt` in carve-rs and carve-php returns a nested list **bigger than it went in**, and the growth is superlinear. carve-js fixed this in markup-carve/carve-js#653; the corpus roundtrip fixture still pins the inflating form, so the fix reads as a divergence.

## Measured

A ladder of `- item N` at 2-space steps, formatted once:

| depth | source | carve-rs | carve-php | carve-js (after js#653) |
|---|---|---|---|---|
| 10 | 180 B | 510 B | 510 B | 180 B |
| 20 | 570 B | 3 230 B | 3 230 B | 570 B |
| 40 | 1 950 B | 23 270 B | 23 270 B | 1 950 B |

Source is O(depth²); the two engines emit O(depth³). js#653 reports 10 KB in, 344 KB out at depth 100, and a 400-deep document going from a 10-minute timeout to 2.2s once the bytes stopped inflating.

The cause it names is worth quoting because it is the same shape in all three: each level is indented twice, once by an absolute `depth` term and again by the parent item's continuation prefix, with a partial strip compensating. The parent's prefix already IS the child's indentation.

## The single-level case, which is what the fixture pins

```
- fruit
  - apples
- vegetables
```

| | output |
|---|---|
| `tests/corpus-roundtrip/03-nested-list.expected.crv` | `    - apples` |
| carve-rs, carve-php | `    - apples` |
| carve-js | `  - apples` |

So today the fixture and two engines agree, and they agree on the behaviour that is cubic at depth. Both spellings re-parse to the same document - the fixture's own guard checks that - so this is a canonical-form question, not a correctness one.

## What needs deciding

Which form PART 11 §2 means. The argument for the js form is that it is the only one whose output is the size of its input, and that the extra indentation is a redundant second application of the parent's prefix rather than a chosen style. If that is the answer, `03-nested-list.expected.crv` moves and carve-rs and carve-php follow.

Worth settling before the next reference-build bump: pinning the current carve-js main makes this fixture fail, which is how it surfaced (markup-carve/carve#661 made the pin honest, and the bump then had something real to say).

## Why nothing caught it

Every check in reach compared HTML or asserted idempotence, and the inflated form is equivalent HTML AND a fixed point. js#653 added a guard on BYTES, which is deterministic and cannot flake; carve-rs and carve-php have no equivalent.

[...]

Issue #626, opened 2026-08-04
The executable spec drops a block-attribute line inside a list item; both engines apply it


dereuromark:
A block-attribute line applies to the block below it at document level, and is dropped inside a list item. Both engines apply it in both places.

```
{.c}
text
```

executable spec: `<p class="c">text</p>` - correct.

```
- a

  {.c}
  text
```

| | output |
|---|---|
| carve-js | `<li><p>a</p><p class="c">text</p></li>` |
| carve-php | the same |
| executable spec | `<li><p>a</p><p>text</p></li>` - **class gone** |

The attribute is not misapplied or misparsed; it is silently discarded, which is the failure mode worth naming: a document that carries styling information through both engines loses it through the reference parse, and nothing reports that.

## Not caused by the recent looseness work

Verified by stashing markup-carve/carve#621 and markup-carve/carve#624 and re-running: the class is dropped on `origin/main` before either. The looseness of this shape is now right (all four agree it is loose); only the attribute is lost.

## Where it probably is

`collectItems` dedents an item's lines and re-parses them as a block sequence, and the attribute-line handling runs in the document-level pass rather than in that recursive one - so the line survives as a block but nothing consumes it as an attribute for the following block.

No corpus case covers it: the corpus has attribute-line cases at document level (`87-block-attribute-lines`) and in-item cases for other constructs, and this intersection is missing. A case wants adding once the parse is fixed, so the fixture and the parser land together.

dereuromark:
My guess at the cause in the description was wrong. The PARSE is fine - the attribute reaches the node:

```
{"t":"para","lines":["text"],"battrs":[[["class","c"]]]}
```

`parseBlocks` handles an attribute line the same way inside an item as at document level, so nothing is lost there.

The renderer drops it. `renderListItem` in `scripts/spec/html.mjs` builds an item's paragraph by hand:

```js
parts.push({ inlineable: true, html: list.tight ? html : `<p>${html}</p>` })
```

That is the one paragraph site in the file that never consults `battrs` - every other one goes through `renderBlockAttrs`. So the attribute survives the parse, survives the tree, and is discarded on the way out.

A tight item renders its paragraph without a `<p>` at all, and there is nowhere to hang attributes; that shape cannot arise, because an attribute line arrives after a blank, and a blank plus a visible paragraph is exactly what makes the item loose.

Fix and a corpus case are on the branch for this.

dereuromark:
[...]

Issue #653, opened 2026-08-04
The canonical writer indents a comment body three ways


dereuromark:
The canonical writer disagrees on how a comment body is indented. Two independent divergences, each 2-1, found by the `carve` target of `npm run compare:impls` on corpus 186 - which has no `carve` fixture, so nothing failed.

## 1. carve-js indents a below-column body one column further

Corpus 186:

```
- a
 %%% n
 x
 %%%
 tail
```

carve-rs and carve-php write:

```
- a
  %%%
  n
  x
  %%%
  tail
```

carve-js writes the same except for the body line, which gets three spaces where the others use two:

```
- a
  %%%
  n
   x
  %%%
  tail
```

The source had ` x` at one column; carve-js appears to carry that column INTO the item's content column rather than replacing it.

## 2. carve-php drops the indentation of a comment body line

```
%%%
  x
%%%
```

carve-rs and carve-js round-trip it unchanged. carve-php writes:

```
%%%
x
%%%
```

Only the FIRST body line is affected - with a line above it, all three agree:

```
%%%
x
  y
%%%
```

is written back unchanged everywhere.

## Why it matters even though the HTML agrees

A comment renders nothing, so every spelling above satisfies `toHtml(fmt(x)) == toHtml(x)` and the html target shows 0 diffs. PART 11 asks for a CANONICAL form, though: two engines writing a document three ways means `carve fmt` output depends on which engine ran it, and a formatted file churns in diffs when the toolchain changes.

Worth deciding which spelling is canonical (preserve the body verbatim, or normalize it to the content column) and pinning it with a `carve` fixture on 186 - the target only compares engines against each other where no fixture exists, which is exactly why this survived.

Both cases also change nothing about the parse: they are writer-only.

dereuromark:
Both divergences are gone. Re-measured on today's mains, each engine rebuilt from source: carve-js `7a4944f`, carve-rs `3c0e9a1`, carve-php `31710ba`.

**1. carve-js indenting a below-column body one column further** - fixed by markup-carve/carve-js#639, "a comment body is indented relative to its fence". The corpus 186 shape now formats identically in all three:

```
- a
  %%%
  n
  x
  %%%
  tail
```

**2. carve-php dropping a comment body's indentation** - fixed by markup-carve/carve-php#782, "a comment body keeps the indentation the author wrote". All three now write:

```
%%%
  x
%%%
```

[...]

PR #695, opened 2026-08-05
fix: the oracle keeps a caption inside a list item, and bump the stale engine pin

### LINKED ISSUE #693: The oracle drops a caption inside a list item, so no corpus case can pin the shape
The executable spec DELETES a `^ ` caption line written inside a list item. Not promoted to a `<figure>`, not rendered as text - absent from the output.

## Measured on `4d078d5`

```
- ![alt](/i.png)
  ^ Figure 1: caption
```

```html
<ul>
  <li><img src="/i.png" alt="alt"></li>
</ul>
```

The caption is gone. Same caption, other containers:

| position | oracle |
|---|---|
| top level | `<figcaption>Figure 1: caption</figcaption>` |
| inside a `:::` div | `<figcaption>Figure 1: caption</figcaption>` |
| **inside a list item** | **nothing at all** |

So the oracle promotes it everywhere except a list item, where it discards it.

## Why this matters beyond the one shape

The engines disagree here too, and this is the row that decides it:

| implementation | output |
|---|---|
| carve-js | `<figure>` + `<figcaption>` |
| carve-php | same |
| carve-rs | `^ Figure 1: caption` as literal text (markup-carve/carve-rs#610) |
| oracle | caption deleted |

carve-rs#610 was filed as an engine gap against the other two. But no corpus case can pin the correct answer while the oracle produces a third one - and its answer is the only one that loses content, which makes it the worst candidate to standardise on.

## Which answer §4 supports

§4 attaches a caption to the immediately preceding captionable block, and a standalone image paragraph inside a list item is still a standalone image paragraph. Nothing in §4 restricts attachment by container, and the oracle itself promotes the same caption at the top level and inside a div. So carve-js and carve-php look right, and both carve-rs and the oracle need to move.

Filed separately from carve-rs#610 because this half needs a corpus case as well as a fix, and because an oracle that silently drops an authored line is worth its own record - the corpus and `formal-core-check` both pass today, since nothing exercises a caption inside a list item.

dereuromark:
Fixes markup-carve/carve#693. Also unblocks CI, which is red on `main` right now - hence two things in one PR.

## The caption

A `^ ` line attaches to the block above it. The oracle did that at the top level and inside a `:::` div, but inside a list ITEM it **deleted** the caption - not promoted, not rendered as text, absent:

```
- ![alt](/i.png)
  ^ Figure 1: caption
```

gave `<li>alt</li>`.

The parse was never wrong - the paragraph node carried `caption` all along. The item renderer builds its `<p>` by hand and never consulted it.

**This is the second field that site has dropped for the same reason.** markup-carve/carve#626 was `battrs` going the same way, and the comment recording it is still there. The fix is the same: delegate a captioned paragraph to `renderBlock`, which owns the figure shape, rather than duplicating the top-level logic. The inline path stays for the plain shape it exists for.

Output now matches carve-js and carve-php byte for byte. carve-rs renders the line as literal text and is tracked separately (markup-carve/carve-rs#610) - with the oracle fixed, a corpus case can finally pin the right answer, which was impossible while the oracle produced a third one.

## The pin, and why it is here

`tests/roundtrip.test.mjs` formats through the PINNED `@markup-carve/carve`, and the pin predated markup-carve/carve-js#653, the nested-list writer fix. So the `.fmt` fixtures added in markup-carve/carve#675 expected the corrected two-space form while the pinned engine still emitted the inflated four-space form:

```
+ actual   '- fruit\n    - apples\n- vegetables\n'
- expected '- fruit\n  - apples\n- vegetables\n'
```

`main` has been failing those two tests since. markup-carve/carve#662 predicted exactly this. Bumped to carve-js `bf8a695`.

I verified those failures reproduce on unmodified `main` before touching anything, so they are not fallout from the caption change - and a caption-only PR could not have been green.

## Verification

One of the seven new assertions fails on the parent commit. The others cover CONTAINERS rather than fields - top level, div, block quote, a plain uncaptioned item, and an orphan `^ ` line that must stay text - because a hand-built path that reads some of a node's fields will keep losing the ones added later. carve#626's `battrs` is re-asserted in the same file for that reason: if the delegation is reverted, both losses return together.

Full suite 917 pass, drift gate clean, `core:check` clean.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
       },
       "devDependencies": {
         "@djot/djot": "^0.3.2",
-        "@markup-carve/carve": "github:markup-carve/carve-js#cb644719152934d97eca037e1d9d3a5a38df315f",
+        "@markup-carve/carve": "github:markup-carve/carve-js#bf8a6958b315f7b000f253bd73c0a9946b669eb2",
         "@markup-carve/carve-grammars": "github:markup-carve/carve-grammars",
         "@markup-carve/vite-plugin-carve": "git+https://github.com/markup-carve/vite-plugin-carve.git",
         "@mermaid-lint/cli": "^0.35.0",
@@ -982,8 +982,8 @@
     },
     "node_modules/@markup-carve/carve": {
       "version": "0.1.2",
-      "resolved": "git+ssh://git@github.com/markup-carve/carve-js.git#cb644719152934d97eca037e1d9d3a5a38df315f",
-      "integrity": "sha512-NRB71uZm9K+CEl0iQfoA+T2Fy3GW6JuTPeWk3yiAZZyfAPUmo2i3/6yCkbGZebN3wwBNIqtnjEZMxcdv9unR/w==",
+      "resolved": "git+ssh://git@github.com/markup-carve/carve-js.git#bf8a6958b315f7b000f253bd73c0a9946b669eb2",
+      "integrity": "sha512-H3izPIhy332zxXf1fl1tYfGd4hGLmnM+ZCI903xVEMH6NnWdei6E+F77IkUzDjjEKhhdhbjrFP993uqUCTERJA==",
       "dev": true,
       "license": "MIT",
       "bin": {
--- a/package.json
+++ b/package.json
@@ -20,13 +20,13 @@
     "bump-carve-pin": "node scripts/bump-carve-pin.mjs",
     "sync-carve-wasm": "node scripts/sync-carve-wasm.mjs",
     "lint:mermaid": "mermaid-lint --ext crv --quiet",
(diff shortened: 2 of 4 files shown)


dereuromark:
Correction to this PR's second half. The caption fix stands; the reasoning I gave for the pin bump was wrong.

I claimed `main` had been failing two `.fmt` fixture tests, and that the pin bump was needed to fix it. It had not, and it was not.

**What I got wrong.** The old pin `cb64471` ALREADY contained markup-carve/carve-js#653, the nested-list writer fix:

```
$ git merge-base --is-ancestor c3aa467 cb64471   # #653 -> old pin
c3aa467 (#653) IS an ancestor of cb64471
```

So the pinned engine emitted the corrected two-space form, and CI on `4d078d5` passed - which I could have checked before writing the claim, and did not. Both `package.json` and `package-lock.json` at that commit pointed at `cb64471`, so CI and my machine were nominally running the same engine.

**What actually happened.** My local `node_modules` in that checkout held an OLDER carve-js than the lockfile specified, so `npm test` locally formatted through a pre-#653 build and produced the inflated four-space form. The two failures were a stale install on my side, not a state of the repository. I have hit this exact trap in this project before, which is the annoying part.

**What that means for the change.** The pin bump `cb64471` -> `bf8a695` is still a legitimate move forward - the drift gate is clean either side of it and the suite passes - but it fixed nothing, and my citing markup-carve/carve#662 as "predicting this" was misapplied. #662 is about the ENGINES' writers, which are all fixed; the oracle's fixture was never affected.

The caption half is unaffected: it is a real defect, one of the seven assertions fails on the parent commit, and the output now matches carve-js and carve-php byte for byte.

Recording it here rather than quietly, because "main was red" is the kind of claim someone will otherwise take at face value when reading this commit later.

PR #728, opened 2026-08-05
spec: the ohm grammar's identStart drops the leading dash

### LINKED ISSUE #722: The ohm grammar admits a dash-first attribute identifier; the normative EBNF and every engine do not
The two spec artifacts disagree about whether an attribute identifier may start with `-`, and the executable spec follows the one the engines do not.

`resources/grammar.ebnf`, PART 7 (normative):

```
identifier = (letter | '_'), {letter | digit | '_' | '-'} ;
```

`resources/carve-core.ohm`, under the comment "Attribute block (PART 4, strict identifiers)":

```
ident      = identStart identRest*
identStart = letter | "_" | "-"
identRest  = letter | digit | "_" | "-"
```

`identStart` admits `-`; the EBNF does not.

## Observable

Four shapes, oracle against carve-js:

| input | carve-js | oracle |
| --- | --- | --- |
| `[x]{--flag}` | `<p>[x]{–flag}</p>` | `<p><span --flag="">x</span></p>` |
| `[x]{#-id}` | `<p>[x]{#-id as a tag}</p>` | `<p><span id="-id">x</span></p>` |
| `-{--flag} item` | `<p>-{–flag} item</p>` | `<ul><li --flag="">item</li></ul>` |
| `1.{#-id} item` | `<p>1.{#-id as a tag}</p>` | `<ol><li id="-id">item</li></ol>` |

DEFECT-bucket in all four: the oracle claims the input and renders something else. `scripts/formal-core-check.mjs` never sees one, because no corpus document writes a dash-first identifier.

`[x]{a:b}` agrees (both literal), so this is specifically the dash, not identifiers in general.

## Two reasons the EBNF looks right

1. All the engines follow it. carve-js renders every shape above as literal text; that is the behavior three implementations already ship.
2. `--flag=""` is not a valid HTML attribute name, so the oracle's output is not just divergent, it is unserializable as HTML. A leading `-` also collides with the CSS custom-property spelling that authors write in `style` values, which is the shape most likely to appear by accident.

So the fix is probably `identStart = letter | "_"` in `carve-core.ohm`, plus whatever the block layer uses - `parseAttrList` in `scripts/spec/render.mjs` accepts it too, since the marker cases above go through the block path rather than the inline grammar.

## Where it came from

Found while porting the marker guard to `intellij-carve`'s annotator (markup-carve/intellij-carve#55). `codex review` flagged the port for rejecting dash-first identifiers and cited the ohm rule as the authority; running both artifacts against the engine is what showed the ohm rule to be the outlier. The port follows the EBNF.

## Also worth a look while here

An unrelated divergence surfaced in the same session, minimized:

```
x !="b"
```

carve-js renders `≠”b”`, the oracle renders `≠“b”` - the quote following a smart-typography replacement flanks differently. Needs no braces and no attributes; happy to split it out if you would rather track it separately.

dereuromark:
Fixes #722.

PART 7 is `identifier = (letter | '_'), {letter | digit | '_' | '-'}`. The ohm grammar's `identStart` also admitted `-`, so the executable spec was the only artifact reading a dash-first identifier as an identifier.

## Measured, before and after

Oracle against carve-js on current mains:

| input | carve-js | oracle before | oracle after |
|---|---|---|---|
| `[x]{--flag}` | `<p>[x]{–flag}</p>` | `<p><span --flag="">x</span></p>` | agrees |
| `[x]{#-id}` | literal, `#-id` as a tag | `<p><span id="-id">x</span></p>` | agrees |
| `-{--flag} item` | `<p>-{–flag} item</p>` | `<ul><li --flag="">item</li></ul>` | agrees |
| `1.{#-id} item` | literal | `<ol><li id="-id">item</li></ol>` | agrees |

`<span --flag="">` is not a serializable HTML attribute name, so the old output was not merely divergent.

## extName shares the rule

`extName = identStart identRest*`, and the EBNF has `extension_name = identifier`, so narrowing `identStart` narrows extension names too. That is correct and now checked rather than assumed: `:-foo[x]` stays literal, which carve-js, carve-rs and carve-php all already do.

## Controls

Only `identStart` narrows, so a dash INSIDE an identifier and a leading underscore are untouched:

| input | result |
|---|---|
| `[x]{a-b=c}` | `<span a-b="c">` in both |
| `[x]{_ok}` | `<span _ok="">` in both |

## Why nothing caught it

No corpus document writes a dash-first identifier, so `formal-core-check` never reached the rule. It stays at 609 of 609 conformant, 0 defects, 0 refused.

One note for whoever writes the corpus case: my first attempt used `--` to comment the rule, which Ohm reads as an alternative's case label rather than a comment. The grammar silently stopped parsing and every shape came back as an ohm error, not as a divergence - a mistake that looks like a much bigger failure than it is.

### DIFF
--- a/resources/carve-core.ohm
+++ b/resources/carve-core.ohm
@@ -322,7 +322,15 @@ CarveCore {
   attrVal    = quoted | squoted | bareVal
   bareVal    = (~(" " | "}" | newline) any)+
   ident      = identStart identRest*
-  identStart = letter | "_" | "-"
+  // PART 7: identifier = (letter | '_'), {letter | digit | '_' | '-'}
+  // A leading "-" is NOT admitted. This file was the only artifact that
+  // allowed one: the normative EBNF does not, and carve-js, carve-rs and
+  // carve-php all render `{--flag}` and `{#-id}` as literal text. The oracle
+  // instead built `<span --flag="">`, which is not a serializable HTML
+  // attribute name (carve#722). `extName` shares this rule, and
+  // `extension_name = identifier` in the EBNF, so `:-foo[x]` stays literal
+  // too - which all three engines already do.
+  identStart = letter | "_"
   identRest  = letter | digit | "_" | "-"
 
   // A word absorbs any bare delimiter glued between two word characters -


dereuromark:
Duplicate of #727, which landed while this was in CI. Closing.

Compared rather than assuming: theirs is the same one-token narrowing (`identStart = letter | "_"`) with equivalent reasoning, and it adds a test file mine did not - covering five shapes plus a raw-format name through `extName`, which is the sixth I had only checked by hand.

Nothing here is worth keeping. The branch is deleted.

PR #753, opened 2026-08-05
spec: render ceiling margin as a multiple

### LINKED ISSUE #650: §25's ceiling 'exceeds the parse cap by construction' only in container-depth units; in AST-level units it truncated legal documents
PART 9 §25 states the render ceiling as a property and then justifies it with a claim about units:

> Refusing costs nothing on any path a document travels: the ceiling exceeds `MAX_NESTING_DEPTH` by construction, so no tree from `parse` can reach it.

"Exceeds by construction" holds only if the renderer counts what the PARSE CAP counts. Two of the three engines count container depth, one per nesting level, and the claim holds for them. carve-rs counts AST levels, where one source level of a list costs two (`list`, then `list_item`) before its body - so `MAX_NESTING_DEPTH + 32` was reachable by an ordinary document.

## Measured

A list ladder, 120 levels, well inside the parse cap of 200:

```
- x
  - x
    - x
```

| | output |
|---|---|
| carve-js | 117001 bytes, complete |
| carve-php | 117001 bytes, complete |
| carve-rs | truncated - the innermost content silently dropped |

At 220 levels (past the parse cap, so the parser's own degradation is in play) carve-js emits 324687 bytes and carve-php 325060, while carve-rs stopped at its ceiling.

Fixed in carve-rs by restating the bound as a factor of the parse cap rather than an offset, which is what makes the §25 property true in AST-level units. But the clause is what let the two readings coexist: an implementer who counts AST levels reads "exceeds by construction" and does not check, because the sentence says there is nothing to check.

## What is worth pinning

The margin is stated as an absolute (`+ 32`) in prose that also fixes the number at 232 for measurement. Either the clause should say the unit the depth is counted in, or - better, since it is the property that matters - the corpus should carry a case: a document at the parse cap renders whole on every target. That is a check an engine cannot pass by accident, and it would have caught this.

Same shape as PART 12 §9(b)'s ingest bound, where `MAX_JSON_DEPTH` had to be written as a multiple of the parser's cap for exactly this reason - the note there already says "THE UNIT IS NOT THE PARSER'S". §25 is the same statement at the other end of the pipe and does not say it.

dereuromark:
Closes #650

## What changed and why

PART 9 §25 justified the render ceiling with "the ceiling exceeds
`MAX_NESTING_DEPTH` by construction, so no tree from `parse` can reach it,"
and fixed the margin as an additive `MAX_NESTING_DEPTH + 32`. That offset was
sized for engines that count *container depth* (one step per nesting level,
matching `MAX_NESTING_DEPTH`'s own unit). An engine that counts *AST node
levels* instead - where a list level costs two node levels (`list`, then
`list_item`) before its body - restated the same absolute number in its own
unit and undercounted the margin by roughly half. An ordinary 120-level list
ladder, well inside the parse cap of 200, reached that engine's render
ceiling and silently dropped the innermost content. carve-rs's engine fix has
already landed elsewhere; this PR is the spec wording and the corpus/test pin
that were still outstanding here.

`resources/grammar.ebnf` (PART 9 §25): restates the margin as a multiple of
`MAX_NESTING_DEPTH`, derived per implementation from the worst per-level cost
of its own counting unit, rather than a fixed offset borrowed from another
implementation - mirroring PART 12 §9(c)'s existing ingest-bound trap ("the
unit is not the parser's") at the render end of the same pipe. The "exceeds
by construction" claim is now scoped explicitly to a ceiling *derived* this
way, not to any single borrowed number.

`tests/render-ceiling-clears-the-parse-cap.test.mjs` (new, registered in
`package.json`'s `test` script): pins the property the ticket asked for - a
document at `MAX_NESTING_DEPTH` (the parse cap) renders whole, not
truncated, on every render target (`renderHtml`, `renderMarkdown`,
`renderCarve`, `renderPlainText`, `renderAnsi`). Uses blockquote nesting
rather than list nesting for the fixture: §25 applies `MAX_NESTING_DEPTH`
uniformly across container kinds, so blockquote depth is an equally valid
witness of "at the parse cap," and it sidesteps an unrelated performance
issue found while building this - list nesting past roughly depth 50 does
not return in practical time from this engine's `carveToMarkdown` today.
That is worth its own report upstream in carve-js; it is not this ticket's
defect and is out of scope here. A corpus fixture (`tests/corpus/*.crv`)
couldn't carry this case either, for the same size reason already documented
in `tests/nesting-cap-degrade.test.mjs`.

## Draft-conflict warning

The draft guard (`autopilot-gh.sh draft-check`) returned WARN (exit 2) for
two open draft PRs that also touch `docs/security.md`:

- `#444` - docs: make the source migration-clean for Carve (migration prep)
- `#291` - spec: file inclusion / transclusion (PART 9 section 19)

This PR does not touch `docs/security.md` in the end (the substantive clause
lives in `resources/grammar.ebnf`, not that page), so there is no overlap in
practice - noting the warning here regardless, since it named both drafts as
needing a rebase against files this plan was scoped to touch.

## Verification

Baseline gate (before any change) and the gate re-run after the change both
came back `partial`, with the same two gates unrunnable both times:

- `npm run ast:check` - unrunnable: "a dependency outside the repository is
  missing: `/tmp/carve-js/dist`" (needs a built sibling carve-js checkout,
  not part of this repo's per-PR CI).
- `npm run claims:check` - unrunnable: "engine-claims: need all three
  engines, found 0 (none)."

Gates that did run, both before and after, all green with zero failures:
`npm test` (1013 tests, 1012 pass / 1 skipped / 0 failed - the skip predates
this change), `npm run core:check` (610/610 corpus inputs executable-spec
conformant, 0 defects), `npm run test:conformance`.

`node --test tests/normativity.test.mjs` passes, including the clause-count
inventory check and the PART 12 cross-reference check - this PR only edits
prose inside existing numbered clauses, so no clause was added, removed, or
renumbered.


dereuromark:
The diagnosis is right and the §9(c) parallel is the correct frame - the unit, not the number, is what does not travel. One tension in the new clause, though, and it is with this repo's own reference.

Measured what the three engines actually state:

| engine | expression | value | form |
|---|---|---|---|
| carve-rs | `MAX_NESTING_DEPTH * 3 + 32` | 632 | a multiple |
| carve-js | `MAX_NESTING_DEPTH + 32` | 232 | a fixed offset |
| carve-php | `MAX_NESTING_DEPTH + 32` | 232 | a fixed offset |

The clause says:

> An implementation MUST THEREFORE STATE its margin as A MULTIPLE OF MAX_NESTING_DEPTH derived from the worst per-level cost of ITS OWN unit ... not adopt another implementation's absolute number.

and, two sentences earlier, that the reference "uses `MAX_NESTING_DEPTH + 32` = 232, a fixed offset that happens to be a 1.16x multiple in ITS unit."

So the paragraph states a MUST and then records the reference not meeting it, in the same breath. On the day this merges, carve-js and carve-php are non-conformant to the letter of a rule about how they spell a constant - while being perfectly correct in substance, because 1.16x genuinely is enough for the unit they count in.

Two ways out, and I do not think it matters much which:

1. **Make the MUST about the derivation, not the spelling.** What actually has to be true is that the margin covers the worst per-level cost of the implementation's own unit, and that it was derived rather than borrowed. `+ 32` derived for container depth satisfies that; `+ 32` copied into an AST-node-counting engine does not, which is exactly the bug. This keeps all three conformant today and still catches carve#650.
2. **Keep the MUST as written** and follow up in carve-js and carve-php to restate their constants as multiples. Same values, different spelling.

Option 1 looks better to me: the defect was never the arithmetic form, it was adopting a number sized in someone else's unit. A rule aimed at the spelling can be satisfied by writing `MAX_NESTING_DEPTH * 1.16` while still not having thought about the unit.

Worth noting that carve-php's own comment already says "What the number SHOULD be across engines is still open", so whichever way this lands, that comment wants updating to point at the resolved clause.

Not merging on this one - a new normative MUST that two engines do not meet as written seems like your call rather than mine. Everything else in the PR reads correct to me, and CI is green.

Issue #748, opened 2026-08-05
Each engine serializes a different subset of the resolution results §5 requires


dereuromark:
PART 12 §5 says resolution results a consumer cannot recompute ARE serialized, and names two. Each engine publishes a different one, and one publishes neither.

## Measured

carve-js `2a865a7`, carve-rs `d7f46a0`, carve-php `73f21a2`, each rebuilt after the fetch. Read through `carveToAstJson` / `--ast` / `--json`, not the internal tree.

| field | schema | carve-js | carve-rs | carve-php |
|---|---|---|---|---|
| `footnote_ref.number` | "Assigned during resolution, by document reference order" | **published** | absent | absent |
| `caption_number.n` | "A resolution result, serialized because recomputing it means reimplementing PART 9R" | absent | **published** | absent |
| generated heading id (`attrs.id`) | - | **published** | absent | absent |

Inputs:

```
[^a]: note

see[^a] again[^a]
```

```
![x](/i.png)
^ Figure #: one
```

carve-js gives both references `"number":1` - correct, a repeat reference reuses the number. carve-rs gives the caption node `"n":1`. Neither field appears in carve-php at all.

## The rule

> **§5 WHAT IS NOT IN THE SERIALIZED FORM.** Resolution results that a consumer can recompute - footnote numbering, caption numbers - ARE serialized, because recomputing them requires reimplementing PART 9R.

Both fields are in `resources/ast-schema.json` with descriptions saying exactly that. They are optional in the schema, which is why nothing fails: a tree omitting them validates.

The generated heading id is the same class - a consumer would have to reimplement the slug algorithm, its collision counters and the ASCII/lowercase options - but unlike the other two it is not named anywhere, so whether it is required is genuinely open. It accounts for **41 of 610 corpus documents** differing between carve-js and each of the other two.

## Why nothing caught it

`scripts/ast-conformance.mjs` compares every engine against carve-js as the reference (markup-carve/carve#747). A field carve-js publishes and the others do not cannot be reported as their gap, and a field carve-rs publishes and carve-js does not cannot be reported as carve-js's. The check is blind in both directions.

The schema cannot catch it either, since both fields are optional - and they have to be, because they are absent before resolution.

I filed the heading id as a carve-js defect first (markup-carve/carve-js#697) on a §3a reading, and withdrew it: §3a forbids folding a result **in place of** an authored construct, which is not what any of these do. §5 is the clause that governs, and it points the other way.

## What would settle it

[...]

PR #814, opened 2026-08-05
spec: a generated heading id is a serialized resolution result

### LINKED ISSUE #750: A generated heading id is uncomputable from the tree, and only one engine publishes it
Split out of #748, which tracks the two fields §5 names by name. This is the third row of that table and a different argument, so it should not ride along with them.

## Measured

carve-js `2a865a7`, carve-rs at markup-carve/carve-rs#639's head, carve-php `0586fa5`, each rebuilt after the fetch.

Input:

```
# Welcome Home

text
```

All three render the same HTML:

```html
<section id="Welcome-Home">
```

The serialized trees do not agree:

| engine | `heading.attrs` |
|---|---|
| carve-js | `{"id":"Welcome-Home"}` |
| carve-rs | `null` |
| carve-php | `null` |

## Why this is not "the consumer can just slug the text"

The id is not a function of the heading. Input:

```
# Notes

a

# Notes

b
```

Both engines number the collision:

```html
<section id="Notes">
<section id="Notes-2">
```

So computing the second heading's id requires having seen every heading before it and replicating the dedup rule, in the same order, with the same suffix spelling. That is the same shape of argument §5 makes for caption numbers - "recomputing them requires reimplementing PART 9R" - and it is stronger here, because the value depends on the rest of the document rather than on a counter a consumer could keep itself.

A consumer that wants to emit a link to a heading (a table of contents, an LSP go-to-definition, a cross-document index) has to have the exact id. Today two of the three engines make that impossible from the tree alone.

## The counter-argument, stated fairly

§3a describes the serialized tree as PRE-resolve, and the id is produced during resolution. The clean reading is that §5 carves out exactly the values it lists, and heading ids are not listed. Under that reading carve-js is the outlier and should stop publishing it.

I filed and withdrew markup-carve/carve-js#697 on that reading a few days ago. Withdrawn because §5's own words are about recomputability, not about which stage assigns the value - and by that test the id qualifies. But the clause does not name it, so this is a decision, not a defect report.

## What to decide

1. **Name it in §5 and add `heading.attrs.id` to the schema's description as a resolution result**, then land producers in carve-rs and carve-php. carve-js needs no change.
2. **Say §5's list is exhaustive**, and carve-js stops publishing it.

Preference is 1, for the dedup reason above: option 2 leaves a value that is genuinely uncomputable from the tree, and the tree is what PART 12 exists to make sufficient.

Whichever is chosen, the schema should say so explicitly - `attrs` is a free-form map, so today neither behavior is a violation and nothing can catch the divergence.

dereuromark:
Closes #750

PART 12 §5 names the resolution results that ARE serialized and listed two: footnote
numbering and caption numbers. A generated heading id meets §5's own criterion -
recomputability - and meets it more strongly than either.

Dedup assigns the next free suffix in document order:

```
# Notes

a

# Notes

b
```

```html
<section id="Notes">
<section id="Notes-2">
```

So deriving the second id means having seen every heading before it and replicating the
dedup spelling, in the same order. A consumer holding only the tree cannot do that, and a
table of contents, an LSP go-to-definition and a cross-document index each need the exact
id. A footnote number, by contrast, only needs a counter the consumer could keep itself.

## What changed

- `resources/grammar.ebnf` PART 12 §5: the value joins the list, and a named
  `A GENERATED HEADING ID IS A RESOLUTION RESULT -- NORMATIVE` sub-clause carries the
  argument, the §3a reconciliation (the id is published beside the heading's authored
  attributes, not in place of them) and the current engine state.
- `resources/normative-clauses.txt`: regenerated, so the clause cannot be deleted silently.
- `resources/ast-schema.json`: `heading.attrs` gains the description, in the voice
  `caption_number.n` and `footnote_ref.number` already use.
- `docs/ast-json.md`: the "What is not in it" restatement of §5 follows.

## Sequencing against the engines

The wire-shape contract moves without invalidating any shipping engine, so this does not
need to wait behind an engine PR:

- `heading.attrs.id` is an OPTIONAL property of an optional `attrs` slot. A tree that omits
  it validates against the schema before and after; only a `description` is added.
- The divergence is already declared in `resources/ast-value-divergence.txt`
  (`heading.attrs.id 45 ... (carve#750)`), so `npm run ast:check` gates it today and its
  verdict does not move. Whichever engine PR closes the gap last deletes that line.

This is the order #748 used for the two fields §5 already names: clause and schema
description first, producers after (carve-rs#639, carve-php#846, carve-php#847). The
carve-rs and carve-php producers are filed as follow-ups inheriting this ticket's sign-off;
carve-js already publishes the field.

## Verification

- `gate-run`: `partial`. Green: `npm test`, `npm run combinatorial:check`,
  `npm run core:check`, `npm run property:check`, `npm run test:conformance`. Not runnable
  on this host, and not part of this repo's per-PR CI: `npm run ast:check`
  ("a dependency outside the repository is missing: /tmp/carve-js/dist") and
  `npm run claims:check` ("engine-claims: need all three engines, found 0 (none)").
- The inventory line was proved load-bearing: deleting it fails
  `tests/normativity.test.mjs` with "A GENERATED HEADING ID IS A RESOLUTION RESULT: grammar
  has 1, inventory says 0", and restoring it passes.

Draft PR #291 (file inclusion / transclusion) also edits `resources/grammar.ebnf`; expect a
rebase there once it lands.


dereuromark:
## Implementing this faithfully rewrites the author's document, and carve-js already shows it

I implemented the stamping pass in carve-php to match what carve-js publishes, and its AST round-trip tests failed on 41 corpus documents. Chasing that found a gap in the wire format rather than a bug in either engine's pass.

**The wire cannot say an id was GENERATED.** The distinction lives in `attrs.order`: an authored id records the `#id` slot, a generated one does not. carve-js publishes exactly that -

```
# Notes            ->  {"id":"Notes"}                       (generated, no order)
{#custom} # Notes  ->  {"id":"custom","order":["#id"]}      (authored)
```

- but `order` is also **absent on a programmatically built tree**, which the schema says in as many words. So a reader seeing `{"id":"Notes"}` cannot tell "the producer generated this" from "the producer built this tree by hand", and the only safe reading of an id with no order is that it is an attribute of the node.

**What that costs, measured on carve-js `main` today.** Take a document with no authored ids at all, serialize it, read it back, and write it out as Carve:

```
# Notes

# Notes
```

```js
const j = carveToAstJson(src)
const doc = fromAstJson(j)
renderCarve(doc)
```

```
{#Notes}
# Notes

{#Notes-2}
# Notes
```

Two attribute lines the author never typed. Any editor round-tripping a document through the published AST rewrites every heading in it. The HTML is unchanged, so PART 11 §1's invariant still holds and nothing goes red - which is why this is not already known.

carve-php's `AstCodecTest` catches it (`testEveryCorpusDocumentEncodesIdenticallyAfterARoundTrip`, 39 documents, plus 2 that lose HTML through a different attribute-order interaction). carve-js has no equivalent assertion, so it ships the behavior.

**So this is a decision, not an implementation task**, and I have left carve-rs#675 and carve-php#897 open rather than porting a defect into two more engines. The options as I see them:

1. **Mark it on the wire.** A generated id gets a flag, or a slot spelling that is not `#id`, so ingest can keep it out of the authored set. New surface in the schema, but it makes the distinction expressible rather than inferred.
2. **Let ingest treat it as authored, and say so.** Simple, and honest about what the wire carries - but then a tree that has been through the AST is not the document the author wrote, and a formatter has no way back.
3. **Do not serialize a generated id.** Reverts #814. The recomputability argument for serializing it is sound, so this is the weakest option, but it is the only one that costs nothing.

Worth noting the asymmetry with the other two §5 results: a footnote number and a caption number are published on nodes (`footnote_ref.n`, `caption_number.n`) that have no authored spelling to be confused with. A heading id is the first §5 result that lands in a slot the author can also write, which is exactly why it needs a marker the others did not.

dereuromark:
## Answered by implementation: the wire already carries the distinction

I raised three options above - mark it on the wire, let ingest treat it as authored, or don't serialize it. carve-php#901 has since shipped a fourth reading that needs none of them: **`attrs.order` is the marker, and it was already there.**

An authored id records the `#id` slot; a generated one does not. Verified on carve-php `main`:

```
# Notes

# Notes
```

publishes `{"id":"Notes"}` and `{"id":"Notes-2"}` - §5 satisfied - and the same document taken through the AST and written back is **byte-identical**. No `{#Notes}` lines.

Publishing the field broke three other places in that engine, and each was fixed by asking the slot list rather than asking whether an id exists: the HTML renderer's authored-versus-generated attribute ordering (PART 10 §1 puts a generated attribute after the authored run), the decoder's placement of the id in the attribute map, and the canonical writer - which had been emitting `{#Welcome}` above headings in 39 corpus documents.

So my earlier framing was wrong in one respect worth correcting: I said the tree "cannot" distinguish them because `order` is also absent on a programmatically built tree. That is true, but it does not make the distinction unusable - it makes the absent case mean "not authored", which is the safe direction. A hand-built tree's ids simply do not get written back as authored source, which is the behavior anyone building a tree by hand would want.

carve-js still has the writer half (carve-js#738) and now has a proven pattern to port; carve-rs has neither half yet (carve-rs#674).

Nothing here needs a spec change, which is the useful part of the answer. If anything wants adding, it is a sentence saying the slot list is how a consumer tells the two apart - because both engines that got this wrong got it wrong by asking whether an id exists instead.

PR #821, opened 2026-08-06
spec: code_block.attrs.order no longer diverges

### LINKED ISSUE #785: A code fence title's synthesized key gets an order entry in two engines and not the third
A code fence's title becomes a `title` key in `attrs`, and the engines disagree about whether that key gets an `order` entry.

````
``` php "src/Auth.php"
$ok = true;
```
````

| | carve-js | carve-rs | carve-php |
|---|---|---|---|
| `code_block.attrs` | `{"keyValues":{"title":"src/Auth.php"}}` | same **plus `"order":["title"]`** | same as carve-rs |

Same with a label alongside the title (`` ``` php "t.php" [lbl] ``). A fence with only a label and no title produces no `attrs` in any engine, so this is specific to the title.

## The engines agree everywhere the attributes were actually written

```
{#id .cls key=val}
text here
```

gives all three, identically:

```json
{"id":"id","classes":["cls"],"keyValues":{"key":"val"},"order":["#id",".class","key"]}
```

So this is not a general `order` divergence. It is only the key the engine SYNTHESIZES from a fence title.

## Which reading is right

`resources/ast-schema.json` defines the field as:

> **`order`** - Source-appearance order of the slots: `#id`, `.class`, or a bare key name. Absent on a programmatically built tree.

A fence title is not a slot in an attribute block - it is a quoted string after the language word, and the author never wrote `{title=...}`. On that reading carve-js is right: there is no source-appearance order of slots to record, so the field is absent.

The counter-reading is that `order` describes the keys in `keyValues` however they got there, and a tree carrying a key with no entry for it is incomplete. On that reading carve-rs and carve-php are right.

I lean carve-js, weakly, because the field's own description is about slots and because "absent on a programmatically built tree" suggests absence means "this did not come from authored syntax" - which is exactly the case here. But it is a one-line decision either way and the schema does not settle it, since `order` is optional and both shapes validate.

## Why it went unnoticed

The rendered HTML is identical in all three engines, so corpus conformance cannot see it. Found by comparing serialized trees across all 610 corpus documents alongside the HTML: 0 HTML disagreements, 107 tree disagreements. This accounts for 5 of them (`11-fenced-code-4`, `-6`, `-9` among them).

Same shape as markup-carve/carve#784 (table column alignment) and markup-carve/carve-php#877 (a crossref's `href`) - a field that renders the same and serializes differently. Worth deciding whether a corpus-wide AST comparison should join the HTML one; it is the only thing that finds these.

dereuromark:
`resources/ast-value-divergence.txt` declares the fields the three engines publish different VALUES for, and it is exact in both directions: a field that stops diverging has to have its line deleted, or `npm run ast:check` fails until it does.

`code_block.attrs.order` no longer diverges. `attrs.order` is the source-appearance order of authored attribute slots, and a fence title is written after the language word rather than as `{title=...}`, so the key the parser synthesizes from it has no authored position to record. carve-js already published none; markup-carve/carve-rs#676 and markup-carve/carve-php#898 landed the same reading in the other two.

Measured on the three current mains - carve-js `9da4fd5`, carve-rs `c25ca3e`, carve-php `fa26840` - against this branch's corpus, all three now publish:

```json
{"keyValues":{"title":"src/Auth.php"}}
```

for

````
``` php "src/Auth.php"
$ok = true;
```
````

## The deletion is what the declaration asks for, and it was observed failing

With the line still present, `ast:check` exits 1 on exactly this:

```
  FIXED      code_block.attrs.order no longer diverges - delete its line
```

With it deleted, the same run exits 0 and reports the remaining three fields as "All declared". No other line moved: `heading.attrs.id` (45), `paragraph.attrs.order` (1) and `list.tight` (1) are unchanged, so this is not a count drift being papered over.

## Gates

`npm test`, `npm run combinatorial:check`, `npm run core:check`, `npm run property:check` and `npm run test:conformance` are green.

Two gates could not run against a plain checkout of this branch, because they compare against sibling engine checkouts that only a scheduled workflow provisions:

- `npm run ast:check` - "a dependency outside the repository is missing: /tmp/carve-js/dist"
- `npm run claims:check` - "engine-claims: need all three engines, found 0 (none)."

`ast:check` is the gate this change is about, so it was run separately with `CARVE_JS_DIR` / `CARVE_RS_DIR` / `CARVE_PHP_DIR` pointed at built checkouts of the three mains named above; that is the run quoted above. `claims:check` was not run.

Closes markup-carve/carve#785.

### DIFF
--- a/resources/ast-value-divergence.txt
+++ b/resources/ast-value-divergence.txt
@@ -27,6 +27,5 @@
 # Format: <type.field>  <document-count>  <who diverges and where it is tracked>
 
 heading.attrs.id        45  carve-js publishes a GENERATED heading id, carve-rs and carve-php publish none (carve#750)
-code_block.attrs.order   4  a fence title's synthesized key gets an order entry in carve-rs and carve-php, not in carve-js (carve#785)
 paragraph.attrs.order    1  carve-php lists a repeated attribute key twice where keyValues holds it once (carve-php#878)
 list.tight                1  carve-rs has not implemented the clause behind corpus 228 yet, so it builds a loose list where the other two build a tight one - the HTML comparison reports the same document (carve#801)


dereuromark:
Superseded by #818 and #819, which deleted the same line while this was in CI. Closing; nothing here is missing from main.

PR #824, opened 2026-08-06
chore: make Node 24 the repo floor and default

jasonworden:
## What

Declares Node 24 as this repo's supported floor, moves the conformance matrix to `[24, 26]`, and bumps `@mermaid-lint/cli` to 0.42.1.

## Why

The repo has never declared a supported Node version — the effective floor was whatever its devDependencies happened to tolerate, stated nowhere.

The immediate trigger is `@mermaid-lint/cli@0.42.0`, which moved its own floor to `engines.node: ">=22"`. But rather than land at 22 — a version that enters Maintenance LTS in October 2026 — this goes to 24, the current Active LTS and the version the docs build and Pages deploy already run on.

_**TLDR; I still recommend this Node 24 change for this repo, even completely separate from that mermaid-lint floor.**_

## The part worth reviewing carefully

This **decouples the matrix floor from `@markup-carve/carve`'s `engines`**, reversing the policy currently documented above the matrix in `ci.yml`.

That comment argued the floor should track carve-js's `">=20"`, because the oldest runtime the published package promises has to be shown conformant somewhere. The counter-argument, now written into the comment in its place:

`carve-docs` is `private: true`, is never installed as a dependency, and ships **data** — the corpus fixtures downstream repos vendor as a submodule — rather than code. Downstream implementations run their own toolchains against those static files; this repo's `package.json` never enters their dependency graph. So its runtime is a contributor-and-CI concern, not a consumer-facing promise.

**The cost is real and I've named it in the comment rather than leaving it implicit:** conformance on Node 20 and 22 is no longer proven anywhere in this repo. If the `">=20"` promise on the published package needs backing, that belongs in carve-js's own matrix, against the code that actually makes the promise.

If you'd rather keep that guarantee here, the alternative is to leave the matrix at `[20, 22, 24, 26]` and drop the `engines`/`engine-strict` half of this PR — the two are load-bearing on each other, because the conformance job runs `npm ci` on every leg.

## Changes

| File | Change |
|---|---|
| `package.json` | `engines.node: ">=24"`; `@mermaid-lint/cli` → `^0.42.1` |
| `.nvmrc` | new — `24` |
| `.npmrc` | new — `engine-strict=true` |
| `ci.yml` | matrix `[20, 22, 24, 26]` → `[24, 26]`; rewrote the floor rationale |
| `ast-conformance.yml` | `22` → `24` — would otherwise fail `npm ci` against the new floor |
| `docs/README.md`, `.github/CONTRIBUTING.md` | document the requirement |

`docs-build`, `mermaid-lint`, and `deploy` were already on Node 24 and are untouched.

### On `engine-strict`

It decides the *quality* of the failure, not whether one happens — `@mermaid-lint/cli@0.42.1` doesn't support Node 20 either way. With it, a contributor on an old runtime fails immediately at `npm ci` with `Required: {"node":">=24"}`. Without it, they get a warning buried in install output and an opaque failure later. Easy to drop if you'd prefer the warning.

## Deliberately not changed

**`bump-downstream.yml` stays on Node 20** (line 103). It sets up the toolchain for *downstream* repos' suites, not this one's, and now serves five `lang: node` repos. Verified this floor cannot reach it:

1. The `bump` job has no `actions/checkout` — it clones the downstream repo into `work/` and runs every `npm ci` from there, so this repo's `package.json` and `.npmrc` are never on disk.
2. npm reads project config from the cwd, not ancestors or subdirectories — an `.npmrc` inside the spec submodule is invisible to a downstream root install.
3. No downstream repo uses npm workspaces (checked all five), so this manifest never enters their dependency graph.

The one job that does check out carve, `matrix-coverage`, runs only bash/`gh`/`grep`.

## Verification

All on Node 24.18.0:

- `npm test` → **1078 tests, 1077 pass, 1 skipped, 0 fail** — identical to baseline on `main`
- `corpus:build` + `git diff --exit-code -- tests/corpus tests/spec` → clean
- `core:check`, `engine:report --check`, `docs:build` → pass
- `lint:mermaid` → output **byte-identical** to 0.40.0 (2 diagrams across 737 files, both valid)
- `npm ci` on Node 22 → `EBADENGINE`, `Required: {"node":">=24"}` — the floor engages
- `actionlint` on all four workflows → clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dereuromark:
Checked the three factual claims this rests on, since the argument is entirely load-bearing on them:

- `package.json` is `"name": "carve-docs"`, `"private": true`, and has **no** `engines` field today - so ">=24" is a new floor, not a raise.
- The published package promising ">=20" is `@markup-carve/carve` in carve-js, a different repo with its own matrix.
- `@mermaid-lint/cli` does declare `>=22`, so 20 is already only nominally supported here.

All three hold. And naming the cost in the diff comment rather than deleting the old rationale is the right shape - the next person to wonder "why doesn't this prove Node 20?" gets the answer in place.

One thing I did not verify and would not merge without: whether anything in `scripts/` is expected to run under a contributor's system Node rather than the `.nvmrc` one. A floor is cheap to raise and awkward to lower once contributors have upgraded, so that is the maintainer's call, not mine - leaving it for Mark.

package.json
jasonworden, line 2: Regarding https://github.com/markup-carve/carve/pull/824#issuecomment-5199792017, this name might benefit from an update in a separate PR 🤷

jasonworden:
@dereuromark - your question was checkable, and the answer was good enough to act on. `engine-strict` is dropped in f43a086 for now (although adding back later would be good!).

Two scripts are documented as standalone invocations:

| Script | Documented at | Needs `npm ci`? |
|---|---|---|
| `compare-impls.mjs` | `implementation-comparison.md:318,387` | No - only `node:` builtins plus local `./lib`, `./spec` |
| `ast-conformance.mjs` | `ast-json.md:13,319` | **Yes** - `import { Ajv2020 } from 'ajv/dist/2020.js'` |

`ast-json.md:319` points implementers at `ast-conformance.mjs` to compare their engine against sibling checkouts, so a hard block at install turned away exactly the outside contributors this repo exists to serve. `engines` is enforced by npm at **install**, not by node at **execution**, and nothing in `scripts/` or `tests/` uses syntax newer than Node 18 - so `engine-strict` was the only thing standing in the way.

`engines: ">=24"` stays, now advisory: npm warns and the install proceeds. CI still enforces it for real, with the matrix at `[24, 26]`.

Verified with the removal, on Node 22:

| Check | Result |
|---|---|
| `npm ci` | exit 0, 5 `EBADENGINE` warnings, 0 errors |
| `npm test` | 1078 tests, 1077 pass, 1 skipped, 0 fail |
| `ast:check` | full run against the pinned carve-js build (`9dc8a94`) |

All byte-identical to the same commands on Node 24, where `npm ci` is warning-free. `lint:mermaid` is byte-identical to 0.40.0, and `actionlint` is clean on all four workflows.

`ast-conformance.yml` is schedule-only, so this PR's CI does not cover its `22 -> 24` bump. I dispatched it against the branch: all Node-dependent steps pass on 24. Two steps go red (`Cross-engine render comparison`, `Grammar cross-engine claims`), but both are pre-existing - I dispatched the same workflow against unchanged `main` minutes later, so the engine checkouts would match:

| | main (Node 22) | branch (Node 24) |
|---|---|---|
| rust | pass=635/637, mismatch=2 | pass=635/637, mismatch=2 |
| js | pass=635/637, mismatch=2 | pass=635/637, mismatch=2 |
| php | pass=635/637, mismatch=2 | pass=635/637, mismatch=2 |
| `cross_impl_diffs` | 0 | 0 |

Same two cases in both, differing only in millisecond timings. `carve-rs` and `carve-php` do not run on Node at all, so no Node version could move them.

PR #852, opened 2026-08-06
spec: pin the tab that overshoots a content column, and fix the oracle that missed it

dereuromark:
Closes the corpus half of markup-carve/carve-js#767.

Indentation is a **column claim** (§24 C1): a space advances one column, a tab advances to the next multiple of 4. `1. ` claims columns 0-2, so the item's content column is 3, and §24 C3 admits a block opener into the item only AT that column - "the content column is the item body's COLUMN 0". A tab from column 0 lands at column 4, one past it, so it makes the same claim four spaces make and gets the same answer: indented text.

The corpus pinned neither spelling for a tab. All three engines already agreed that three spaces nests and four is text, and it was that agreement which settled the tab case (carve-js#770, carve-php#922).

## The rule had four implementations, not three

`scripts/spec/layout.mjs`'s `dedent` consumed a straddling tab whole - exactly as two of the three engines did - so the **oracle every corpus fixture is measured against** nested the quote too.

That is why the documents could not land earlier: adding them alone turns this repo red, and the failure looks like a bad fixture rather than what it is. The oracle now gives the extra columns back as spaces, which is the only thing distinguishing AT the content column from PAST it.

## Two documents, not one

The four-space spelling goes in beside the tab, because neither states the rule alone: what makes the tab case decidable is that the two claims are the same claim.

## The pin

The published `@markup-carve/carve` still has the old behavior, so **243** is declared in `resources/engine-pin-drift.txt` until the pin moves past carve-js#770. **244 reproduces on the pin today** - that asymmetry is what makes this a tab-dedent defect rather than a disputed rule, and it is why the drift can be declared rather than the document deferred.

## Verification

- All three engines, fresh builds of their mains, render both documents byte-identically to the fixtures.
- Comparison page re-run: 647 documents, `661/661` per engine, zero mismatches across html, markdown, plain, carve and ansi.
- Position tests pass on both documents, which matters because carve-js#771 reports that residual spaces shift source offsets - that defect does not reach these shapes, but the technique is the same.


dereuromark:
Superseded by #853, which landed the same oracle fix and a corpus pair for the same rule while this was in review. Closing as the duplicate - #853's version is on `main` and this adds nothing to it.

One difference is worth keeping, and I will open it as a small follow-up rather than leave it in a closed PR.

**#853 pins the tab against the THREE-space spelling** (`243-...` and `243-...-2`), which shows the contrast: three spaces nests, the tab does not. That is the right pair for "these differ".

**The FOUR-space spelling is not pinned anywhere**, and it is the one that makes the tab case decidable. Its prose says so in as many words - "one column PAST it, which is what four spaces reach too" - but the claim that a tab and four spaces are the same claim is stated and not checked. An engine could pass both current documents while still answering column 4 differently depending on which whitespace reached it, which is exactly the shape of the defect this rule was written for.

Everything else here was duplicated: the oracle's `dedent` swallowed the straddling tab (same fix, same reasoning), the drift declaration, and the re-run comparison page.

Issue #913, opened 2026-08-06
Does a node's pos cover its opening markup? 279 of 690 documents disagree, and no check compares spans at all


dereuromark:
Split out of markup-carve/carve#534, whose remaining items are all implementable and need no ruling. This one does need a ruling, and it is larger than the gaps `ast:check` currently reports.

**Does a node's `pos` cover its opening markup?**

Measured on fresh clones - carve `cf5c03a`, carve-js `68ea9ba`, carve-php `876e312`, carve-rs `2ec3c1c`, across all 690 corpus documents.

`ast:check` reports 59 remaining findings, all of them a missing `pos`. It reports **zero** disagreements about where a present `pos` points, because it does not compare them: the three-way panel excludes positions (`scripts/spec/ast-values.mjs:27`, "compared elsewhere"), and the only content-level position rule runs on `text` nodes alone.

Build the panel it does not have, and **279 of 690 documents** carry a span the three engines do not agree on. Two representative rows:

- carve-rs gives a `block_quote` a span covering just the marker `"> "`, where the others cover the quoted content.
- carve-php gives a `text` node `[0,1]` where the other two give `[4,5]`. **Both slices are `"*"`** - which is why the existing slice-validity rule passes for both and has never flagged it.

That last detail is the reason this went unseen: the check that exists asserts the span *slices to plausible text*, a property every one of these divergences preserves. It is the shape catalogued in markup-carve/carve#755.

## The question

PART 12 does not say whether a node's `pos` begins at its opening markup or at its content. Two coherent answers:

- **Markup-inclusive.** `pos` covers the construct as written, marker included. Pro: a span round-trips to the source text that produced the node, which is what a formatter, a linter and an editor selection all want. Con: carve-js and carve-php both move, and every consumer relying on content offsets shifts.
- **Content-only.** `pos` covers what the node renders. Pro: fewer engines move. Con: a node's span no longer identifies the source construct, and nested constructs get spans that do not contain their children's - which makes a span tree that is not a tree.

**Recommendation: markup-inclusive**, with the containment invariant stated explicitly - a parent's span must contain every child's. The containment property is testable, which matters more than which convention wins: it turns 279 silent divergences into a rule a gate can enforce, whereas either convention without it leaves the same class of drift undetected.

## Why this is not the same ticket

[...]

PR #917, opened 2026-08-06
test: compare the spans, and split the position findings into permitted and owed

### LINKED ISSUE #534: ast:check still reports concrete PART 12 JSON conformance gaps across JS, Rust and PHP
`npm run ast:check` currently exits successfully but still reports concrete JSON AST conformance findings across the three main implementations.

This overlaps with the open PART 12 questions in #490 and #524, but the implementation-level findings are broader than those two spec issues alone.

Audit command:

```sh
CARVE_DISPLAY_LIMIT=80 npm run ast:check
```

Observed summary over 538 corpus documents:

- `carve-js` reference: 19 findings, 4 distinct
  - mostly missing `pos` on `text`, `hard_break`, `table_cell`, and `paragraph`
- `carve-rs`: 39 findings, 33 distinct
  - missing positions
  - tree-shape differences around heading refs, refs, abbreviations, comments, invisible definitions, and related cases
- `carve-php`: 19 findings, 15 distinct
  - missing positions
  - tree-shape differences around nested links/autolinks, cross-references, cyclic refs, and heading ids

Representative Rust/PHP tree-shape clusters include:

- `03-links-*`
- `15-heading-ids*`
- `35-cross-reference`
- `47-numbered-cross-references-*`
- `111-cross-references-resolve-inside-footnote-bodies`
- `118-cyclic-cross-reference-resolves-to-one-level*`

Why this matters:

HTML conformance can pass while JSON AST consumers still see different trees or missing positions. That needs its own tracking separate from rendered-output corpus agreement.

Expected:

- Decide which `ast:check` findings are spec-side ambiguities versus implementation bugs.
- Add implementation-specific issues/fixtures where needed.
- Eventually make `ast:check` either clean or fail on the remaining non-waived findings.

dereuromark:
The checker half of #534. Nothing here needs a ruling, so this does not carry
`needs-decision`.

`npm run ast:check` reported 83 position findings across the three engines,
every one of them "missing pos", and **zero** disagreements about where a
present `pos` points. Not because the engines agree, but because nothing
compared them.

## The span panel

`scripts/spec/ast-values.mjs` drops every position key from the value signature,
on the stated grounds that they are "compared elsewhere". Elsewhere is
`checkPositions`, which compares each engine against the SOURCE and never
against another engine, and whose only content-level rule runs on `text` nodes
alone.

New `scripts/spec/ast-spans.mjs` compares them three ways round. Measured over
716 documents at carve-js `3d95e94`, carve-rs `83ab9c1`, carve-php `876e312`:
**30 rows, across 309 documents** carrying at least one span the three engines
do not agree on.

Most of that is the extent convention. Two rows are defects, and both are the
trap this repo keeps cataloguing:

`117-footnote-definition-inside-a-container-is-collected` - carve-rs gives the
block quote a span covering the marker alone:

```
See [^a].

> [^a]: note body
```

carve-rs publishes `block_quote [11,13]`, which slices `"> "`. carve-js and
carve-php publish `[11,28]`, the whole quote. Invisible to the containment rule
because the footnote hoists to the root, leaving the quote with no children to
contain.

`132-thematic-break-requires-contiguous-markers` - source:

```
* * *
```

carve-php publishes `text [0,1]` where carve-js and carve-rs publish `[4,5]`.
**Both slices are `"*"`**, so the one content-level position rule passes for
both. That rule asserts the span's CONTENT, and a span pointing at the wrong
occurrence of the same character survives it. So this panel asserts nothing
about what a span slices to. It compares the spans.

### What it does not decide

Whether a node's `pos` covers the markup that opens it - the list marker, the
`>`, the `[^a]: ` - is #913, `needs-decision`. Most of the EXTENT rows below are
that question, and the panel reports the disagreement **without naming a side**.
That is worth having before the ruling: a declared count that moves means an
engine changed its mind about a span, and nothing else in this repo can see that
happen.

Declared in `resources/ast-span-divergence.txt` rather than gated at zero, for
the reason the value panel already gives: in this fleet a fix lands in one engine
first, so the engine that is right is routinely the odd one out for a while.

## The position split

`resources/ast-position-waivers.txt` declares each finding per engine, per
document and per node type, as either permitted or owed:

| engine | findings | waived | outstanding |
| --- | --- | --- | --- |
| carve-js | 34 | 28 | 6 |
| carve-rs | 33 | 16 | 17 |
| carve-php | 20 | 16 | 4 |

The permitted lines are **CONTROLS**: they pin behavior PART 12 section 4
exempts, so no A/B over an engine can fail on them. A merged `text` run whose
pieces are not adjacent is named verbatim at docs/ast-json.md:148-153, and a
table cell continued on a `+` line at docs/ast-json.md:113, with
docs/ast-json.md:427 already naming carve-js as the engine that omits it. What
is falsifiable is the declaration itself, in three directions.

The split earns itself immediately: the permitted half is 16 findings per engine
and grew with the corpus over the past week, while the owed half did not move at
all. Printed as one number, those two were indistinguishable.

A DERIVED engine is not reconciled. carve-rb serializes carve-rs's tree, so its
findings are carve-rs's arriving a second time; declaring them separately would
record one engine's debt twice and go red on every host with the Ruby binding
built, which is every run of the scheduled workflow. Same reason the shape panel
withholds that engine's vote.

## The early return

`reportValueDisagreements` advertised "a listed field no longer diverges ->
fixed, delete the line, fails until you do" and returned before reading
`resources/ast-value-divergence.txt` whenever nothing diverged - the one state
that direction describes. Appending a fabricated line to the declaration exited
0 against the code on main; it now exits 1 naming `FIXED table_cell.align`.

The reconciliation moves to `reconcileDeclared` and is called unconditionally.
The three declarations also stop calling `process.exit` individually, so a run
with drift in more than one reports all of it rather than revealing the next
after each fix.

## The grouping key

Grouping stripped the filename and kept one example, so carve-php's entire
report read `14x missing pos on "text" [03-links-12.crv]` under "1 distinct".
It was six documents and at least two causes: a permitted merged run in the
document it named, and four placeable text nodes in a document it did not. The
line named the cause nobody should act on and hid the one somebody should.
Groups now carry a document set.

## Also

PART 12 section 3a's source-shape rule returns how many referencing nodes it
could not apply to. `sliceOf` returning null makes it go quiet on exactly the
nodes an engine failed to place, so a rule that stopped covering part of the
corpus read identically to one that covered it and found nothing. Zero in all
three engines today, printed when it is not.

Corpus 257 pins a tab-continued list item, with the two-space spelling as the
control in the same section:

```
- item
	more

x
```

carve-js publishes no position for the item's paragraph or any of its three
inlines. With two spaces, or with a `1.` marker, it places all four. carve-rs
and carve-php place them either way and agree to the offset. The HTML is
byte-identical in all three, which is why nothing here could see it.

## Engine gaps filed

| gap | tracked |
| --- | --- |
| tab-indented footnote continuation drops five positions | markup-carve/carve-rs#736 |
| figure over a reference image loses its position | markup-carve/carve-rs#737 |
| the four text runs at the nesting cap | added to markup-carve/carve-rs#716 |
| the four text runs at the nesting cap | markup-carve/carve-php#965 |
| emptied definition description | markup-carve/carve-js#813 |
| tab-continued bullet item | markup-carve/carve-js#814 |

One more the panel exposes and this does not file: carve-rs publishes
`definition_description [8,14]` on corpus 227, which slices `:  [r]` and ends
inside the reference definition token. It is declared under
`definition_description (extent)` and is not the #913 question - an end offset
inside a token is wrong under either convention.

## What could not be measured on this host

`gate-run` reports `partial`: `npm test`, `npm run combinatorial:check`,
`npm run core:check`, `npm run property:check` and `npm run test:conformance`
ran and passed. Four gates could not run, verbatim:

- `npm run ast:check` - "a dependency outside the repository is missing:
  /tmp/carve-js/dist"
- `npm run claims:check` - "engine-claims: need all three engines, found 0
  (none)."
- `npm run degradation:check` - "degradation-claims: need all three engines,
  found 0 (none)."
- `npm run fmt:check` - "fmt-fixture-claims: need all three engines, found 0
  (none)."

`ast:check` was run separately against freshly cloned and built checkouts of the
three engines' mains, passed at exit 0, and is where every number above comes
from. The other three drive `compare:impls` and were not run here.

The cross-engine numbers on docs/implementation-comparison.md are NOT
hand-edited. Corpus 257 is added to the declared-lag line the mechanism from
#908 reads, so the test counts the fixtures itself.

## Conflicts to expect

#914 is open and unmerged. It claims corpus **257 and 258**, and edits the same
declared-lag line on docs/implementation-comparison.md that this does. Whichever
merges second needs a rebase, `npm run corpus:build`, and the renumbered
document names carried into `resources/ast-position-waivers.txt`. It also
touches tests/separator-role-split.test.mjs, which this does not.

Closes #534


dereuromark:
Renumbered this PR's corpus category from 257 to **259**, because markup-carve/carve#914 claims 257 and 258. The union check over `tests/corpus` on `main` plus the files of every open PR returns 258 as the highest claimed number, so 259 is the next free one.

One correction to how the renumber had to be done. A category's number is not a name that can be edited: `scripts/generate-corpus.mjs` assigns it as the section's **position** among the example sections that carry a `compare` pair, and the append-only guard actively pulls a hand-renamed file back. Renaming the fixtures to `259-*` and re-running `corpus:build` on this branch alone produces:

```
generate-corpus: the corpus numbering is APPEND-ONLY.

  1 existing category would be renumbered:
    259-a-tab-continues-a-list-item-just-as-two-spaces-do -> 257-a-tab-continues-a-list-item-just-as-two-spaces-do
```

259 exists only once two sections precede this one, and those two sections are markup-carve/carve#914's. So this branch is now rebased onto `issue-907` and the PR targets it; `corpus:build` then yields 259 with no guard override needed at review time, and GitHub will retarget this to `main` when markup-carve/carve#914 merges. The trade is explicit: this PR is now stacked on markup-carve/carve#914 rather than silently colliding with it at merge time.

Carried across with the rename:

- `tests/corpus/259-a-tab-continues-a-list-item-just-as-two-spaces-do{,-2}.{crv,html}`
- `tests/spec/259-a-tab-continues-a-list-item-just-as-two-spaces-do.test`
- the three `carve-js` lines in `resources/ast-position-waivers.txt`
- the declared-lag list in `docs/implementation-comparison.md`
- the corpus count in `docs/index.md`, now 730

`resources/engine-pin-drift.txt` needs no entry for this category and never had one: the divergence here is in PART 12 positions, which the waiver file owns, not in HTML, which is what that file declares.

Grepping the worktree for `257-a-tab` returns nothing. Grepping for the bare token `257` returns only markup-carve/carve#914's own category and two unrelated CVE-2021-42574 mentions.

Issue #1003, opened 2026-08-08
The AST value-divergence ledger declares two divergences that no longer exist, and the engine pin predates the ruling the docs now state


dereuromark:
`resources/ast-value-divergence.txt` declares two cross-engine divergences. **Both are gone.** A ledger that tolerates a divergence which no longer exists cannot catch a regression back to it, which is the opposite of what it is for.

Surfaced while landing `markup-carve/carve#1002`; codex flagged the first line on review and the second turned up on measuring it.

## Measured, all three engines built fresh from `main`

carve-js `c245eed`, carve-php `ebd9386`, carve-rs `39e6968`.

### `link.ref` - the ledger's claim is inverted

The entry says carve-php publishes the heading's rendered text where carve-js and carve-rs publish the authored label. Given `` # `code()` heading `` and `` [`code()` heading][] ``:

```
js   "ref": "code() heading"
rs   "ref": "code() heading"
php  "ref": "code() heading"
```

All three publish the rendered text. carve-js and carve-rs moved onto that value at `markup-carve/carve-js#875` and `markup-carve/carve-rs#799`; carve-php never moved. The entry describes the pre-merge picture.

It also cites `markup-carve/carve#962` as tracking an open question. That ticket is now RULED and closed: `ref` is the resolution key, which is what all three publish.

### `text.value` - also gone

The entry says carve-php's table caption keeps the trailing space its own HTML drops, publishing `"Cap "` where the others publish `"Cap"`. On corpus `268-trailing-whitespace-on-a-content-line-is-dropped-6` (source: `| a |` then `^ Cap `):

```
js   "value":"a"    "value":"Cap"
rs   "value":"a"    "value":"Cap"
php  "value":"Cap"  "value":"a"
```

No trailing space anywhere. carve-php was fixed under `markup-carve/carve#963`.

(The node ORDER differs - carve-php emits the caption before the cell. That is outside this ledger's field-value claim and is not asserted here as a defect; it is noted only so the next reader does not think the rows were misread.)

## Why this is worth a ticket rather than a cleanup

The file's own header explains its contract: a declared divergence with a document count, so the check "still fails when the fact changes". Two entries now declare facts that have already changed and did NOT fail.

That is the allowlist form of the shape `markup-carve/carve#755` collects. An entry that tolerates behavior no engine produces is not neutral - it is a hole sized exactly to the old defect. If carve-php regressed to `"Cap "` tomorrow, the ledger would wave it through.

[...]

Issue #1001, opened 2026-08-08
A denied URL scheme split by DEL reaches the HTML target live in two of three engines


dereuromark:
Found while measuring PART 9 section 29 for `markup-carve/carve-js#896`. It is not that clause and was deliberately left out of `markup-carve/carve-js#914`, but it should not stay unfiled: two of the three engines emit a live `javascript:` href on the HTML target for a destination the third one blanks.

## The input

A denied scheme with DEL (U+007F) sitting inside it:

```
[t](java<DEL>script:alert(1))
```

written as one line, where `<DEL>` is the single byte U+007F.

## The three answers, on the HTML target

carve-php `ebd9386`:

```html
<p><a href="">t</a></p>
```

carve-rs `39e6968`:

```html
<p><a href="java&#127;script:alert(1)">t</a></p>
```

carve-js `76dadb6`:

```html
<p><a href="javascript:alert(1)">t</a></p>
```

carve-js is the sharp one. It NORMALIZES the DEL away and emits a working `javascript:` URL - the writer manufactures the live scheme out of a destination the probe had already looked at, which is the same shape as `markup-carve/carve-js#893` one target over. carve-rs keeps the byte, which most browsers ignore when reading a scheme, so it is live there too in practice.

## Why the probe misses it

The probe strips the C0 controls and whitespace before it reads the scheme, so `java<VT>script:` and `java<SOH>script:` are both caught. DEL is not in that class:

- carve-js `SCHEME_PROBE_STRIP_RE` is `/[U+0000-U+0008, U+000E-U+001F, plus the whitespace shorthand]+/gu` - U+007F is above U+001F, so the probe reads `java<DEL>script` and finds no scheme, and the destination is then normalized on the way out.
- carve-php strips a broader class inside its probe, which is why it blanks.

So the divergence is one character wide, and it is exactly the character PART 9 section 29 T5 keeps OUTSIDE the content class for this reason.

## Measured

Section 29's own class is not affected - every one of U+0000..U+0008, U+000B, U+000C, U+000E..U+001F is removed by the probe in all three engines, and `java<that>script:` is blanked. The C1 controls should be checked the same way; U+009B and U+009D at least deserve a row.

## Not proposed here

Which behavior is correct is a maintainer call: blanking the destination (carve-php) versus percent-encoding the byte so the scheme cannot re-form. Both are defensible; emitting `javascript:alert(1)` is not.

dereuromark:
[...]

Issue #994, opened 2026-08-07
PART 11 section 1 is normative, unenforceable by any gate, and violated identically by two engines in six shapes


dereuromark:
PART 11 section 1 states the canonical writer's invariant: `to_html(fmt(x)) == to_html(x)`. It is normative, it is violated in five distinct ways, and the violations are IDENTICAL in two independent engines.

That combination is the point of this ticket. A defect two engines share, byte for byte, is not two engine defects.

## Measured

carve-js at `198edda` and carve-php `main` at `9375df1`, each built from a fresh clone. The `fmt` output strings are identical between the engines for every row, and so is the broken HTML.

| cause | what `fmt` produces | result |
|---|---|---|
| a leading break gains a closer | `***`/`a`/`---`/`b` becomes `---`/`a`/`---`/`b` | `<hr> <p>a</p> <hr> <p>b</p>` becomes `<p>b</p>` |
| a relocated definition promotes the next block to byte 0 | `[a]: /u`/`---`/`p`/`---` becomes `---`/`p`/`---`/`[a]: /u` | renders **empty** |
| a `+`-attached image loses its caption | the `+` becomes indentation | `<figure>` becomes `<img>` plus literal `^ cap` |
| only the LAST line of a `+`-attached block is indented | mixed indentation in one block | `<hr>` becomes an em dash |
| a header cell's first character is read as alignment | `\| ~x~ \|` becomes `\|=~x~\|` | strikethrough lost, column silently centered |
| an empty footnote body | `[^f]:` with nothing after it | not a definition; reference and note both go literal |

Three of the six destroy the document outright. None is a styling difference.

Engine tickets: `markup-carve/carve-js#899`, `#901`, `#902`, `#903`, `#904`, and `markup-carve/carve-php#1069`. carve-rs is unmeasured because its checkout has been under edit; expect it to reproduce, but measure rather than assume.

## Why two engines agree

Not coincidence, and worth stating because it decides where the fix belongs. Section 1 gives the invariant. Sections 2 to 11 give the writer its rules. **No clause says the writer must check that what it emits re-parses as what it was written from.** Each of the six is a locally reasonable decision:

- section 6a says a thematic break is written `---`, unconditionally;
- PART 12 section 10 hoists definitions to document level, and the writer emits the tree's order;
- converting a `+` attachment to indentation is correct for a fenced block, which is why the fence control passes;
- `|=` is the canonical header cell;
- an empty body is written as an empty body.

[...]

PR #1008, opened 2026-08-08
docs: draft implementation capability labels

dereuromark:
## Summary

Adds a draft PR writeup for implementation capability labels so Carve support can be described more precisely than one broad compatible/conformant status.

## Notes

- proposes labels such as `core-html`, `ast-json`, `positions`, `formatter`, `multi-target`, and `syntax-tooling`
- keeps this as a non-normative draft under `docs/pr-drafts/`
- intended as review material before applying changes to live docs

## Validation

- Not run; documentation draft only.

### DIFF
--- a/docs/pr-drafts/001-capability-labels.md
+++ b/docs/pr-drafts/001-capability-labels.md
@@ -1,0 +1,141 @@
+# Draft PR: Add Implementation Capability Labels
+
+## Summary
+
+Add explicit capability labels for Carve implementations so "supports Carve"
+does not imply AST exchange, source positions, formatting, optional extensions,
+or multi-target rendering.
+
+The minimum full-parser label becomes `core-html`: parse Tier-1 source and
+match the core HTML corpus. Other capabilities are named independently.
+
+## Problem
+
+`docs/implementing-carve.md` currently defines the bar for "a Carve
+implementation" as Tier-1 core corpus conformance, but the ecosystem also tracks
+AST JSON, source positions, formatter round-trip, optional corpus support, and
+non-HTML targets. Those are real capabilities, but bundling them under one
+informal "compatible" status raises the apparent cost of writing a new engine.
+
+The result is a messaging problem:
+
+- a new engine can be fully useful for source-to-HTML before it has AST JSON;
+- an editor grammar can be useful without being a renderer;
+- a Pandoc bridge can be meaningful without being a parser;
+- formatter and position support are higher bars than core rendering.
+
+This PR separates those claims.
+
+## Proposed Changes
+
+### 1. Add capability labels to `docs/implementing-carve.md`
+
+Add this section after "Conformance tiers":
+
+```md
+## Implementation capability labels
+
+Use capability labels for implementation status. A project may support Carve at
+different depths; do not imply AST exchange, source positions, formatting, or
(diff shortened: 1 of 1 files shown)


dereuromark:
Useful, and the entry-bar argument is right - a new engine should be able to claim `core-html` without implying AST exchange or positions.

One problem, measured rather than argued: **three of the labels are gated on checks that cannot fail for the defects they are meant to certify.** Overnight work in this org gives each a number.

## `formatter`

Proposed gate: `tests/corpus-fmt-roundtrip.test.mjs`, meaning "preserves parse/fmt round-trip".

Measured last night: **six distinct canonical-writer defects violating PART 11 section 1**, three of which destroy the document outright - a leading thematic break gaining a closer and swallowing everything to the next one; a relocated definition promoting a block to byte 0; a header cell whose first character is read back as an alignment marker.

All 839 corpus documents satisfy the invariant, and always did. It took **1,711 generated documents across six axes** to find those six causes. The corpus contains no document in any of the six shapes, so the gate cannot fail for them - not "did not fail", cannot.

A label whose gate cannot fail for the defects it certifies is worse than no label, because it converts an unknown into a claim. This is the shape `markup-carve/carve#755` collects, one level up: not a check that cannot fail, but a LABEL resting on one.

## `safe-mode`

All three engines would have claimed this yesterday morning while carrying:

- a denied URL scheme split by a control character reaching the `href` (carve-js and carve-rs, `markup-carve/carve#1001`);
- a profile link policy whose denylist is defeated by any control character inside the scheme (`markup-carve/carve-js#917`, `markup-carve/carve-rs#835`);
- an author's U+0001 corrupted into a space on the default HTML target by a sentinel collision (`markup-carve/carve-php#1077`).

So the open question this draft raises - should `safe-mode` be separate when hardening is mandatory for `core-html` - answers itself. **Separate, and with a gate that can fail.** Otherwise the label is claimed truthfully-by-the-gate and falsely-in-fact, which is the worst of the three outcomes.

## `ast-json`

Proposed gate: `npm run ast:check`. That workflow is **red on `main` right now**, and its divergence ledger (`resources/ast-value-divergence.txt`) declares two divergences that no engine produces any more - `markup-carve/carve#1003`. An allowlist sized to a defect that has been fixed cannot catch a regression back to it.

## What I would change

Either cite each gate honestly - "this gate covers X and not Y" - or make the label depend on a gate that can fail. For `formatter` the missing gate already has a ticket: `markup-carve/carve#994` asks for a generated invariant sweep, and its numbers are the argument for funding it. The two proposals are complementary, and this one is the better case for the other.

## Two smaller points

- The suggested rows assert capabilities per engine. Those claims want measuring before they land, not after - three of tonight's tickets exist because a claim about which engine did what was inherited rather than measured.
- `carve-wasm` "inherits carve-rs labels" is the same assumption that has `markup-carve/carve-rb` publishing a tree from a carve-rs build **33 commits** behind its parent (`markup-carve/carve-rb#55`). A binding inherits its parent's labels only at the pin it actually carries.

dereuromark:
Closing this draft: it captured the discussion as docs, but the actionable work needs to be implemented as real gates first. The review points out that several proposed labels would certify behavior through checks that cannot fail for the defects they name, so keeping this open would misrepresent the next step.

Issue #981, opened 2026-08-07
corpus: pin the fence-boundary class, six rows across five collectors


dereuromark:
A read-only sweep measured 12 collectors x 8 boundaries x 3 fence kinds x 3
engines (774 measurements, 0 errors) and found a class no corpus document
reaches: a boundary line that appears INSIDE an open fence ends the container
anyway. Severing cells: carve-js 24, carve-php 34, carve-rs 23. Of the 840
corpus documents, 15 carry a lone `+` and none combines a `+` with a fence and a
blank line, which is why the class survived every gate.

Summary on `markup-carve/carve-php#1047`. Engine tickets:
`markup-carve/carve-js#884`, `markup-carve/carve-php#1049`,
`markup-carve/carve-rs#802`.

## What to pin

One category, six rows, each reaching a DIFFERENT collector rather than one row
per severing cell. The rule is the same in all six and comes from the clause,
not from any engine:

- PART 9 §17 L3 attaches ONE BLOCK "up to the next blank line, sibling marker,
  or a further `+`". That list bounds the attached BLOCK; a fenced block's own
  extent runs opener to matching closer, so a line between the two is inside the
  one block and the boundary list never sees it.
- §24 S1/S2 place a line by the column it reaches, never by its first character.
- §28 makes a comment fence's body verbatim and invisible.
- §17 L6 collects a definition wherever it is written and has it interrupt the
  open paragraph, which is what decides the colon-fence row.
- §10 I2/I6: list markers never interrupt an open paragraph, in every container.

Rows, and the collector each one reaches:

1. A definition line inside a `+`-attached open CODE fence in a footnote body.
   The worst cell in the grid: all three engines end the note body there, leave
   the fence unterminated, and the trailing text escapes to become the
   document's FIRST block, ahead of the paragraph that references the note.
   §17 L3's boundary list does not name a definition line at all.
2. A blank line inside the same fence. Same collector, different predicate.
3. The same blank line in a definition list's body, which is a separate loop in
   every engine (`markup-carve/carve-php#1003` is the precedent for one row
   letting the other half stay broken).
4. A definition line inside a `+`-attached open COLON fence in a block quote.
   The body is not verbatim here, so the definition IS collected, leaves no
   trace, and interrupts the paragraph; what no reading gives is a div that ends
   at that line, which is what one engine produces.
5. A list marker at the body's column inside an item's indented colon fence. All
   three engines close the div, open a real list, and leave a stray empty div.
[...]

PR #1009, opened 2026-08-08
docs: draft tree-first renderer guidance

dereuromark:
## Summary

Adds a draft PR writeup recommending tree-first resolution for production renderers instead of copying the executable oracle's sentinel-based HTML post-processing shape.

## Notes

- explains why the oracle shape is useful for corpus checks but fragile as production architecture
- sketches a tree-first parse/resolve/render pipeline
- keeps this as a non-normative draft under `docs/pr-drafts/`

## Validation

- Not run; documentation draft only.

### DIFF
--- a/docs/implementing-carve.md
+++ b/docs/implementing-carve.md
@@ -58,6 +58,40 @@ A new parser only needs to:
 2. Produce HTML.
 3. Compare against the paired `.html`, normalizing trailing whitespace.
 
+## Renderer architecture
+
+The executable spec in this repository is an oracle, not a recommended engine
+architecture. Its sentinel-based passes are allowed because they run against the
+trusted corpus and make byte comparisons easy. That trust bounds the security
+argument; it does not make the oracle maintenance-free. The oracle can still
+drift from the engines when a rule is copied into another local string pass.
+
+Production engines should resolve document-level semantics on the tree before
+serializing HTML:
+
+1. Parse block and inline structure.
+2. Collect definitions and generated ids.
+3. Resolve references, footnotes, cross-references, caption numbers, and
+   abbreviation applications into tree state.
+4. Render the resolved tree to the requested target.
+
+Avoid rendering temporary HTML and then rewriting it with string replacements.
+That style is acceptable in the oracle, but production renderers should prefer
+semantic nodes and target-specific serializers.
+
+For example, a paragraph that resolves to a single image with a caption should
+be represented internally as block-level image or figure state before
+serialization. The resolved state can then render directly as `<figure>` /
+`<figcaption>` on HTML, or as the target's own block-image form elsewhere.
+
+The same resolved tree should feed HTML, Markdown, plain text, ANSI, and static
+HTML renderers. Target-specific renderers may choose different output shapes,
+but they should not reimplement reference and caption resolution separately.
+
+Keeping the oracle's current shape is a deliberate maintenance cost, not an
+exemption from the same class of drift production engines avoid by resolving
+tree state first.
+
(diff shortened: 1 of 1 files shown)


dereuromark:
Land this. Its problem statement is not hypothetical - every bullet in it has a measured instance from the last day, in one engine.

## "The renderer has to prevent sentinel injection"

Five defects in carve-php, all reproducing on `main` before they were fixed:

- `markup-carve/carve-php#1077` - `HtmlRenderer` used U+0001 as a soft-break sentinel, so an author's own U+0001 was corrupted into a space **on the default HTML target**. carve-js and carve-rs both kept the character; carve-php was alone.
- `markup-carve/carve-php#1087` - four more fixed in-band sentinels colliding with authored content. One of them, `CarveRenderer::MARKER_COLUMN`, did not merely eat the character: it **dedented the paragraph out of its list item**. A sentinel collision changing block structure is the sharpest possible version of this draft's argument.
- And the mechanism the fixes depend on had its own version of the bug: `DocumentSentinels::pick()` stepped a whole run at a time, so **1066** authored characters exhausted the search and it returned the colliding run - against a docblock claiming every private-use code point was available.

## "Semantic decisions are made after serialization instead of before it"

`markup-carve/carve-js#902`: a `+`-attached image lost its caption, because converting the attachment to indentation left the image no longer a standalone image paragraph, so PART 9 section 4 did not attach the caption. The image-and-caption relation was not a tree fact, so it did not survive a change in how the block was written.

That is this draft's own `image_block` open question, arriving as a defect. I would answer it the same way: **resolved tree state**, because the bug was that it wasn't.

## "Non-HTML targets need the same resolution work but cannot reuse HTML rewrites"

`markup-carve/carve-php#1082` - a single ill-formed UTF-8 byte emptied a paragraph, and it did so **twice over through two unrelated functions**: `htmlspecialchars()` without `ENT_SUBSTITUTE` returning the empty string on the HTML target, and a `/u` pattern making `preg_replace()` return null on the Markdown, plain and ANSI targets. Patching either site would have left the other. It was fixed at the parse door - which is the tree-first move.

## One thing worth adding

The draft says the oracle's shape is acceptable "because it runs against the trusted corpus". Last night the oracle's own sentinel-based passes needed a fix before corpus rows could be committed (`markup-carve/carve#981`), and `scripts/spec/layout.mjs` turned out to carry a **ninth copy** of a rule already fixed in eight engine sites. "Trusted corpus" bounds the security argument, not the maintenance one - the oracle drifts too.

That does not weaken the draft. It strengthens the non-goal: keeping the oracle as-is is a deliberate cost, and saying so explicitly is better than implying the oracle is exempt from the problem.

## Open questions, from measurement

- **`image_block` public or internal?** Internal is enough for the defect above, but PART 12's wire shape is `additionalProperties: false`, so publishing it later is a schema change that must land engine-first. Deciding now is cheaper than deciding twice.
- **Abbreviation expansion tree-level or renderer-level?** Tonight's evidence leans tree-level: the abbreviation budget was sized per renderer in carve-php and needed the same fix in three places.

dereuromark:
Closing this draft: it is guidance text, but the requested actionable PR should target a concrete renderer simplification or behavior change, not a proposal document.

dereuromark:
Reopening after reading the measured follow-up in https://github.com/markup-carve/carve/pull/1009#issuecomment-5225751721. This draft is not the code-change PR, but the comment makes the rationale concrete enough to land as guidance: sentinel collisions, delayed semantic decisions, and target-specific rewrites have all produced recent defects.

PR #1026, opened 2026-08-08
docs: make examples migration-clean Carve

dereuromark:
Reapplies the useful parts of the stale `docs/quote-admonition-titles` branch onto current main:

- quote every prose admonition title so examples use valid Carve opener syntax
- add explicit heading ids where migration-sensitive links rely on them
- keep the newer wording added since the old branch diverged

Tested: `npm test` (1878 passed).

### DIFF
--- a/docs/case-study/index.md
+++ b/docs/case-study/index.md
@@ -5,7 +5,7 @@ description: The original design research behind Carve - historical, not normati
 
 # Case Study
 
-::: info Historical document
+::: info "Historical document"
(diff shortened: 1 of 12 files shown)


dereuromark:
Closing. The branch is 409 commits behind `main` and 10 of its 12 files have changed since (`docs/index.md` 83 times), so a rebase would be a rewrite. If the goal still matters, it should come back as a new ticket written against the current docs.

Issue #1040, opened 2026-08-08
Cross-engine render comparison is red on 12 documents: the markdown and carve targets disagree where HTML does not


dereuromark:
`AST conformance` runs two independent checking steps. `PART 12 conformance` is
green again as of markup-carve/carve#1039 and markup-carve/carve-php#1114. The
other, `Cross-engine render comparison (all targets)`, is red and has been for
days - it failed on the 2026-08-07 runs too - so
markup-carve/carve#1023 cannot close while it stands.

## Measured

`npm run compare:impls -- --fail-on-diff` on fresh clones built for this run:
carve-js `e35cfcb`, carve-rs `fa4d3f7`, carve-php `177989a`, over carve
`02a947c`. Identical to what CI run 31265464556 reports, so this is not a local
artifact.

```
cross_impl_diffs=12
html:     compared=880 diffs=0
markdown: compared=880 diffs=9
plain:    compared=880 diffs=0
carve:    compared=880 diffs=3
ansi:     compared=880 diffs=0
```

Every engine passes its own HTML fixtures (901/901, 0 mismatches). What
disagrees is the non-HTML targets, where most documents have no golden and the
check is engine agreement alone.

The 12 documents, with the date each entered the corpus:

| document | target | added |
| --- | --- | --- |
| `14-frontmatter` | carve | 2026-07-17 |
| `134-footnote-definition-requires-an-inline-body` | markdown | 2026-08-02 |
| `135-footnote-definition-separator-must-be-a-space` | markdown | 2026-08-02 |
| `159-indented-reference-and-footnote-definitions-stay-literal-2` | markdown | 2026-08-02 |
| `163-unresolved-footnote-reference-with-a-trailing-attribute-stays-literal` | markdown | 2026-08-02 |
| `184-a-definition-below-every-content-column-folds-as-text` | markdown | 2026-08-05 |
| `267-a-definition-marker-s-separator-is-a-space-and-it-is-a-run-8` | markdown | 2026-08-07 |
| `267-a-definition-marker-s-separator-is-a-space-and-it-is-a-run-9` | markdown | 2026-08-07 |
| `279-a-boundary-line-inside-an-open-fence-does-not-end-the-container-2` | carve | 2026-08-07 |
| `279-a-boundary-line-inside-an-open-fence-does-not-end-the-container-3` | carve | 2026-08-07 |
| `284-a-ragged-table-keeps-each-row-s-cell-count` | markdown | 2026-08-08 |
[...]

Issue #1028, opened 2026-08-08
The three canonical writers disagree on 1 generated document in 17, which is what blocks gating the cross-engine sweep


dereuromark:
PART 11 section 2a makes the canonical form a byte property: for one document there is one canonical source, and every conforming writer must produce it. Over the corpus that holds, and `ast-conformance.yml` gates it (`npm run fmt:check`, 
`compare:impls --roundtrip --fail-on-diff`).

Over documents the corpus does not reach, it does not hold, and this is what blocks wiring the one mode of the new generated gate that would measure it.

## Measured

`scripts/property-check.mjs --engines` writes each generated document to a file and compares the three engines' `carve --carve` output byte for byte. Fresh clones, built: carve-rs `9304128` (release, isolated `CARGO_TARGET_DIR`), carve-js `3747584`, carve-php `main`.

**400 generated documents, seed 12345: 23 diverge.** Roughly one in 17. Pipeline proved on the way in: all three agree on the other 377, so this is a real disagreement rather than a broken harness.

The first five, in full.

### 1. A block attribute before a container is kept, dropped, or escaped

Input:

```
/i/
---
{.cls}
- item
$m$
{.cls}
> quote
```

carve-rs:

```
/i/

---

{.cls}
- item
  $m$

{.cls}
> quote
```

carve-js drops the second attribute line entirely:

```
/i/

---

{.cls}
- item
  $m$

> quote
```

carve-php keeps it, escaped, inside the list item:

```
/i/

---

{.cls}
- item
  $m$
  \{\.cls\}

> quote
```

Three answers, and carve-js's loses the author's attribute.

### 2. An escaped space at end of line

Input:

```
\ 
@user
```

carve-js writes `\ ` back; carve-rs and carve-php write `\`. This one is already `markup-carve/carve#1027` and is called out here only so it is not counted twice.

### 3. A comment directly after an inline construct loses its separating space

Input:

```
{,y,} %% c %% c \# <- | a | b |
```

carve-rs writes `{,y,}%% c ...`, gluing the comment to the subscript. carve-js and carve-php keep the space. The same shape appears with a highlight:

```
::: note =h= %% c =h= {^x^} !img \# =>
```

carve-rs writes `::: note =h=%% c ...`, the other two keep the space.

### 4. A caption line after a definition term gets a blank line, or does not

Input:

```
\-\-
:: term
^ cap
\.
```

[...]

PR #1043, opened 2026-08-09
The ragged-table delimiter row gets an expected-output file

### LINKED ISSUE #1042: All three engines write a delimiter row wider than its header, and the Markdown target stops being a table
PART 11 §10b rules on the delimiter row as well as on the data rows:

> Where a delimiter row is required to promote the first row to a header, that
> delimiter carries exactly one cell for each cell in the HEADER ROW, not one for
> each column reached by a wider body row. Its job is to mark that row as a
> header; extending it to the table width both manufactures header cells and can
> stop the delimiter from matching the row it belongs to.

Every engine sizes it from the TABLE width on the Markdown target. Found while
clearing markup-carve/carve#1040, where the data-row half of §10b was the fix for
two rows.

## Measured

Corpus `284-a-ragged-table-keeps-each-row-s-cell-count-3`, which is in the corpus
today:

```
| h |
|---|
| |x |
```

Its HTML is a two-column table with a one-cell header:

```html
<table>
  <thead><tr><th>h</th></tr></thead>
  <tbody>
    <tr><td></td><td>x</td></tr>
  </tbody>
</table>
```

`carve --markdown` on carve-js `e35cfcb`, carve-rs `fa4d3f7` and carve-php
`e8daf1e` - byte-identical in all three:

```
| h |
| --- | --- |
|  | x |
```

Through python-markdown 3.5.2 with `extensions=['tables']`:

```html
<p>| h |
| --- | --- |
|  | x |</p>
```

and through `marked`:

```html
<p>| h |
| --- | --- |
|  | x |</p>
```

A delimiter row whose cell count differs from the header row's is not recognized
by either reader, so the table is not a table - the document publishes as a
paragraph of pipes. A span-free shape reaches it too:

```
|=a|
| x | y |
```

writes the same wider delimiter in all three, with the same result.

## Why it is filed rather than fixed

It is not one of the twelve rows markup-carve/carve#1040 covers, and it is the
shape that ticket's screen describes one level up: all three engines agree, so
`Cross-engine render comparison` is GREEN on this document while every reader
refuses it. Narrowing it in one engine would put a shape the corpus currently
agrees on into disagreement, which is the failure §8a's "ALL THREE ENGINES LAND
IT IN ONE RELEASE" paragraph is about - so it wants one change across carve-js,
carve-rs and carve-php, landed together.

## What the fix is

Size the delimiter from the header row's own cell count, which is what §10b
already says. Both halves of §10b then hold on this target: a row keeps its own
cells, and the delimiter matches the row it promotes.

The corpus case above should gain a `.md` expected-output file at the same time,
so the answer is pinned against the spec rather than against engine agreement -
the same treatment markup-carve/carve#1041 gave the other twelve.

dereuromark:
Corpus `284-a-ragged-table-keeps-each-row-s-cell-count-3` gains a `.md`
expected-output file carrying the bytes PART 11 section 10b requires, so the
Markdown target's answer on a ragged table is pinned against the spec instead of
against engine agreement.

The clause:

> Where a delimiter row is required to promote the first row to a header, that
> delimiter carries exactly one cell for each cell in the HEADER ROW, not one for
> each column reached by a wider body row.

The case is a one-cell header over a two-cell body row:

```
| h |
|---|
| |x |
```

All three engines sized the delimiter from the TABLE width and emitted

```
| h |
| --- | --- |
|  | x |
```

which neither python-markdown 3.5.2 (with the `tables` extension) nor marked
16.4.2 reads as a table - both render a paragraph of pipes, so the document loses
its table. The pinned bytes are what section 10b asks for and what both readers
accept:

```
| h |
| --- |
|  | x |
```

This is the shape the corpus README warns about: engine agreement is a necessary
invariant and not a sufficient one, and it "cannot tell 'all three are right'
from 'all three are wrong'". The render comparison scored this row green
throughout while every reader refused the output. The same treatment
markup-carve/carve#1041 gave the twelve markup-carve/carve#1040 rows.

The engine side lands in the same release: markup-carve/carve-js#941,
markup-carve/carve-rs#858 and markup-carve/carve-php#1117.

Closes #1042


dereuromark:
Reduced this branch to the one file `#1042` asks for, and dropped three things
it had picked up. Recording why, so none of it is lost silently - the previous
revision is still on the PR timeline.

**Dropped: corpus `286-a-caption-marker-after-a-soft-break-is-ordinary-text` and
`287-an-escaped-space-at-line-end-is-canonical-inside-a-list`.** Both were
hand-written into `tests/corpus/`, and the corpus is GENERATED from
`docs/examples/`. `npm run corpus:build` refuses them outright, which is what the
red `spec-corpus conformance` job was reporting:

```
generate-corpus: the corpus numbering is APPEND-ONLY.

  2 existing categories are gone from the examples:
    286-a-caption-marker-after-a-soft-break-is-ordinary-text
    287-an-escaped-space-at-line-end-is-canonical-inside-a-list
```

The 286 pair is also wrong on its own terms: its `.html` said

```html
<p>para</p>
<p>^ cap</p>
```

while its `.test` sibling in the same commit said

```html
<p>para
^ cap</p>
```

Measured against the engines, the `.test` file is right - a caption marker after
a soft break is ordinary text in the SAME paragraph, so there is one paragraph
and not two. Both readings cannot be pinned at once, and a fixture that
contradicts its own mirror pins nothing.

The behavior those two cases were reaching for is real and is fixed:
`markup-carve/carve-php#1113` in `markup-carve/carve-php#1117`, and
`markup-carve/carve-rs#855` in `markup-carve/carve-rs#858`. Each is covered by a
regression in its own repo. If they want corpus coverage as well, it has to
start from a `::: compare` block appended to the END of
`docs/examples/edge-cases.md`, which is the only route the generator accepts.

**Dropped: turning on `property:check --engines` in
`.github/workflows/ast-conformance.yml`.** That gate belongs to `#1028`, and
whether it can be turned on is a measurement rather than an assumption. Measuring
it is the second half of the work on this ticket and the result goes on `#1028`,
not here.

**Dropped: the `680` to `682` and `880` to `882` counts in
`docs/implementation-comparison.md` and `docs/index.md`.** Those followed from
the two corpus cases above, and with those gone the counts are unchanged - `npm
run corpus:build` still writes 880 pairs on this branch. Adding a `.md`
expected-output file moves only the `markdown: fixtures=N` line, and that line is
already behind: the snapshot says `fixtures=1` while ten such files exist on
`main` after `#1041`. Correcting it wants a regenerated snapshot rather than a
hand edit, so it is left alone here.

What remains is the single file the ticket names, with the bytes section 10b
requires.

PR #1082, opened 2026-08-10
Specify bidi stripping for all presentation targets

### LINKED ISSUE #1076: Section 26 requires rendered text to strip bidi overrides; only the HTML target does
PART 9 section 26 requires rendered text to strip the bidi-override and isolate controls. Only the HTML target does. The Markdown, plain-text, ANSI and canonical-Carve targets emit every one of them unchanged.

## Measured

`a <U+202E> b` through each target, checking whether the control survives in the output. carve-js `004361b`:

| target | LRE 202A | RLE 202B | PDF 202C | LRO 202D | RLO 202E | LRI 2066 | RLI 2067 | FSI 2068 | PDI 2069 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| html | stripped | stripped | stripped | stripped | stripped | stripped | stripped | stripped | stripped |
| markdown | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT |
| plain | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT |
| carve | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT |
| ansi | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT | KEPT |

Nine of nine on four targets. Confirmed on a second engine - published carve-php `0.1.4` strips U+202E on HTML and keeps it on both markdown and plain - so this is shared behaviour rather than one implementation drifting.

## Why this reads as a defect rather than an intended scope

Section 25 opens `SECURITY REQUIREMENTS -- NORMATIVE (HTML output target)`. Section 26 opens `TROJAN-SOURCE / INVISIBLE-UNICODE HARDENING -- NORMATIVE (CVE-2021-42574 class)` with **no target qualifier at all**, and its rule reads:

```
RENDERED TEXT and CODE-SPAN / CODE-BLOCK content strip the
bidi-override / isolate controls (U+202A..U+202E, U+2066..U+2069)
```

The spec demonstrably knows how to scope a rule to HTML, and chose not to here. `docs/security.md` heads the section "Invisible Unicode and Trojan Source (**always on**)" and `docs/index.md` advertises "Always-on ... Trojan-Source stripping".

The stated justification is HTML-shaped - "an HTML parser decodes `&#x202e;` back to the live control, so removal is the only DOM-inert mitigation" - which may be why the implementation went that way. But the attack is not HTML-shaped. A terminal honours bidi overrides directly, so the ANSI target hands the original CVE-2021-42574 display attack to exactly the surface it was discovered on. Markdown and plain text are passed downstream to readers nobody here controls.

## The one target with a real argument for keeping them

`carve` is the canonical writer, and a writer that drops content is lossy. Worth noting that stripping there would NOT break the round-trip invariant - `to_html(fmt(x)) == to_html(x)` holds either way, because the HTML target strips them regardless - but "rendered text" is a strange description of a canonical writer's output, and this is the one target where the rule's intent is genuinely unclear.

The other three are renderers producing rendered text, which is what section 26 names.

## What needs deciding

Whether section 26 binds every target or only HTML. If every target, three renderers need the strip and the corpus needs a case per target - there is currently no non-HTML fixture for this shape, which is why nothing caught it, the same coverage gap as `markup-carve/carve#1074`. If only HTML, section 26 needs the scope qualifier section 25 already carries, and both docs pages need their "always on" claim narrowed.

Either way `carve` deserves an explicit sentence rather than being decided by omission.

No code written. Found by sweeping which characters each renderer transforms, per `markup-carve/carve#755`; the sweep harness was validated against three known-transforming controls first, since a sweep reporting "nothing found" is otherwise indistinguishable from a broken one.

dereuromark:
Closes #1076.

Defines HTML, Markdown, plain-text, and ANSI as presentation targets that strip bidi override/isolate controls, while canonical Carve preserves source. Adds target-specific corpus fixtures and documents the security boundary, including the independent §25/§26 control classes.

The lint rule ID is normative here; its validation-table row follows with the JS implementation so the spec contract test never advertises a rule absent from the pinned engine.


dereuromark:
The fixtures this adds assert behaviour no engine has yet, so merging it alone turns `AST conformance` red. Same shape as the sequencing that broke main earlier today, flagged before rather than after this time.

## What the fixtures want, and what the engines do

Source is `a<U+202E>b`. The new fixtures:

```
120….md    "ab"          stripped
120….txt   "ab"          stripped
120….ansi  "ab"          stripped
120….fmt   "a<U+202E>b"  preserved
```

That is the right design and it answers the canonical-writer question `carve#1076` raised - presentation targets strip, the writer preserves source. The `.fmt` row already matches.

The other three do not. Measured just now against carve-js `main` and carve-rs `main`:

```
             carve-js        carve-rs
markdown     a<U+202E>b      a<U+202E>b     fixture wants "ab"
plain        a<U+202E>b      a<U+202E>b     fixture wants "ab"
ansi         a<U+202E>b      a<U+202E>b     fixture wants "ab"
carve        a<U+202E>b      a<U+202E>b     matches
```

carve-php was measured on the same shape earlier and behaves identically. So all three engines mismatch these three fixtures, nine scored comparisons in total.

## Why that is red rather than just pending

`compare:impls` scores every corpus case that has a fixture for a target, and `check-compare-report.mjs` fails the job on `mismatches > 0`. These fixtures create four new scored comparisons per engine where three currently disagree. `crossImplDiffs` stays 0 - the engines agree with each other, they just all disagree with the new goldens - so the failure will read as `rust=3 js=3 php=3` against this one document.

## Sequencing

The engines need the strip before, or with, these fixtures. Three ways:

- Land the engine change first in all three, then this. Main stays green throughout.
- Land this and the three engine PRs close together and accept a short red window, which is what happened this morning between `carve-php#1142` and its two follow-ups.
- Split this PR: the prose and `docs/` now, the four fixtures with the engine work.

Not touching the PR either way - the ordering is yours. Happy to do the engine side on the next tick if that helps; it is a small change in the same `escapeText`/`escape_text` seam the ampersand work just went through, plus whatever the ANSI path needs since that target strips C0 already and would gain the bidi range.

One detail worth confirming while the fixtures are fresh: `carve#1079` recorded that all three engines also let DEL and the C1 controls through on the `carve` target while stripping them on the render targets. If "presentation targets strip, the writer preserves" is now the rule, that behaviour is consistent with it - worth a sentence in the same clause so the two control classes are not read as having different answers.

dereuromark:
One thing for whoever writes the engine side, found while measuring the fixtures.

**The bidi range is already spelled more than once per engine.** In carve-js:

```
src/render-html.ts:1794        const BIDI_CONTROL_RE = /[‪-‮⁦-⁩]/g
src/table-of-contents.ts:45    s.replace(/[‪-‮⁦-⁩]/g, '')
src/heading-ids.ts:389         /[‪-‮⁦-⁩​‌‍⁠﻿­]/gu
```

Two identical copies plus a deliberately wider one for ids, which section 26 does describe as a different set. Adding the Markdown, plain and ANSI strips as three more literals would make five spellings of one rule in one engine, and the class of defect that produces is well represented in this repo's history - `markup-carve/carve#755` collects them.

Worth exporting one helper - the repo already has small single-purpose modules like `trim-non-nbsp.ts` and `deny-listed-destination.ts`, so `bidi-controls.ts` would be idiomatic - and having render-html and table-of-contents import it too, so the count goes to one rather than five. The id set stays separate and deliberately wider.

Not doing that here: the strip cannot land in one engine alone. `markdown`, `plain` and `ansi` are compared engine-against-engine and gated on `crossImplDiffs`, so carve-js stripping while carve-rs and carve-php do not is red for a different reason than the fixtures are. All three move together or none do, which is the same constraint the ampersand change just went through.

Happy to take the three-engine change as one unit once this PR's shape is settled.

PR #1086, opened 2026-08-10
Pin the ANSI output of the heading that diverged

dereuromark:
Case `119-…-3` is `# A<U+202E>B<U+200B>C` and had only an `.html` fixture, so its ANSI output was checked by engine agreement alone. That is why a real divergence sat there until the nightly reported it:

```
carve-js   45 bytes, a five-cell heading rule
carve-rs   42 bytes, a four-cell rule
carve-php  42 bytes, a four-cell rule
```

Fixed in `markup-carve/carve-js#967` - the ANSI widths were computed before the bidi strip that runs over the assembled output - and all three engines now emit the same 42 bytes, verified by rendering the case through each.

## What the fixture buys, stated honestly

It would NOT have caught the original defect any sooner. Engine agreement is precisely what was broken, and `crossImplDiffs` reported it on the first run after the change.

What it catches is the other direction: agreement cannot distinguish "all three are right" from "all three are wrong", and it cannot see a regression that moves every engine the same way. There are only three `.ansi` fixtures in the whole corpus, so that target is almost entirely on agreement, and this is the shape that just demonstrated a defect can live there.

## Scope

Only `-3` gains a fixture. The siblings are `# Café` precomposed and `# Cafe` plus a combining acute - NFC cases with no bidi control, which do not exercise this shape. The rule is not new: §26 removes the override from rendered text and keeps the zero-width space, which is what the bytes record, and all three engines produce it.

`tests/corpus-targets.test.mjs` and `tests/fixture-bytes.test.mjs` both pass with the file in place.

Sweep lock, since this is corpus.


dereuromark:
Verified in the gate rather than only locally. Run `31444129658` against `a00cd2b`, the first to include this fixture:

```
rust: pass=935/935 mismatch=0
js:   pass=935/935 mismatch=0
php:  pass=935/935 mismatch=0

ansi: compared=892 diffs=0 errors=0 fixtures=4
cross_impl_diffs=0
```

Two things worth naming. The scored denominator moved 934 to 935 and the ANSI fixture count moved 3 to 4, so the file is actually being picked up - a fixture that exists but is never read would be exactly the shape this repo keeps finding, and adding one is a way to introduce it. And `ansi: diffs=0` is the divergence this case was added for, gone.

Everything else is clean in the same run: every target 0 diffs, and `roundtrip_diffs=0 semantic_failures=0 idempotence_failures=0 cross_read_failures=0` over 892 documents.

Issue #315, opened 2026-07-22
0.2: blocks do not interrupt open paragraphs


dereuromark:
## Tier

Tier-1 (core language)

## Target

0.2

## Decision

Adopt Djot's rule that **block openers do not interrupt an open paragraph**. A heading, quote, table, fence, definition, attribute, comment, div, or extension block needs block position (normally a preceding blank line); otherwise its marker remains paragraph text.

```carve
intro
# Heading
```

is one paragraph. Adding a blank line before `# Heading` makes it a heading.

## Why

This replaces Carve's visible/list/image/fence/invisible interruption matrix with one author rule:

```text
paragraph open + nonblank -> append
paragraph open + blank    -> close
no paragraph open         -> classify blocks
```

It is hard-wrap safe, applies automatically to extensions, and removes repeated interruption state from paragraphs and nested containers. The source break is appropriate for 0.2. A partial hybrid is explicitly out of scope because it would retain most implementation complexity while losing Markdown-familiar behavior.

### Pros

- One uniform rule at top level and in quotes, lists, definitions, divs, and extensions.
- Block-looking wrapped text cannot silently change document structure.
- New block extensions do not need to join an interrupter taxonomy.
- Paragraph/container logic becomes materially easier to specify, test, and maintain.

### Cons

- Breaks Markdown expectations for adjacent headings, quotes, thematic breaks, tables, and fences.
- Existing 0.1 documents may change meaning without inserted blank lines.
- Canonical formatting must consistently separate paragraph/block boundaries.
- Heading direction stays asymmetric: `# H\ntext` is heading + paragraph, while `text\n# H` is one paragraph.
- Fence matching, indentation, nesting, list content columns, captions, and attributes remain inherently complex.

## Measured scope / estimate

The reference-spec prototype changes 70 of 892 existing corpus examples (7.8%) in addition to the focused paragraph cases. The golden corpus passes 892/892 after updating those expectations.

| Area | Expected reduction | Relative effect |
|---|---:|---:|
| Normative grammar/spec | 100–180 lines | ~1–2% overall; ~20–30% of paragraph/container semantics |
| Executable spec | 80–180 net lines | ~3–7% |
| carve-js block parser | 250–500 net lines | ~3–6% of parser; ~15–25% of block-state logic |
| Explanatory docs | 150–300 lines rewritten/removed | interruption behavior is currently cross-cutting |
| Tests | Little net reduction | expectations change; coverage must remain |

[...]

PR #1089, opened 2026-08-11
0.2: require block position after paragraphs

dereuromark:
Implements #315 in the 0.2 `next` line.

- replaces the interruption matrix with uniform paragraph extent
- updates the normative grammar, executable oracle, generated corpus, and docs
- retains tight nested-list content-column and caption attachment exceptions
- updates formatter fixtures for literal misplaced openers

Validation: `npm test` (1,916 pass, 1 intentional skip), shared corpus 892/892.


dereuromark:
## The conformance failure is a stale engine pin, not a defect in this change

`spec-corpus conformance` fails on `tests/malformed-colon-fence-absorbs.test.mjs`:

```
expected  <p>text not a div :::</p>
actual    <p>text not a div</p> <div> </div>
```

That test measures the PUBLISHED engine (`import { carveToHtml } from '@markup-carve/carve'`), and this branch pins it to a pre-0.2 commit:

```json
"@markup-carve/carve": "github:markup-carve/carve-js#56735159d54a58ef91210c1577da1bfb6c0bf940"
```

Measured directly:

```
carve-js at the pinned commit          <p>text not a div</p> <div> </div>
carve-js on feat/315-noninterrupting   <p>text not a div :::</p>
```

The pinned engine still interrupts an open paragraph on a bare `:::`, which is exactly the behaviour this ticket removes. The actual output is the OLD rule; the expectation is the new one. Nothing in `resources/` or `scripts/spec/` is producing it.

So the sequencing is the blocker: markup-carve/carve-js#975 has to land, and this branch's pin has to move to a commit carrying it, before this suite can go green. Until then the failure is expected and says nothing about the spec change.

## On the codex finding for this PR

Codex raised a P1 against `scripts/spec/layout.mjs:1743-1749`, arguing `findColonCloser` classifies nested fences without consulting paragraph extent, using `::: outer` / `text` / `::: inner` / `body` / `:::` as the example.

That is a claim about the ORACLE. This failure is not evidence for it, because the failing test never reaches the oracle - it imports the published package. The two are independent, and the finding is neither confirmed nor refuted by the red build. It is worth checking on its own terms against `scripts/spec/`, with the pin question settled first so a real oracle defect is not masked by an expected engine mismatch.

dereuromark:
## Verified: the oracle's colon-boundary scan does diverge, and it swallows content

I said in the previous comment that codex's `findColonCloser` finding was neither confirmed nor refuted by the red build. Measured now against `scripts/spec/` on this branch, it is REAL - the predicted symptom is not what happens, but the mechanism is right and the consequence is worse than described.

Input:

```
::: outer
text
::: inner
body
:::

after
```

```html
oracle    <div class="outer"> <p>text ::: inner body :::</p> <p>after</p> </div>
carve-js  <div class="outer"> <p>text ::: inner body</p> </div> <p>after</p>
```

Two things differ, and the second is the damaging one:

1. The closing `:::` lands INSIDE the paragraph in the oracle and closes the div in the engine.
2. **`after` is swallowed into `outer` by the oracle.** The container never sees a closer and runs to end of input, so every following block is captured.

## Why, and the control that isolates it

Removing only the labeled line makes them agree:

```
::: outer
text
body
:::

after
```

```html
oracle    <div class="outer"> <p>text body</p> </div> <p>after</p>
carve-js  same
```

So it is not "a bare `:::` after prose" in general - that closes correctly in both. Walking the shapes:

```
::: outer / text / :::                     -> closes, both
::: outer / text / more / :::              -> closes, both
::: outer / text / ::: inner / :::         -> oracle absorbs both fences
::: outer / text / ::: inner / body / :::  -> oracle absorbs both fences
```

The trigger is an absorbed LABELED opener. Once `::: inner` has been read as paragraph prose by the renderer, the boundary scan has still counted it as opening a container, so the next `:::` is consumed against `inner` instead of `outer`. The renderer and the scan disagree about whether that line opened anything, which is exactly the mismatch codex named at `scripts/spec/layout.mjs:1743-1749`.

## Scope

This is independent of the engine-pin problem in the previous comment. That one explains the `malformed-colon-fence-absorbs` failure and clears once markup-carve/carve-js#975 lands and the pin moves. This one is a defect in the change itself: the oracle is the executable spec, so it disagreeing with the reference engine on a shape the new rule creates is exactly the class of thing the oracle exists to catch.

Worth a corpus case once fixed - nothing pins a labeled colon opener absorbed into a paragraph, which is how the two halves came to disagree unnoticed.

PR #1103, opened 2026-08-11
ci: close superseded downstream bump drafts

### LINKED ISSUE #1101: bump-downstream pins each PR to the rev at dispatch, so drafts go stale before they are read
`bump-downstream.yml` pins each downstream PR to the carve rev that existed when the job dispatched. Carve main keeps moving, so by the time a human looks at the draft the target can already be behind - and the workflow has no step that notices.

Today's batch shows both failure modes at once. Four drafts were opened against `da8ed8c`; carve main is now `8490461`, two commits further on:

| draft | that repo's pin | verdict |
|---|---|---|
| carve-php#1173 | `8490461` | superseded - merging would move the pin **back** 2 commits |
| carve-js#984 | `8490461` | superseded - same |
| carve-grammars#211 | `a00cd2bf` | wanted, but target stale by 2 |
| tree-sitter-carve#190 | `723bcb31` | wanted, but target stale by 2 |

carve-rs#901 was the fifth and was closed for exactly this.

So half the batch was obsolete within hours of being opened, and the two that are still wanted would each need a second bump immediately after merging. The drafts stay open reporting a red state that no longer describes anything, which is the failure mode carve#755 tranche 9 calls a check whose answer nobody acts on: each of these ran, answered correctly, and the answer aged out before it was read.

Worth separating two things the current design conflates:

1. **The pin is captured at dispatch, not at merge.** Nothing re-evaluates it. A bump PR is the one kind of PR whose whole content is a moving target, so a stale pin is the expected steady state, not an edge case.
2. **A superseded draft is indistinguishable from a pending one.** Both are "open draft, red checks". Deciding which needs an `is-ancestor` check run by hand, per repo - and that check is one command:

```
git merge-base --is-ancestor <pr-target> <repo's current pin>   # true => superseded
```

Options, roughly in increasing order of effort:

- **Detect and close.** A scheduled step re-runs the ancestor check on every open bump draft and closes the superseded ones with a comment. Cheap, and removes the two dead drafts in this batch without touching the live two.
- **Retarget on re-dispatch.** When the workflow re-runs and finds an open bump draft for the same repo, force-push the branch to carve main's current head instead of opening a second PR.
- **Target a moving ref.** Bump to carve `main` at merge time rather than a captured sha. Removes the staleness class entirely but gives up the reproducibility of a pinned bump, so probably not worth it.

The first alone would have removed half of today's noise.

Not filed as a defect in any engine: both live drafts turned out to need no engine or grammar change at all (12 goldens plus one ratchet number in carve-grammars, two `covered` entries in tree-sitter-carve, both verified green). The cost here is the review time spent rediscovering that per batch.

dereuromark:
Closes #1101.

Adds a six-hour cleanup pass for open draft automation/bump-spec PRs. It closes a draft only when its Carve pin is a strict ancestor of the downstream main pin and the PR changes no file outside the spec submodule. Drafts carrying implementation work, equal pins, and divergent pins remain open.

Scheduled cleanup does not run the expensive downstream bump or matrix-coverage jobs. The cleanup matrix is regression-tested to match the bump matrix exactly.


dereuromark:
Reviewed this against the four real drafts that motivated carve#1101. The logic is right, and it is red for one small reason.

**The suite fails because the new test never runs.**

```
npm test
# tests 1927   # pass 1925   # fail 1
    test file(s) never run by npm test: close-superseded-bump.test.mjs
```

`tests/close-superseded-bump.test.mjs` is not in the `test` script's file list, and there is no glob runner, so the manifest test caught it. Adding the filename to `package.json` should be the whole fix. Worth noting the gate did its job here - an orphan test file is exactly the kind of check-that-cannot-fail carve#755 is about, and this repo already refuses to ship one.

**The safety guards have no test - only the destructive path does.**

The one script test asserts that a superseded submodule-only draft IS closed. Nothing pins the three branches that decide NOT to close, and those are the ones standing between this job and a wrongly-closed PR. I ran them through the same fake-`gh` harness the existing test uses:

```
A superseded, submodule-only     CLOSED   markup-carve/example#7: closed superseded draft
B equal pins                     kept     draft and main both pin e745c85; not superseded
C main BEHIND the draft          kept     e745c85 is not an ancestor of main's 9a7a0d8
D superseded + impl work         kept     ::warning:: ... but also changes src/parse.rs
```

All four behave correctly, so this is not a defect report - it is that the verification disappears when I close this tab. B, C and D are cheap to add beside A, since the harness is already written.

D is the one I would least want to lose. carve-rs#901 carried a commit titled "+ impl attempt" whose diff touched only the submodule, so it would have closed correctly - but the next one might really carry the implementation, and that guard is the only thing between this job and deleting it with `--delete-branch`.

**Checked against the real cases.** The direction is right in both senses:

```
carve-rs#901, carve-php#1173, carve-js#984   pin behind main   -> would close (all three were superseded)
tree-sitter-carve#190, carve-grammars#211    main behind pin   -> left open  (both bumps still wanted)
```

That last row is what case C protects, and it is currently the untested one.

**One edge worth a thought, not a change.** The ancestry test runs in the carve checkout and fetches `origin main` when an object is missing. A downstream that pins a commit on a carve BRANCH rather than main leaves both fetches short, `merge-base` errors, and the `if !` treats it as "not an ancestor" - so the draft stays open, which is the safe direction. The message will say "is not an ancestor" when the truth is "could not be found". Harmless today; misleading the day someone debugs it.

Issue #1114, opened 2026-08-12
Add {:fr} semantic language attributes in 0.1.x


dereuromark:
## Goal

Add a compact semantic natural-language attribute to the Carve 0.1.x language:

```carve
The title is [Le Bon Usage]{:fr}.

{:de .quotation}
> Das ist ein deutschsprachiger Absatz.

[Mongolian]{:mn-Cyrl-MN}
[project terminology]{:x-acme}
[language deliberately unknown]{:}
```

`:` is an attribute sigil. `{:fr}` is exact syntax sugar for `{lang=fr}`.

The empty form `{:}` means **“the language of this content is explicitly
unknown.”** It desugars to `lang=""`. This matters inside content that would
otherwise inherit a surrounding language: omitting a language attribute keeps
the inherited language, while `{:}` stops that inheritance. It is also distinct
from `{:und}` (the BCP 47 `und` tag: an undetermined language) and `{:zxx}`
(the BCP 47 `zxx` tag: no linguistic content).

This is related to jgm/djot#5, but does not depend on Djot adopting the shorthand. Djot already supports the long `{lang=fr}` form.

## Versioning decision

Target **0.1.x**, as a deliberately narrow allocation of an unclaimed attribute-syntax slot.

This is technically an observable parsing change: `[x]{:fr}` is literal in current 0.1 and will become an attributed span. We accept that limited compatibility exception only because:

- `{:TAG}` has never been documented or assigned a meaning;
- attribute names cannot currently start with `:`, so it cannot collide with a supported attribute atom;
- recognition is restricted to structurally well-formed BCP 47-shaped tags and the exact empty form;
- the deferred braced-symbol spelling `{:name:}` remains literal;
- malformed candidates fail visibly as literal source rather than being partially consumed.

Before implementation is called ready, audit the organization and public GitHub code search for attached or standalone `{:…}` in `.crv` files. Record the query and result in the spec PR. If meaningful literal use is found, move this ticket to 0.2 instead of inventing contextual exceptions.

## Normative syntax

Add `language_attribute` to both inline and block attribute lists:

```ebnf
attribute          = language_attribute | id_attribute | class_attribute
                   | key_value_attribute | boolean_attribute ;
language_attribute = ':', [language_tag] ;
language_tag       = language_subtag, {'-', language_subtag} ;
language_subtag    = ASCII_alphanumeric{1,8} ;
```

Normative consequences:

- ASCII letters and digits only within subtags.
[...]

PR #1115, opened 2026-08-12
spec: add the {:TAG} language attribute production

dereuromark:
Draft for markup-carve/carve#1114. Grammar only, deliberately - see the scope note at the end. Opened as a draft because the ticket is a spec-change awaiting a ruling, and because one part of it needs reconciling before any of this should land.

## The audit the ticket asks for, recorded here as it asks

Queries:

```
gh search code --owner markup-carve --extension crv '{:'
gh search code --extension crv '{:'
```

Both return exactly one file, `tests/corpus/138-unclaimed-openers-stay-literal.crv`. Cross-checked with a local grep of this checkout at `48356f3`, because a search index can lag behind:

```
1 of 970 .crv files contain `{:`
```

Same file. Its content is the deferred braced-symbol form:

```
[>foo]

{:tada:}
```

```html
<p>[&gt;foo]</p>
<p>{:tada:}</p>
```

It is unaffected: a trailing `:` is not a subtag character, so the attribute is malformed and the block stays literal.

**Result: no meaningful literal use exists to break.** The ticket's own criterion for staying in 0.1.x is met.

## The premise measured in all three engines

Eleven shapes, byte-identical output from carve-js `132bca2`, carve-php `6621ef0` and carve-rs `2e5c88c`:

```
[x]{:fr}          <p>[x]{:fr}</p>
[x]{:}            <p>[x]{:}</p>
[x]{:de-CH}       <p>[x]{:de-CH}</p>
[x]{:tada:}       <p>[x]{:tada:}</p>
[x]{:en_US}       <p>[x]{:en_US}</p>
[x]{:français}    <p>[x]{:français}</p>
[x]{: fr}         <p>[x]{: fr}</p>
[x]{lang=fr}      <p><span lang="fr">x</span></p>
```

`{lang=fr}` already produces the desugaring target, so the semantics half needs nothing new. The `{@fr}` claim in the ticket also holds: all three parse it as a mention even between literal braces.

## Two things the ticket states that measurement corrects

**"Malformed candidates fail visibly as literal source"** does not describe the mixed form. All three engines render this identically today, and it is not literal:

```
[x]{#id :fr .cls}
```

```html
<p>[x]{<span class="tag"><strong>#id</strong></span> :fr .cls}</p>
```

The block fails as an attribute block, the text falls back to inline parsing, and `#id` becomes a tag. So a document containing that form renders a tag span today and an attributed span afterwards - a second observable change beyond plain literalization, worth naming in the compatibility section.

**The cost is not only the AST surface.** "No new AST node, no new AST field, no profile-vocabulary change" is accurate and measures one dimension. `docs/dismissed-syntax.md` prices the other one for exactly this opener: a new inline form ripples through the tree-sitter grammar, the TextMate/Prism/highlight.js grammars, eight editor plugins, the corpus and all three engines. Desugaring into an existing attribute does not reduce that bill.

## What still needs reconciling

`docs/dismissed-syntax.md` reserves this opener as a recorded decision, not an omission:

> **Decision:** Deferred, not rejected. The design is recorded here so it is not re-litigated; `{:` stays unclaimed. If a real document ever needs an intraword symbol, this is the form to add

That entry relies on the same ground this ticket does - "an attribute block cannot start with `:`" - as the reason the symbol form was available. The two are not syntactically exclusive, since a symbol ends `:}` and a language tag cannot, and the grammar comment in this diff says so. But one of them holds the slot by a written decision, and that entry should be amended in the same change rather than left contradicting the grammar.

## Scope

Grammar only, and that is deliberate. An example in `docs/examples/` becomes a corpus document, and a corpus document no engine implements turns every downstream bump red until they do. The cases belong with the first engine implementation, not ahead of it.

Also worth knowing before anyone estimates the engine work: this is not the one-function change the desugaring makes it sound like. I patched carve-js's `parseAttrs` and `isValidAttrPayload` and verified both regexes accept `:fr` and still reject `{a:b}` - and `[x]{:fr}` still rendered literal, because a further gate on the inline span path rejects the block. Whoever picks up the engine side should budget for finding that rather than for one regex.

Verified on this branch: the spec suite passes, 1948 tests, 0 failures.


dereuromark:
Pushed the executable side of this production plus the two answers it left open. Rebased onto `main` first, since it had moved two commits.

## The production had nothing executable behind it

`resources/carve-core.ohm` had no rule for `{:TAG}`, so the oracle could not parse a document using the shorthand and the corpus had no case to pin it. `attrItem` now has a `langAttr` alternative, and the desugaring lives in the `parseAttrs` action and nowhere else: it returns the same `['kv', 'lang', TAG]` tuple every `lang=` produces, so merge order, last-wins, hardening and the serializer reach it with no case of their own. That is what makes "exact sugar" falsifiable rather than restated - if it were not exact, the collision fixtures below would say so.

## Collision, ruled

`:TAG` and `lang=TAG` are the same key, so RENDER ORDER already decides it: last value wins, slot stays at first appearance. Holds inside one block and across merged block-attribute lines, because the desugaring happens before either merge sees it.

```
[a]{:fr lang=de} [b]{lang=de :fr}
```

```html
<p><span lang="de">a</span> <span lang="fr">b</span></p>
```

The surviving value lands in the slot the first of the two opened:

```
[x]{k=1 :fr lang=de title=t}
```

```html
<p><span k="1" lang="de" title="t">x</span></p>
```

The clause adds no rule. It records that the shorthand needed none, which is worth writing down where a reader would otherwise look for a special case.

## Canonical form, ruled

The writer collapses a repeated `lang` to the value that survives the merge and spells it `:TAG`, so `{:fr lang=de}` formats to `{:de}`. An empty value is `{:}`, which is therefore also the canonical spelling of `{lang=""}`.

The `.fmt` byte fixture that would pin this is deliberately NOT here. Those fixtures are read through the pinned build by `tests/corpus-fmt-roundtrip.test.mjs`, which has no drift valve of its own - the same reason `283-an-empty-footnote-body-is-written-with-the-empty-sentinel` defers its fixture to a pin bump. It belongs with the engine work.

## Fixtures

Thirteen cases in three categories, appended so no existing number moves:

- `292-a-language-attribute-is-exact-sugar-for-lang` - the shorthand, subtag shapes, the empty form stopping inheritance inside a `{:de}` quote, mixed with the other attribute kinds, and on a block-attribute line.
- `293-a-malformed-language-tag-leaves-the-whole-block-literal` - underscore, leading and trailing hyphen, non-ASCII, the eight-character cap, `{:name:}`, and `{xml:lang=en}`.
- `294-a-language-attribute-and-lang-are-one-key` - the collisions above, two shorthands colliding, and the block-attribute-line merge.

Ten are declared in `resources/engine-pin-drift.txt`; the pinned build has no production for the shorthand and leaves those documents literal. The four `293` cases that are literal under both are not declared, which keeps the declaration honest about which half is drift.

## Two things measurement turned up

**Ohm's `alnum` follows Unicode.** A first cut accepted `{:français}` as a language attribute reading `français` - one of the shapes this production names as staying literal. The subtag alphabet is now spelled from `asciiLetter` and `digit`. The eight-character cap is eight optional characters rather than a check in a semantic action, because an action cannot fail a match: a ninth character has to have no rule to match, so the block fails at the `}` and stays literal, which is what PART 9 §14 already promises.

**The oracle does not merge a boolean with a key/value of the same name.** Filed as markup-carve/carve#1123 with the measurements. It emits the boolean unconditionally:

```
[x]{a=1 a}
```

executable spec:

```html
<p><span a="1" a="">x</span></p>
```

carve-js `8815874` and carve-php `d04f88c`, byte-identical to each other:

```html
<p><span a="">x</span></p>
```

Two attributes of one name is not valid HTML, and PART 4 already says a boolean is `name=""`, so the engines are right and the oracle is not. It predates this production and is not confined to `lang`, so it is fixed there rather than here - and this is why no fixture in `294` pins a boolean `lang`, even though the ticket names it as the third spelling of the key. The clause says so explicitly instead of leaving a rule with no fixture behind it.

## Gates

`npm test` 1963 pass, 0 fail. `npm run core:check` 918/918 conformant, zero refusals, zero defects, zero sentinel leaks. `npm run engine:report -- --check` reports "Drift matches what is declared" against the ten lines this branch adds. `fmt:check` and `claims:check` need the three engine checkouts and did not run locally; CI has them, and CI is green on all four jobs.

A correction to my own first run of those gates, since it is the failure mode this repo keeps writing tickets about. I first ran them against a `node_modules` symlinked in from another checkout, which held carve-js `79175a5` while this branch pins `5f5ae4c`. Every engine-dependent result was then about a build the branch does not pin: seven tests appeared to fail and `engine:report` appeared to report four undeclared slugs in `288` and `291`. With `npm ci` against the actual pin, all of it is clean and there is nothing pre-existing to report on `main`. The stale artifact, not the code, was the finding.

## Still open, and not decided here

`{: fr}` - a space after the sigil - parses as `lang=""` plus a boolean `fr`, which follows from `':', [ language_tag ]` exactly as written. markup-carve/carve-js#1005 raised whether a typo should instead leave the block literal. No fixture here pins either answer.

The reservation in `docs/dismissed-syntax.md` still needs amending in whichever change carries the ruling. Untouched, deliberately.

PR #1138, opened 2026-08-13
docs: the compact semantic span, where the rest of the page still contradicts it

dereuromark:
The compact semantic span shipped with a normative clause and HTML fixtures. Three consequences of it stayed behind.

## The boolean-attribute clause still teaches the old output

`resources/grammar.ebnf` (BOOLEAN ATTRIBUTES, PART 4) and `docs/divergence-from-djot.md` both illustrate "a bare word rendered `name=\"\"`" with the one bare word that is now consumed instead:

```
Press [Tab]{kbd} to indent.
```

```html
<p>Press <kbd>Tab</kbd> to indent.</p>
```

The same edit was made to `docs/blocks-and-attributes.md` in a previous PR and these two were missed. The example moves to a name outside the registry, both places point at the nine names, and the djot page gains the divergence as an entry of its own - `<span kbd="">` really is what Djot renders there.

## The non-HTML targets were prose only

PART 9 §10 states the AST keeps an ordinary span, that plain and terminal output render the content, and that the writer spells a consumed boolean `name=""`. No `.fmt`, `.txt`, `.md` or `.ansi` fixture existed beside any semantic-span case, so an engine whose writer dropped the attribute, or whose plain renderer leaked it, would pass every gate.

Four sidecars now pin it:

| fixture | pins |
| --- | --- |
| `97-boolean-attributes.fmt` | `Press [Tab]{kbd=""} to indent.` - the writer keeps the consumed attribute and spells the boolean |
| `97-boolean-attributes.txt` | `Press Tab to indent.` - plain renders the content, not the wrapper |
| `45-inline-extensions-9.fmt` | value forms plus a six-name combination survive the round trip |
| `45-inline-extensions-9.txt` | the whole registry renders as its content in plain text |

Verified byte-for-byte in all three engines before being written down, not only in the pinned build: carve-js `a59df0fa`, carve-php `520ea41`, carve-rs `410845c` produce every one of these four files exactly.

## Authors had nowhere to learn it

The feature reached the extension catalog and one sentence of the attribute list. `docs/blocks-and-attributes.md` gains a section mirroring the language-attribute section beside it: the registry, which three names keep their value and which six discard it, the fixed nesting order, what remains on the outer span, and the three things that are easy to get wrong - the wrapper survives a stripped attribute, the rule is scoped to an ordinary span (`` `c`{kbd} `` is `<code kbd="">`), and only HTML changes. The cheatsheet and the Markdown migration table get the entry, since a Markdown author reaching for `<kbd>` has no raw HTML to reach for here.

Every HTML example on those pages is the pinned engine's actual output, compared byte for byte.

### DIFF
--- a/docs/blocks-and-attributes.md
+++ b/docs/blocks-and-attributes.md
@@ -194,6 +194,58 @@ A language tag never sets writing direction. Direction follows the script, and a
 
 A tag that is not structurally well formed leaves the whole block as literal text rather than half-parsing it - `{:en_US}`, `{:-en}` and `{:français}` all stay visible in the output. And the sigil takes no padding: `{: fr}` (with a space) is the empty language attribute plus a separate boolean `fr`, not a language tag.
 
+### Semantic spans: `{kbd}`, `{abbr="…"}`
+
+On an **inline span**, nine attribute names are consumed and become the HTML element of the same name: `abbr`, `time`, `code`, `mark`, `samp`, `var`, `kbd`, `cite`, `dfn`.
+
+```carve
+Press [Tab]{kbd} to indent.
+```
+
+```html
+<p>Press <kbd>Tab</kbd> to indent.</p>
+```
+
+Three of them keep what you wrote: an `abbr` or `dfn` value becomes `title`, a `time` value becomes `datetime`.
+
+```carve
+[HTML]{abbr="HyperText Markup Language"} is a markup language.
+```
+
+```html
+<p><abbr title="HyperText Markup Language">HTML</abbr> is a markup language.</p>
+```
+
+On the other six a value only picks the wrapper and is **not** written out, so `[x]{cite="https://example.org/dune"}` renders `<cite>x</cite>` and the URL reaches no output at all.
+
+Several at once nest in a **fixed** order - `abbr`, `time`, `code`, `mark`, `samp`, `var`, `kbd`, `cite`, `dfn`, innermost first - regardless of the order you typed them, so no document can come to depend on the spelling:
+
+```carve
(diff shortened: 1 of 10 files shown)


dereuromark:
Verified the four sidecars against all three engines at their current main rather than taking the claim on trust - carve-js `a59df0fa`, carve-php `056eea0`, carve-rs `9130a7c`, byte-identical:

```
97-boolean-attributes.fmt    Press [Tab]{kbd=""} to indent.
97-boolean-attributes.txt    Press Tab to indent.
45-inline-extensions-9.fmt   [CSS]{dfn="" abbr="Cascading Style Sheets"}
                             [Noon]{time=12:00} [x]{code="" mark="" samp="" var="" kbd="" cite=""}
45-inline-extensions-9.txt   CSS Noon x
```

Merging on that basis. Two things in it are about to change, and both are handled in markup-carve/carve#1143 rather than here:

The `.fmt` bytes. PART 11 §6c makes a value-less attribute come back as the bare name, so those two files become `{kbd}` and `{dfn abbr=...}` / `{code mark samp var kbd cite}`. They are correct as written today and are updated with the pin bump, once every writer ships the clause - a fixture is read through the pinned build, so new bytes now would be red in this repo and in all three engines.

The two PROSE paragraphs, which are claims about the RULE rather than about the current build, so they are wrong the moment §6c lands: `resources/grammar.ebnf`'s "the canonical writer, which spells the same span `[Tab]{kbd=\"\"}`" and `docs/blocks-and-attributes.md`'s "a boolean as `name=\"\"`, so `[Tab]{kbd}` formats to `[Tab]{kbd=\"\"}`". #1143 is rebased onto this branch and carries both corrections.

Issue #1156, opened 2026-08-13
A tag immediately after a tag: one tag in every engine, two in the executable spec


dereuromark:
## What

Two `#tag`s written with nothing between them: all three engines make only the FIRST one a tag, the executable spec makes both.

```
[a]{#i#j}
```

carve-js `7d6a2bcf`, carve-php `b4a343b`, carve-rs `c0192d3` - byte-identical to each other:

```html
<p>[a]{<span class="tag"><strong>#i</strong></span>#j}</p>
```

executable spec (`scripts/spec`, same input):

```html
<p>[a]{<span class="tag"><strong>#i</strong></span><span class="tag"><strong>#j</strong></span>}</p>
```

The braces are incidental - they are literal text here, because `{#i#j}` is not an attribute block (PART 7 requires `space+` between two attributes, and markup-carve/carve-js#1029 with its php and rs companions make all three engines agree with the spec on that). What is left is ordinary prose containing `#i#j`, and the question is only how the tag syntax reads it.

## Why it surfaced now

It could not be reached before. `{#i#j}` parsed as an attribute block in every engine, so the text never went through the inline pass. Making the block literal is what exposed the second question underneath it, and the fix for the first one is what makes this one measurable at all.

## Which side is right

Not obvious, which is why this is a ticket rather than a patch.

**For the engines (one tag).** A tag is a word-ish token, and `#j` glued to the end of `#i` is not at a word boundary any more than `word#j` is. Carve's symbol shortcode already takes this position: `:name:` only opens at the start of content or after a non-word character, and that guard is described in `docs/dismissed-syntax.md` as load-bearing rather than incidental.

**For the spec (two tags).** `#` is not a word character, so the boundary the guard asks for is arguably present. If the rule is "a tag opens after any non-word character", then the `#` that ended the previous tag qualifies and both open.

The answer decides a second, more common shape too - `#a#b` in ordinary prose, with no braces anywhere - so it is worth settling on the general case rather than on the bracketed one.

## Work

- [ ] Decide whether a tag may open immediately after another tag.
- [ ] `markup-carve/carve`: state it where the tag boundary rule lives, and add a corpus case - `#a#b` in prose, not the braced form, since that is the shape an author writes.
- [ ] Whichever side moves: the executable spec, or all three engines.

## Note

[...]

PR #1157, opened 2026-08-13
spec: pin that two attributes need a separator

dereuromark:
`attribute_list` is `attribute, {space+, attribute}` (PART 7) and no corpus document contained an adjacent pair, so the rule had nothing behind it. The executable spec refused `{.a.b}` the whole time, all three engines accepted it, and every suite in every repo stayed green - markup-carve/carve#755's shape one layer down: not a check that fails open, but a rule with no case to fail on.

Maintainer ruling: the spec is correct. Engine fixes are markup-carve/carve-js#1029, markup-carve/carve-php#1236 and markup-carve/carve-rs#942.

## Four cases, and why each one is there

Three are the shapes that were accepted:

```
[x]{.a.b}    two classes touching
[x]{.a#i}    two KINDS touching
[x]{#i.c}    the same pair the other way round
```

**`{.a#i}` earns its place** - it is the one a strip-based validator still gets wrong after the obvious fix. carve-php's patterns replace each match with a SPACE, so adding the missing separator lookbehind to the class and id patterns fixed `{.a.b}` and `{#i.c}` and left `{.a#i}` accepted: by the time the id pattern ran, the class it followed had become the separator it needed. A category holding only `{.a.b}` would have passed that half-fix and shipped it.

**The fourth is the control.** `{k=a:b}` stays a valid block, because an unquoted value runs to the next whitespace and the colon inside it is not a separator question. Without it the category cannot distinguish "adjacency is refused" from "anything with two sigils is refused".

## Drift

The three literal cases are declared in `resources/engine-pin-drift.txt` - the pinned build still reads them as attributes. They clear with the three engine fixes and a pin bump.

## Not in here

`{#i#j}` is literal under this rule too, but its text then runs through the `#tag` inline syntax, where all three engines make only the first `#i` a tag and the executable spec makes both. That is tag adjacency rather than attribute adjacency, it was unreachable until the block stopped parsing as attributes, and a corpus case here would settle it by accident. Filed as markup-carve/carve#1156.

## Verification

`npm test` 1983 pass 0 fail. `npm run core:check` 936/936 conformant, zero refusals, zero defects. `engine:report --check` reports drift matching what is declared.


dereuromark:
Rebase needed, and the conflict is a category-number collision rather than a text clash.

This branch appends its section to `docs/examples/edge-cases.md` and gets **299**; markup-carve/carve#1150 landed first and took 299 for `the-semantic-registry-holds-no-element-carve-already-spells`. After `git rebase origin/main` this section becomes **300**, and three things move with it:

1. **The generated pairs.** `scripts/generate-corpus.mjs` will not renumber over a stale sidecar set, so delete this branch's `tests/corpus/299-two-attributes-need-a-separator-between-them*` (and its `tests/spec/299-*.test`) before running `npm run corpus:build`, or the build refuses and leaves orphans behind. I hit exactly this while renaming a section on carve#1150.
2. **`resources/engine-pin-drift.txt`** names slugs WITH their number, so every line this branch adds needs the new one.
3. **`docs/implementation-comparison.md`'s** "Corpus added since this run" list, same reason.

Then `npm test` should be green, and `npm run engine:report -- --check` should report drift matching what is declared in both directions.

Worth doing on top of markup-carve/carve#1158 rather than plain `main`: that bump moves the pinned build past the uppercase-`LANG` writer bug and is what makes `main` green again, so a rebase onto a red `main` will look like this branch broke `corpus-fmt-roundtrip` when it did not.

dereuromark:
Rebased onto current `main`. Four files conflicted and one thing changed substantively while this sat: **the engine fix landed**.

`package.json` now pins carve-js `fb7deec`, which IS markup-carve/carve-js#1029 - the fix this PR declared drift against. So the three literal cases reproduce under the pinned build, and `engine:report --check` refuses to keep them listed:

> a slug is listed and now reproduces -> the pin moved past it. Delete the line in the same commit that moves the pin.

The drift block is therefore gone rather than rebased. Two consequences, both mechanical:

- The three documents rejoin the core corpus, so the published comparison moves `693 / 693` to `697 / 697` and `corpus_pairs=697`. Same in-place count bump markup-carve/carve#1127 made for the same reason, engine revisions and timings untouched.
- The landing page count follows.

The other conflicts:

- `docs/examples/edge-cases.md` - both sides appended a new section at the same point. Verified the branch's side is a pure 64-line append onto the merge base, so the resolution is main's file plus that block, in order.
- `resources/engine-pin-drift.txt` - main's 44 `scope` entries kept, this branch's three dropped as above.
- `docs/implementation-comparison.md` - regenerated counts as described.

One renumber, and it is deliberate: main's `299-the-semantic-registry-holds-no-element-carve-already-spells` took 299, so this category is now **300**. Re-ran `corpus:build` with `CORPUS_RENUMBER=1` and no other category moved.

Verified after the rebase: `npm test` 1988 pass 0 fail, `engine:report --check` drift matches what is declared (44, all `scope`), `corpus:build` clean, `docs:build` clean.

Issue #1152, opened 2026-08-13
Three .fmt fixtures were withdrawn for the §6c rollout and nothing tracks their return


dereuromark:
## What

Three `.fmt` fixtures were withdrawn while PART 11 §6c was mid-rollout, and nothing tracks their return.

- `297-the-language-sigil-takes-no-padding.fmt`, pulled by carve#1143.
- `97-boolean-attributes.fmt` and `45-inline-extensions-9.fmt`, pulled by carve#1144.

The reason was sound: `tests/corpus-fmt-roundtrip.test.mjs` reads a `.fmt` through the pinned build and has no drift valve, so no spelling of those files is green on both sides of the writer change - `[Tab]{kbd=""}` before §6c, `[Tab]{kbd}` after.

## The gap

§6c now has **zero executable coverage**. `resources/engine-fmt-drift.txt` carries three unrelated slugs and not these, `resources/engine-pin-drift.txt` is about reading rather than writing, and the only record that the files should come back is a sentence in two commit messages:

> All three come back in the pin bump, with the bare spelling, once every writer emits it.

A clause with no fixture is the shape this repo keeps removing rather than adding. Nothing here is wrong today; what is missing is the thing that makes it wrong LOUDLY if it is forgotten.

## Work

- [ ] Land the writer change in all three engines (markup-carve/carve-js#1025 merged; markup-carve/carve-php#1233 and markup-carve/carve-rs#938 open).
- [ ] Bump the pin past it.
- [ ] Restore the three `.fmt` fixtures in the same PR as the bump, with the bare spelling.
- [ ] Close this.

The pin bump is the natural home, and this ticket exists so the restore is a checklist item there rather than a memory.

PR #1160, opened 2026-08-13
spec: declare the drift main is already carrying

dereuromark:
`npm run engine:report -- --check` is red on `main` with seven undeclared slugs. CI runs it, so this is not latent.

## Two causes, one of them mine

**`:code` and `:mark` left the semantic registry** (markup-carve/carve-js#1027) - Carve already spells both, so the extension form renders as the generic `span.ext-NAME` fallback rather than as `<code>` / `<mark>`. The pin predates it and still emits the elements:

```
:code[*b*] :mark[*b*]
```

```html
<p><span class="ext-code"><strong>b</strong></span> <span class="ext-mark"><strong>b</strong></span></p>   <!-- corpus -->
<p><code><strong>b</strong></code> <mark><strong>b</strong></mark></p>                                       <!-- pin -->
```

Four documents. The window is normal; not declaring it is not.

**The separator category was renumbered under its own declarations.** markup-carve/carve#1157 appended `two-attributes-need-a-separator-between-them` as 299 and declared its three slugs under that number. `the-semantic-registry-holds-no-element-carve-already-spells` landed alongside it and took 299, so the separator category regenerated as **300** and the three declarations pointed at slugs that no longer exist.

That is the hazard `scripts/generate-corpus.mjs` warns about in its append-only guard - *"each engine allowlists categories by `NN-slug`"* - reached from a direction the guard does not cover. It is not a renumber inside one change; it is **two changes appending a category concurrently**, where whichever merges second is renumbered and every line naming its old number goes stale. The guard fires on the FILES. Nothing checks the declarations that reference them.

Worth a follow-up on its own: a check that every slug named in `resources/engine-pin-drift.txt` exists in `tests/corpus`. The file already fails in both directions for slugs that stop drifting; it does not fail for slugs that stop existing.

## Verification

Each of the seven was measured against the pinned build rather than assumed. `engine:report --check` now reports "Drift matches what is declared".

`npm test` is 1985 pass / 3 fail, and those three fail identically on `main` before this change - the writer round-trip pair and the `.fmt` fixture check, which are the §6c rollout's open window (markup-carve/carve#1152).

### DIFF
--- a/resources/engine-pin-drift.txt
+++ b/resources/engine-pin-drift.txt
@@ -72,3 +72,21 @@
 89-block-attribute-lines-4  A header cell now carries `scope`; the pin emits none.
 98-table-span-marker-in-first-column  A header cell now carries `scope`; the pin emits none.
 99-table-cell-attributes  A header cell now carries `scope`; the pin emits none.
+
+# `:code[…]` and `:mark[…]` left the semantic registry (carve-js#1027): Carve
+# already spells both, so the extension form renders as the generic
+# `<span class="ext-NAME">` fallback rather than as `<code>` / `<mark>`. The pin
+# predates that and still emits the elements.
+45-inline-extensions-7  `:code` and `:mark` render as the generic ext-span; the pin emits <code> and <mark>.
+45-inline-extensions-9  As 45-inline-extensions-7.
+299-the-semantic-registry-holds-no-element-carve-already-spells-2  As 45-inline-extensions-7.
+299-the-semantic-registry-holds-no-element-carve-already-spells-3  As 45-inline-extensions-7.
+
+# PART 7 requires `space+` between two attributes (carve#1157). The pin accepts
+# an adjacent pair and the corpus refuses it. These three were declared under
+# their old numbers and the category was renumbered 299 -> 300 when
+# `299-the-semantic-registry-…` landed alongside it, which left the
+# declarations pointing at slugs that no longer exist.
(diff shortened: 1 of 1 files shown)


dereuromark:
Closing this: **`main` is not red, and I measured it wrong.**

My worktree's `node_modules` held carve-js `a59df0fa` while `main`'s `package.json` pins `fb7deec`. I ran `npm ci` in that checkout during earlier work and never re-installed after the pin moved, so every engine-dependent result was about a build this branch does not pin. The seven "undeclared" slugs were the newer engine's fixes showing up as differences against an older engine.

With the correct pin installed:

```
npm run engine:report -- --check
Drift matches what is declared.

npm test
# pass 1988
# fail 0
```

So `main` is clean on both, and the three test failures I also reported as pre-existing were the same artifact.

This is trap 9 in my own notes and the second time today, which makes it a process failure rather than bad luck: **re-install after any checkout that moves `package.json`, before believing anything the pinned build says.** The tell was available and I did not look - the CI log said "reproduces now" for all seven, which is what a stale local engine looks like from the other side.

The one thing worth keeping from this: the renumber hazard is real even though this instance was not. `resources/engine-pin-drift.txt` fails in both directions for slugs that stop drifting, and not at all for slugs that stop existing - so a category renamed or renumbered under a declaration would leave a line pointing at nothing, silently. That is worth a check on its own and I will file it separately.

Issue #1164, opened 2026-08-13
The ext-NAME fallback reorders authored attributes in carve-rs


dereuromark:
## What

On the generic `ext-NAME` fallback, carve-rs emits the structural class FIRST, while carve-js keeps the author's source order and merges the base class at the first class position.

```
:widget[x]{#copy .shortcut data-key="copy"}
```

```html
<span id="copy" class="ext-widget shortcut" data-key="copy">x</span>   <!-- carve-js -->
<span class="ext-widget shortcut" id="copy" data-key="copy">x</span>   <!-- carve-rs -->
```

Measured on carve-js `main` and carve-rs `main` with the tier-split branches applied; the same shapes differ on `main` today for any `:name[…]` with an id before a class.

## Which is right

carve-js. PART 10 §1 emits authored attributes in SOURCE order, and a mandatory base class merges into the class slot at the position of the FIRST class - the rule the math span and the task-list class already follow. carve-rs writes `class="…"` unconditionally first and appends the rest afterwards (`render_attrs_after_class`), which reorders the author's attributes.

## Why it surfaced now

Nothing pinned it: before markup-carve/carve#1162 the seven semantic names never took the fallback, and no corpus case combines `:name[…]` with an id BEFORE a class. The tier split routes every `:name[…]` through the fallback, so the divergence is now one authored id away from any document.

## Work

- [ ] A corpus case: `:widget[x]{#i .c k=v}`, pinning source order with the base class merged at the first class slot.
- [ ] `markup-carve/carve-rs`: merge the base class at the first class position instead of prepending it.

Not part of markup-carve/carve#1162 - it is a pre-existing attribute-order bug in one engine, and folding an unrelated fix into a tier migration is how a migration stops being reviewable.

dereuromark:
Measured across all three engines on `main`, and the scope is wider than the ticket states: **carve-php has the same bug, and it is not limited to the `ext-NAME` fallback.**

The `ext-NAME` case from the ticket:

```
:widget[x]{#copy .shortcut data-key="copy"}
```

```html
<p><span id="copy" class="ext-widget shortcut" data-key="copy">x</span></p>   <!-- carve-js -->
<p><span class="ext-widget shortcut" id="copy" data-key="copy">x</span></p>   <!-- carve-php -->
<p><span class="ext-widget shortcut" id="copy" data-key="copy">x</span></p>   <!-- carve-rs -->
```

[...]

PR #1169, opened 2026-08-13
spec: a base class keeps the class slot in place

dereuromark:
Pins the placement rule from markup-carve/carve#1164, and fixes the executable spec, which had the same defect.

## The rule

A mandatory base class - a math span's `math inline`, the extension fallback's `ext-NAME` - is prepended INSIDE the class slot. It does not move the slot.

PART 10 §1 already says both halves:

> All classes accumulate (space-joined) into one `class` slot at its first-appearance position. ... A mandatory base class (math `math inline|display`) is prepended to that class slot.

and from the structural side:

> A structural CLASS instead merges into the class slot at its FRONT, which is the mandatory-base-class rule above said from the other side.

So an id written before any class is still serialized first:

```
:widget[x]{#i .c k=v}
```

```html
<p><span id="i" class="ext-widget c" k="v">x</span></p>
```

## Why it was never caught

No corpus case combined a base class with an id BEFORE a class. Until #1162 routed every `:name[…]` through the generic fallback, the shape was hard to reach at all, and the math span - which has carried a base class from the start - had no such case either.

The result: this oracle emitted `class` unconditionally first in BOTH places, so the executable spec agreed with the two engines that disagree with its own clause. carve-php and carve-rs produce the class-first form; carve-js produces the clause's form. Under a counting rule the wrong answer wins 2-1, which is why the clause decides it and not the engines.

## The cases

Three, appended as category `302` so nothing renumbers:

- `:widget[x]{#i .c k=v}` - the fallback, id before class.
- `` $`E=mc^2`{#i .c k=v} `` - the math span, same shape. This is the one that would have caught it years ago.
- `:widget[x]{#i k=v}` - the control: with no authored class there is no slot to keep, so the base class leads.

## The lag declaration

carve-js renders all three correctly today; carve-php and carve-rs do not. The category is declared on the comparison page as added since the quoted run, rather than the published three-engine measurement being hand-edited to claim numbers nobody measured. It clears when the two engine fixes land.


dereuromark:
Superseded: #1168 landed the `ext-NAME` half of this while the branch was open, so the remaining defect is the math span alone. Reopened as a smaller PR against the new main, with the math case and the proof it fails without the fix.

PR #1211, opened 2026-08-14
spec: a figure may wrap a table, and no Carve source spells it

dereuromark:
## What

PART 12 gains section 17: a `figure` may wrap a `table`, and no Carve source spells that shape.

## Why

`figure.target` admits a `table`, and a `table` carries its own optional `caption`. Those are two different trees for two different documents, and both render:

```html
<figure>
  <table>...</table>
  <figcaption>...</figcaption>
</figure>
```

```html
<table>
  <caption>...</caption>
  ...
</table>
```

PART 9 section 4 rules the source spelling, and it only ever produces the second. A `^ ` line after a table attaches to the table, so a parser never produces the wrapper.

The wrapper is not unused, which is why this is a clause rather than a deletion from the schema. carve-js's HTML importer produces it from a figure-wrapped table, a shape HTML itself distinguishes from a table carrying a caption element:

```html
<figure>
<table><tr><td>1</td></tr></table>
<figcaption>Cap</figcaption>
</figure>
```

`htmlToAst` on that returns `figure{target: table}`, and all three engines decode and render it identically.

What nothing said is what happens on the way back out. The canonical writer has no spelling for the wrapper, so it writes the table and its caption line, which re-reads as `table{caption}`, whose HTML puts the caption inside the table. This is the one caption shape whose source round trip changes the rendering, and until now no clause said so and no diagnostic reported it. Section 14 already declares exactly this situation for `shortCaption` and section 15 for `rowGroups`: no source spelling, the writer omits it, a bridge SHOULD report the structural loss.

Re-measured for this rebase rather than inherited from the original run. Feeding `figure{target: table, caption}` through the pinned carve-js:

```
=== HTML of the wrapper ===
<figure>
  <table>
    <thead><tr><th scope="col">A</th></tr></thead>
    <tbody><tr><td>1</td></tr></tbody>
  </table>
  <figcaption>Cap</figcaption>
</figure>
```

The canonical writer emits:

```
|=A|
| 1 |
^ Cap
```

which reparses as a `table` carrying a `caption`, rendering:

```html
<table>
  <caption>Cap</caption>
  <thead><tr><th scope="col">A</th></tr></thead>
  <tbody><tr><td>1</td></tr></tbody>
</table>
```

## What the rebase changed

**Section 16 was taken.** Composite figures claimed it for `figure_group` while this sat, so the clause is section 17 and the references in `docs/ast-json.md` and the schema test move with it. Verified free rather than assumed: nothing else in PART 12 uses 17, and the other `(section 17)` hits in the grammar are the tight/loose rule in a different PART.

**The clause predicted its own obsolescence, and the prediction was wrong.** It said the wrapper would gain a source spelling once `:::` became a captionable host. PART 9 section 4c has since made it one, and the wrapper still has no spelling. A `figure_group` holds its panels in `children` as the nodes the inner caption rules built, so a table panel is a plain `table` child of the group, never a `figure` targeting a table. A group wraps a sequence of panels and captions the sequence; this wrapper wraps one table and captions that table. That paragraph now states the distinction instead of predicting it, because it is the first question a reader hits after composite figures. `docs/ast-json.md` gets the same distinction in one sentence.

**The other direction inverted.** The original PR justified its schema test by section 4a having removed `block_quote` from `figure.target`. #1213 withdrew that, so a quote is a figure target again, and the earlier follow-up commit already pointed the clause at section 4b. The test now pins the quote branch as valid rather than as removed. It is still load-bearing, and measured so: deleting the `table` branch from `figure.target` in `resources/ast-schema.json` turns it red.

## One thing that needs your call

The clause said the writer emits "the target and its caption" without saying whose, which is only unambiguous when the table has none. A `table` carries an optional `caption`, so `figure{ target: table{ caption }, caption }` is a schema-valid tree with two captions and one `^ ` slot.

Measured on the pinned carve-js:

```
table caption present: false
  writer -> "|=A|\n^ Outer\n"
  reparse -> table

table caption present: true
  writer -> "|=A|\n^ Inner\n^ Outer\n"
  reparse -> table, paragraph
```

With the slot free the figure's caption fills it. With the slot taken the writer emits a second `^ ` line, and that line is not a caption: a `^ ` line after a caption line is an ordinary paragraph, so the output carries "Outer" as literal body prose in a `<p>`. That is worse than losing the wrapper, because the loss stops being observable as a loss.

The last commit rules it the way the clause's own logic already points: the wrapper is lost, the figure's caption is part of the wrapper, so it goes with it, and a writer MUST NOT emit a second `^ ` line. It reads to me as a consequence rather than a new decision, but it is a normative sentence that carve-js currently violates, so it is worth your nod before this merges rather than after. Happy to drop that commit and file it separately if you would rather rule on it on its own.

## Scope

The AST schema is untouched. The clause states the rule the schema already carries, and no engine behavior changes on the source path.


dereuromark:
Blocked until the section 4a revert lands. This PR's section 16 closes with a paragraph contrasting the table wrapper against a quote, and its schema test asserts that `figure.target` refuses a `block_quote` - both of which the revert reverses (WHATWG requires the attribution OUTSIDE the quote, and `figure` is already the generic captioned wrapper, so a captioned quote becomes a figure again). The clause itself, the table wrapper having no source spelling, still holds; only that closing paragraph and that one assertion change. Rebasing on the revert.

dereuromark:
The blocker is cleared. #1213 landed, so the two things you named are both handled:

- The closing paragraph no longer contrasts the wrapper against a quote. It points at PART 9 section 4b and lists a quote among the hosts that DO become a figure from source, which is the post-revert reading.
- The schema test now asserts the opposite of what it did: `figure.target` ACCEPTS a `block_quote`. It still pins the table and image branches, and still rejects a heading.

Rebased onto current main and renumbered, because section 16 was taken by `figure_group` in the meantime. Two further things the rebase surfaced, both written up in the description:

- The clause predicted it would go obsolete once `:::` became a captionable host. Section 4c has since made it one and the prediction did not come true, so that paragraph now states the distinction instead. Measured: `::: figure` around a captioned table parses to a `figure_group` whose child is a plain `table +caption`, never a `figure` targeting a table.
- A `figure` whose target is an ALREADY captioned table has two captions and one `^ ` slot. The writer currently emits a second `^ ` line, which is not a caption but an ordinary paragraph, so the outer caption lands in the output as literal body prose. The last commit rules that out. It reads as a consequence of the wrapper being lost rather than a new decision, but it is a normative sentence carve-js currently violates, so it is the one thing worth your explicit nod, and it is easy to drop into its own PR if you would rather rule on it separately.

dereuromark:
Rebased, renumbered and split.

**Split:** the fourth commit, which ruled the two-caption collision, is now `markup-carve/carve#1236`. It is a normative sentence that carve-js currently violates, so it wants your explicit nod rather than riding along with three descriptive commits. This PR is now purely descriptive: the wrapper exists, no Carve source spells it, and here is how it relates to a figure group's table panel.

**Your stated blocker is cleared.** This was drafted "blocked until the section 4a revert lands" - #1213 landed, and both things it named are handled here.

**Renumbered to PART 12 §17.** The PR originally claimed §16, which composite figures took while this sat. §17 was verified free by grep rather than assumed; the other `(§17)` hits in the grammar are the tight/loose rule in a different PART.

**The measurement was re-done, not inherited.** `figure{target: table, caption}` renders `<figure><table>…<figcaption>`; the writer emits the table plus its `^ ` line; that reparses as a `table` carrying a `caption` whose HTML puts the caption inside the table. And a `::: figure` around a captioned table parses to a `figure_group` whose child is a plain `table +caption`, never a figure targeting a table. So the wrapper still has no source spelling after composite figures.

One correction to the original body, which is now fixed in place: the clause **predicted its own obsolescence and the prediction was wrong**. It said the wrapper would gain a spelling once `:::` became a captionable host. §4c made `:::` one, and the wrapper still has none. That paragraph now states the distinction instead of predicting it away.

PR #1246, opened 2026-08-16
docs: say that a body cell can align itself

### LINKED ISSUE #1244: Support cell-level alignment in tables
### Tier

Tier-1 (core language)

### Problem or motivation

Currently tables cells can only be aligned by marking them as header cells; it would be nice to be able to specify a cell's alignment without making it a header cell.

For example, a table like:

```
| foo bar |
| baz     |
| qux     |
```

renders like this:

Image

If I want to right-align the second cell, I would currently need to do:

```
| foo bar |
| baz     |
|=> qux   |
```

But that makes it into a header cell:

Image

Ideally, cell alignment syntax should be orthogonal to cell heading syntax. In fact, since Carve already supports the `:`-style delimiters for column alignment specification, it would be more consistent to use `:` instead of `=` as a way to indicate the alignment of a cell.

### Proposed syntax / behavior

Input:

```
| foo bar |
| baz     |
|:> qux   |
```

or simply:

```
| foo bar |
| baz     |
|> qux    |
```

should produce:

Image

The options could be `:<`, `:>`, and `:~` for horizontal alignment, and perhaps even extended to `:^`, `:v` and `:-` for vertical alignment, supporting combinations like `:^>` for top-right alignment, etc. :)

### Alternatives considered

_No response_

### Additional context

_No response_

dereuromark:
Closes #1244 as a documentation fix.

Per-cell alignment already works in all three engines and is corpus-pinned (`49-table-per-cell-alignment-override`, `50-headerless-table-alignment`). The only place a reader could find it was a parenthetical on the header-cell line of the cheatsheet - "a data-cell marker overrides per cell" - with no heading and no example. #1244 asked for the feature and proposed making the cell a header cell to get the alignment, which is exactly the workaround the docs left a reader to invent.

## What changes

A short **Alignment** section on the cheatsheet: the marker set in one line (`<` left, `~` center, `>` right, glued to the pipe), one table showing the header form and a body cell overriding it, and the trap named - the marker has to be glued, because a standalone `| < |` cell is the colspan merge and `| ^ |` the rowspan merge.

The home page mentioned only the header-cell form; its sample now shows both markers and the prose matches it.

The separator-row form is deliberately left out. It is a compatibility alias rather than the canonical way to write a table, so it does not belong on the page a reader learns the syntax from.

## One thing found on the way

The new sample is written as plain Carve rather than in the annotated style the rest of the page uses, because samples in that style **do not render as the construct they document**: the trailing parentheticals turn every row into paragraph text, and `=>` / `=~` get smart-typography'd. Rendering the existing Tables sample produces paragraphs and no table.

`tests/doc-carve-samples.test.mjs` catches samples that fail to parse, not samples that parse into something else, so this has been silent. That is a page-wide problem rather than a tables one, and is filed separately as #1247.

### DIFF
--- a/docs/cheatsheet.md
+++ b/docs/cheatsheet.md
@@ -123,17 +123,24 @@ two
 | Cell    | Cell    |         rows for ROW headers)
 ^ Table caption
 
-|= Name |=> Age |=~ City |   (column alignment glued to |=: < ~ >;
-| Sum    |< 12   | NYC    |   a data-cell marker overrides per cell)
-
-| Name  | Age |              (GFM separator row accepted as a
-|-------|----:|               compatibility alias: marks the header
-| Alice |  30 |               row + column alignment)
-
 | ^      | spanned |         (^ = rowspan)
 | Header | <       |         (< = colspan)
 + continuation cell |        (+ = multi-line cell)
 ```
+
+### Alignment
+
+`<` left, `~` center, `>` right, glued to the pipe. On `|=` it aligns the whole
+column; on a plain `|` it aligns that one cell and overrides the column.
+
+```carve
+|=~ Item |=> Qty |
+| Apple | 12 |
+| Subtotal |< 12 |
+```
+
+Glued is what makes it an alignment: a standalone `| < |` cell is the colspan
+merge, `| ^ |` the rowspan merge.
 
 ## Captions (images, quotes, tables, code listings, equations, figure groups)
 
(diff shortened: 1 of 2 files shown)


waldyrious:
@dereuromark you might want to also update the docs here: https://markup-carve.github.io/carve/examples/core#tables (especially since there's already a section for table column alignment — it sounds like it could be expanded to cover table cell alignment in general.

Issue #1248, opened 2026-08-16
Table headings in list-table extension


waldyrious:
### Tier

Tier-2 (standard extension)

### Problem or motivation

It seems that the [list-table](https://markup-carve.github.io/carve/examples/extensions#listtable) extension currently doesn't support a way to mark rows (or individual cells, for that matter) as headers.

I can do:

```
::: list-table
- - A
  - B
- - cell 1A
  - cell 1B
- - cell 2A
  - cell 2B
:::
```

which produces:

<img width="177" height="137" alt="Image" src="https://github.com/user-attachments/assets/54e0239b-250c-4ecd-ba6c-acb5df62120b" />

But I don't see a way to mark the A and B cells as headings

### Proposed syntax / behavior

For entire rows, the heading property could be specified in the row-level markers:

```
::: list-table
-= - A
   - B
-  - cell 1A
   - cell 1B
-  - cell 2A
   - cell 2B
:::
```

or perhaps, using a cleaner syntax:

```
::: list-table
= - A
  - B
- - cell 1A
  - cell 1B
- - cell 2A
  - cell 2B
:::
```

For individual cells, the heading property would be specified in the column-level markers:

```
::: list-table
- -= A
  -  B
- -= cell 1A
  -  cell 1B
- -= cell 2A
  -  cell 2B
:::
```

or using the cleaner syntax:

```
::: list-table
- = A
  - B
- = cell 1A
  - cell 1B
- = cell 2A
  - cell 2B
:::
```

Open question: would there be a better way to mark an entire column as a heading in this format, than duplicating the heading markers on each cell (and thus having to ensure they remain in sync)? Perhaps something like:

```
::: list-table
- == A
  -  B
- - cell 1A
  -  cell 1B
- - cell 2A
  -  cell 2B
:::
```

(Just an example — I'm not convinced this is the best choice of syntax; for instance, `=+` could be a more intuitive alternative)

dereuromark:
Row and column headers already exist - as attributes on the preceding line rather than markers in the list, which is why they are easy to miss:

Input:

```
{header-rows=1}
{header-cols=1}
::: list-table
- - Region
  - Q1
- - EMEA
  - 10
:::
```

Output:

```html
<table>
  <thead><tr><th scope="col">Region</th><th scope="col">Q1</th></tr></thead>
  <tbody>
    <tr><th scope="row">EMEA</th><td>10</td></tr>
  </tbody>
</table>
```

[...]

PR #1252, opened 2026-08-16
docs: pin a list-table header case so the examples page shows one

dereuromark:
Addresses the docs half of #1248.

`{header-rows=N}` / `{header-cols=N}` have worked in all three engines since ListTable shipped and are documented in the extensions contract (§5.1). The [examples page](https://markup-carve.github.io/carve/examples/extensions#listtable) did not show them, because it is generated from the optional corpus and the only list-table fixture there pinned the caption. Headers were pinned inside each engine's own suite, where a reader never looks - which is why #1248 asked for the feature and proposed new markers for something two attributes already do.

## What this adds

`tests/corpus-optional/42-list-table-header-rows-cols` pins both attributes in one case:

```
{header-rows=1}
{header-cols=1}
::: list-table
- - Region
  - Q1
- - EMEA
  - 10
:::
```

```html
<table>
  <thead><tr><th scope="col">Region</th><th scope="col">Q1</th></tr></thead>
  <tbody>
    <tr><th scope="row">EMEA</th><td>10</td></tr>
  </tbody>
</table>
```

carve-js and carve-php produce that byte-for-byte (checked directly). carve-rs pins the same behavior in its own `tests/list_table.rs`, but the comparison harness has no adapter to enable the extension there, so it skips the case - visible in the run as `list-table (html): js, php`.

`npm run docs:pages` puts the example on the generated page; `docs/examples/` is gitignored, so the fixture *is* the docs change.

The case is declared as added-since-the-run rather than retaking the published snapshot in `docs/implementation-comparison.md` - the device that block already carries for the six cases before it. README optional-corpus count goes 41 → 42.

## A measurement note worth having in the record

The first optional sweep I ran reported `cross_impl_diffs=1` on `03-smart-quotes-locale-de`. That was not a divergence: the harness resolves carve-rs through `<carve-rs>/target/release/carve`, and that binary was five days old and predates the `--quote-locale` flag, so the run compared a build that could not do the thing being measured. Rebuilt at `origin/main`, the sweep is clean (`cross_impl_diffs=0`, 42 pairs, every case reaching at least two engines).

That also means the published snapshot understates the current state in two ways worth a separate look: it says one case reaches fewer than two engines, and its table attributes that to carve-js having no quote-locale option (carve#560) - the fresh run shows `smart-quotes-locale-de (html): rust, js, php`. Not touched here; it needs its own PR and its own reasoning.

### DIFF
--- a/tests/corpus-optional/manifest.json
+++ b/tests/corpus-optional/manifest.json
@@ -74,7 +74,7 @@
     {
       "slug": "15-citations-locator-labels",
       "feature": "citations-numbered",
-      "description": "Multiple locator labels (chap., §) parse to their canonical citeproc names."
+      "description": "Multiple locator labels (chap., \u00a7) parse to their canonical citeproc names."
     },
     {
       "slug": "16-citations-locator-default-page",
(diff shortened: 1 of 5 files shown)


tests/corpus-optional/manifest.json
```diff
@@ -74,7 +74,7 @@
     {
       "slug": "15-citations-locator-labels",
       "feature": "citations-numbered",
-      "description": "Multiple locator labels (chap., §) parse to their canonical citeproc names."
+      "description": "Multiple locator labels (chap., \u00a7) parse to their canonical citeproc names."
```
waldyrious: Why this change? 🤔

dereuromark: Nothing intentional, and you are right that it should not be there. Adding the case re-serialized the whole file through a writer that escapes non-ASCII, so two untouched descriptions had their section sign rewritten as a `\u` escape. Identical JSON, worse to read, and two unrelated lines in the diff of a docs change.

Reverted in 0dee00e - the new case is now the only entry the diff touches. I checked whether a script regenerates this manifest, in which case the escaping would come back: nothing does, so the two lines were the whole fix.

Issue #1092, opened 2026-08-11
0.2: ROADMAP


dereuromark:
# 0.2: ROADMAP

Each item states what it buys, so the case for spending a source break on it is on the page rather than in a thread. Detailed status, PR-by-PR, is in the comments.

---

- [x] **Blocks do not interrupt an open paragraph** (rule simplification) — markup-carve/carve#315 — **merged into `next`**, all five PRs squashed green (carve#1089, carve-js#975, carve-php#1161, carve-rs#891, carve-bench#3)

  **What it buys.** One rule replaces a per-construct taxonomy. Today every block kind has to answer "do I interrupt a paragraph?" — heading, quote, list, image, fence, definition, attribute, comment, div, and every extension — and each answer is implemented four times, in four engines, plus the oracle. After this there is one statement: `paragraph open + nonblank -> append`, `paragraph open + blank -> close`, `no paragraph open -> classify`. carve-php's 193 test failures on the change are a fair measure of how much behaviour that matrix was silently carrying.

  **Hard-wrap safety, which is a live defect class and not a hypothesis.** Under the current rules a line that merely *looks* like a marker can restructure the block above it, and that is where real defects keep being found: markup-carve/carve-rs#889 (a tab-indented line opened a fence, heading, quote or thematic break that the same column in spaces did not), markup-carve/carve-js#976 (a definition-shaped line below an item's content column changed the item's looseness). Under the 0.2 rule neither input can change structure at all, because nothing interrupts. Every author who hard-wraps prose stops being able to trip it.

  **Extensions inherit the answer.** A new block extension today must be classified into the interrupter taxonomy, in every engine, before it is correct. Afterwards it needs no entry: block position is decided by the paragraph state, not by the construct.

  **The cost, stated plainly.** It is a source break: `intro` / `# Heading` becomes one paragraph. That is the reason it belongs in 0.2 rather than a patch, and the reason the migration mode exists.

- [ ] **File inclusion / transclusion?** — markup-carve/carve#292 (normative spec already written in markup-carve/carve#291)

  **What it buys.** A book, manual or long article can be split into per-chapter files instead of one monolith, with `{{ chapter-2.crv }}`, `{{ path #section }}`, `{{ path @lines:N-M }}` and `{{ path @shift:2 }}`. That is the single most common reason people reach past a markup language for a build tool, and it is the gap most visible to anyone writing something book-sized.

[...]

PR #1256, opened 2026-08-16
Fix formatting in case-study/background.md page

waldyrious:
## What and why

The Strengths/Weaknesses/Lesson lists in the "Background page" were missing proper list syntax,
which made them be rendered with collapsed whitespace — i.e., handled as hard-wrapped prose
and thus resulting in a single continuous line rather than three lines as intended.

This commit fixes the formatting and adjust line-wrapping
to be actually useful when reading the source content
(i.e using [semantic line breaks](https://sembr.org)).

## Checklist

- [ ] Tests added or updated for the changed behavior
- [ ] If this changes parser output: conformance corpus case added/updated in `markup-carve/carve` and the spec submodule bumped
- [ ] Linter / formatter passes locally (`npm run lint` / `cargo clippy` / `vendor/bin/phpcs`)
- [x] No `@mention` tokens outside fenced code blocks in this PR description
- [x] No AI-tool signature lines in commit messages

### DIFF
--- a/docs/case-study/background.md
+++ b/docs/case-study/background.md
@@ -25,11 +25,13 @@ Before designing something new, we must understand what exists and why.
 ### 1.2 What Each Teaches Us
 
 #### Markdown
-**Strengths**: Ubiquitous, feels natural for basic text
-**Weaknesses**: Ambiguous, fragmented (CommonMark, GFM, etc.)
-**Lesson**: Simplicity wins adoption, but ambiguity creates chaos.
+
+- **Strengths**: Ubiquitous, feels natural for basic text
+- **Weaknesses**: Ambiguous, fragmented (CommonMark, GFM, etc.)
+- **Lesson**: Simplicity wins adoption, but ambiguity creates chaos.
 
 #### reStructuredText
+
 ```rst
 This is *emphasis* and **strong emphasis**.
 
@@ -38,11 +40,13 @@ This is *emphasis* and **strong emphasis**.
 
 `Link text <https://example.com>`_
 ```
-**Strengths**: Explicit directives, extensible, great for documentation
-**Weaknesses**: Verbose, underscore suffix for links is bizarre
-**Lesson**: Explicitness is good, but syntax should feel natural.
+
+- **Strengths**: Explicit directives, extensible, great for documentation
+- **Weaknesses**: Verbose, underscore suffix for links is bizarre
+- **Lesson**: Explicitness is good, but syntax should feel natural.
 
 #### AsciiDoc
+
 ```asciidoc
 = Document Title
 :author: John Doe
@@ -60,11 +64,14 @@ def hello():
     print("Hello")
 ----
 ```
-**Strengths**: Document attributes, admonitions, includes, powerful tables
-**Weaknesses**: Learning curve, multiple syntaxes for same thing
-**Lesson**: Metadata and document structure matter. Built-in admonitions are valuable.
+
+- **Strengths**: Document attributes, admonitions, includes, powerful tables
+- **Weaknesses**: Learning curve, multiple syntaxes for same thing
+- **Lesson**: Metadata and document structure matter.
+  Built-in admonitions are valuable.
 
 #### Org-mode
+
 ```org
 * Heading 1
 ** Heading 2
@@ -84,12 +91,15 @@ def hello():
     print("hello")
 #+END_SRC
 ```
-**Strengths**: Incredibly powerful, outlining, TODO states, time tracking
-**Weaknesses**: Emacs-centric, `#+` syntax is ugly, steep learning curve
-**Lesson**: Plain text can be a complete productivity system. Checkboxes and TODO
-states are genuinely useful. The `/italic/` convention works!
+
+- **Strengths**: Incredibly powerful, outlining, TODO states, time tracking
+- **Weaknesses**: Emacs-centric, `#+` syntax is ugly, steep learning curve
+- **Lesson**: Plain text can be a complete productivity system.
+  Checkboxes and TODO states are genuinely useful.
+  The `/italic/` convention works!
 
 #### Textile
+
 ```textile
 This is *strong* and _emphasis_ and -deleted- and +inserted+.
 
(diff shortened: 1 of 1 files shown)


docs/case-study/background.md
```diff
@@ -100,12 +110,14 @@ This is *strong* and _emphasis_ and -deleted- and +inserted+.
...
 ```
-**Strengths**: Intuitive emphasis, simple links, readable
-**Weaknesses**: Largely abandoned, some ambiguous cases
-**Lesson**: `"text":url` for links is genuinely more readable. The `|_.` for
-headers is clever.
+
+- **Strengths**: Intuitive emphasis, simple links, readable
+- **Weaknesses**: Largely abandoned, some ambiguous cases
+- **Lesson**: `"text":url` for links is genuinely more readable.
+  The `|_.` for headers is clever.
```
waldyrious, line 117: Side note: I don't really follow this. What's clever about that syntax? Am I missing something? 😅

docs/case-study/background.md
```diff
@@ -116,11 +128,13 @@ This is **bold** and //italic//.
 |= Header |= Header |
 | Cell    | Cell    |
 ```
-**Strengths**: Designed for wiki interoperability, clear delimiters
-**Weaknesses**: Never achieved widespread adoption
-**Lesson**: `//italic//` is visually perfect. `|=` for table headers is elegant.
+
+- **Strengths**: Designed for wiki interoperability, clear delimiters
+- **Weaknesses**: Never achieved widespread adoption
+- **Lesson**: `//italic//` is visually perfect. `|=` for table headers is elegant.
```
waldyrious, line 134: Another side-note: In WikiCreole, using `|=` for table headers is indeed elegant, because it matches the heading syntax (`== foobar ==`). For Carve, `|#` would arguably be the equivalently elegant choice.

docs/case-study/background.md
dereuromark: Fair question, and the line undersells it - what is clever is not `_`, it is the slot it sits in.

Textile gives every cell one modifier slot: everything between the `|` and a terminating `.`. Header, both axes of alignment and both spans live in that one place. Verified against python-textile:

| written | emitted |
| --- | --- |
| `\|_. Header \|` | `<th>Header</th>` |
| `\|=. c \|>. r \|<. l \|` | `text-align: center / right / left` |
| `\|^. t \|-. m \|~. b \|` | `vertical-align: top / middle / bottom` |
| `\|\2. wide \|` | `colspan="2"` |
| `\|/2. tall \|` | `rowspan="2"` |

The terminator is the whole trick: a modifier can never be mistaken for content, so `<` meaning "align left" and `<` meaning "a less-than sign" need no rule to tell them apart.

That is exactly the seam Carve pays for elsewhere. We glue markers to the pipe with no terminator, so `| < |` is the colspan merge while `|< x |` is align-left, and the two are separated by whether the marker is glued - which is a real rule people have to learn, and which markup-carve/carve#1244 is currently poking at. Textile pays a mandatory `.` on every modified cell and a small modifier language instead. Different bill, same problem.

Worth noting Textile also has vertical alignment, which Carve does not - the open half of that same issue.

For provenance: this section is design research from the first commit of the repo, written before any Carve syntax existed, and the case study is history rather than normative. So it is fine to sharpen it. If you want to fold it into this PR:

> **Lesson**: `"text":url` for links is genuinely more readable. Textile's cell modifier slot - everything between `|` and a terminating `.`, covering header, alignment and spans - is the cleverer idea: the terminator means a modifier can never be confused with content.

Happy to leave it for a follow-up too, since this PR is about formatting.

docs/case-study/background.md
dereuromark: Well, ideally headers would already be
```
= H1
== H2
```
etc ;) 
Then it would be also consistent (and why Creole uses it).

But for now  `|= ` at least somewhat relates to "highlighting".

docs/case-study/background.md
waldyrious: > Happy to leave it for a follow-up too, since this PR is about formatting.

Of course :) Thanks for the explanation.

docs/case-study/background.md
dereuromark: I mean, we could really make `=[space]Content` headers and canonical.
And keep `#` as fallback.
This way it would really be consistent.

Something for 0.2 maybe?

waldyrious: Ha, ok, that's clever :) Still, I'd leave it as a note to consider in the future (as a separate discussion, of course), whether `|#` would work. As much as "highlighting" can be connected to header rows, the "header" analogy would be the ideal match, assuming the issue is merely the current implementation rather than a fundamental technical limitation.

**Edit:** sorry, didn't see your second comment — I was already composing mine when you wrote it 😅  
Yeah, considering `==` headers for 0.2 sounds like a good plan. Maybe even freeing up `#` for comments? :grin:

dereuromark: The comments part is probably too radical, given that pasted markdown content would suddently have headers becoming comments and drop from visible output.. I think thats somewhere in the dismissed syntax for that reason - and also why I didnt go for `=` directly (or Djot interrupt behavior for that matter). The MD familiarity is important for adaptability.

waldyrious: Ah, makes sense. In that case then I'd recommend reconsidering `|#` for v0.2 :)

dereuromark: I would probably rather want headers to go into `=` direction with the # fallback in place.
But otherwise I would agree.

waldyrious: And I would back you on that, as I mentioned above, but you said yourself that that's a non-starter as it would break markdown compat, so... I'm confused: which of those two requirements (keep markdown compat vs. preferring `=` to `#` for header syntax) does take precedence? Something's gotta give 😅

**Edit:** ah, I see that's already considered as choice B in #1092. 👍

dereuromark: Undecided, so far. But yes, `|#` would be less "fallout".

PR #1292, opened 2026-08-16
spec: lazy continuation extends an open paragraph and nothing else

### LINKED ISSUE #1280: Lazy continuation after a non-paragraph block: a list item folds, a block quote does not
`resources/examples/edge-cases.md:2376` states the rule once:

> Lazy continuation only ever extends an **open paragraph**. After a block inside an item, a dedented line therefore folds in only when that block leaves a paragraph open.

A heading leaves no open paragraph, so neither shape below should fold. One of them does, in every engine and in the executable spec, which is why no gate can see it.

## The shape

```
- # H
tail
```

carve-js, carve-php and carve-rs:

```html
<ul>
  <li>
    <h1 id="H">H</h1>
    tail
  </li>
</ul>
```

The same last child in a block quote:

```
> # H
tail
```

carve-js, carve-php and carve-rs:

```html
<blockquote>
  <h1 id="H">H</h1>
</blockquote>
<p>tail</p>
```

## Where the line actually falls

Sweeping the container's last child, each followed by a column-0 `tail` line:

| last child | list item | block quote |
| --- | --- | --- |
| heading | folds | ends |
| line comment | folds | folds in js/php, ends in rs |
| comment block | folds | ends |
| table | folds | ends |
| thematic break | folds | ends |
| link reference definition | folds | ends |
| footnote definition | folds | ends |
| attribute block | folds | folds in js/php, ends in rs |
| closed code fence | **ends** | ends |
| bare div | **ends** | ends |
| nested block quote | folds | folds |

The nested-quote row is right and the clause says so, since a quote's trailing paragraph is open. The closed-fence and div rows are right. Every other list-item row is a line folding into an item that has no open paragraph to fold into, and the closed fence and the div are then unexplained exceptions to whatever rule produced the rest.

## The question

One of two things has to be written down:

- a list item absorbs a dedented line after **any** block it holds, whatever that block was, and the clause gains an explicit carve-out plus an explanation for why a closed fence and a div are excluded; or
- lazy continuation means what the clause says, an item ends exactly like a quote does, and the ten folding rows above become corpus cases pinning that.

Either answer is implementable. What cannot stay is a rule stated in one sentence that describes one container and not the other.

## No corpus case pins it

`grep` over `resources/examples` finds cases for lazy continuation into an item after a paragraph and after a nested list, and for the block quote's fenced-code stop. None places a heading, a table or a definition as an item's last child with a column-0 line after it.

---

Found by a differential sweep on 2026-08-16 against carve `8a9724d`, carve-js `a985a1ea`, carve-php `eaf39fa` and carve-rs `b345b3b5`, with all three engines rebuilt from `origin/main`. Every gate was green at those revisions: 1142/1142 corpus pairs per engine, `cross_impl_diffs=0` over five targets, 1053 documents conformant to the executable spec, 1056 unanimous AST trees, 39/39 canonical-writer claims, `combinatorial:check` 120/0 and `property:check` 800/0.

dereuromark:
Closes markup-carve/carve#1280.

`PART 1 S4`'s *NO OPEN PARAGRAPH, NO LAZY LINE* is already normative and already
works the list-item case by name. Nothing followed it. A column-0 line after a
container's last block folded into a list item and ended a block quote, in all
three engines and in this repo's own executable spec - which is why no gate
could see it.

## What the clause was missing

It is spelled with an empty quote, which reads as though EMPTINESS were the
property doing the work. It is not, any more than verbatimness is two clauses
down. The parameter S4 names is whether any container in the open stack holds an
OPEN PARAGRAPH, and a block that leaves none leaves none wherever it was
written: `- # H` writes a heading as the item's first block exactly as an
indented `# H` would.

The clause now says that (`THE MARKER LINE'S CONTENT IS THE ITEM'S FIRST BLOCK`,
normative), with the eight shapes it disposes of.

## The oracle

`scripts/spec/layout.mjs` asked the question only of the QUOTE spelling, so the
item collector recorded an open paragraph for a heading, a table, a break, a
comment, a definition and an attribute block on the marker line. `opensParagraph`
is now the predicate, seeded from the marker line's own content.

Input:

```
- # H
tail
```

Output before:

```html
<ul>
  <li>
    <h1 id="H">H</h1>
    tail
  </li>
</ul>
```

Output after:

```html
<ul>
  <li>
    <h1 id="H">H</h1>
  </li>
</ul>
<p>tail</p>
```

which is what the block-quote spelling already produced in every engine:

```
> # H
tail
```

```html
<blockquote>
  <h1 id="H">H</h1>
</blockquote>
<p>tail</p>
```

## No existing corpus document moves

Measured before any fixture was written, which was the point of doing this item
first: with the oracle corrected and the corpus otherwise untouched, all 1057
existing documents still pass. Corpus goes 1057 -> 1069, and every one of those
12 is new.

## What is deliberately NOT decided

The same shapes at the item's CONTENT column - `- a`, blank, `  # H`, `tail` -
still fold. No engine was measured there, and corpus
`75-list-nesting-and-looseness-4` pins the folding answer for a nested spelling
of it:

```
- a
  - b
    # N
lazy
```

Extending the fix that far turns that fixture red and its expected output would
have to be re-derived and re-agreed. It is named as out of scope in the clause
and stays open.

## The twelve cases

The eight shapes that move (heading, table, thematic break, line comment,
comment fence, link reference definition, footnote definition, attribute block),
the block-quote mirror, a sibling-marker control, and the two documents where a
paragraph IS open and the line still folds (`- > q` / `tail` and `- a` / `tail`).

The four controls are the four the pinned carve-js build still reproduces, which
is a useful independent check that they are controls.

## One judgment call, flagged rather than buried

`resources/engine-pin-drift.txt` gains the eight moved slugs. I was told not to
add drift declarations for engines that will fail the new cases, and this is not
that: it is the pinned `@markup-carve/carve` build this repo renders through,
`engine:report --check` fails in BOTH directions on it, and the tool's own
failure text says to add the line "in the commit that causes" the drift. Without
it `npm test` here is red and the PR cannot merge. It self-expires: the lines
have to be deleted in the commit that moves the pin past the carve-js fix.

Nothing was added to `scripts/spec/refused-allow.mjs` or
`resources/engine-fmt-drift.txt`.

`docs/implementation-comparison.md` declares the new category as post-dating its
quoted three-engine run, the same mechanism that page already uses; `docs/index.md`
takes the new corpus count.

Engines: all three change on the eight rows. No CHANGELOG entry - the maintainer
is doing one pass at the end.


dereuromark:
Codex review pass on the branch raised two P2 findings. Both reproduced; one is fixed, one is recorded as deferred.

**Fixed - a quote was tested for non-emptiness rather than for what it carries.** The predicate said any non-empty quote holds an open paragraph, which is the same shape of mistake the clause is about:

```
- > # H
tail
```

folded `tail` into the ITEM, with no paragraph open anywhere in the stack. A quote is a container like any other, so it is now asked the same question recursively and its own last block answers:

```html
<ul>
  <li>
    <blockquote>
      <h1 id="H">H</h1>
    </blockquote>
  </li>
</ul>
<p>tail</p>
```

`- > q` / `tail` still folds, `- > > q` / `tail` still folds, `- >` / `tail` still ends. A thirteenth corpus case pins it, mutation-proved both ways: mutating the expected file names it red, and reverting the recursion to the non-emptiness test names the same case red.

**Deferred - a wrapped attribute block.** `- {.a` / `  .b}` / `tail` still folds. The opener leaves no paragraph open, but the continuation line arrives at the item's CONTENT column, where the classifier reopens one on any non-empty residue - so deciding it needs the content-column half this ticket deliberately leaves open. I wrote the multi-line window for it, measured that it changed no output, and removed it again: a predicate that looks right while deciding nothing is worse than one that plainly does not reach the case. Recorded in the clause and beside the seed.

Corpus 1069 -> 1070, and still no existing document moves.

PR #1298, opened 2026-08-16
spec: a floating attribute is scoped to the container that holds it

### LINKED ISSUE #1281: A floating attribute as a container's last line: five of twelve boundary shapes diverge
PART 9 section 15 A2a says a floating attribute applies to the next block. When the container ends before that block arrives, "the next block" is ambiguous, and the three engines resolve it three ways. Five of twelve boundary shapes diverge and no corpus case pins any of them.

## The attribute escapes the container

```
> {.k}

tail
```

carve-php:

```html
<blockquote> </blockquote>
<p class="k">tail</p>
```

carve-js and carve-rs:

```html
<blockquote> </blockquote>
<p>tail</p>
```

An authored class crosses out of a block quote and over a blank line onto a document-level paragraph. Whatever the ruling is, an attribute reaching content the author did not write it next to is the worst of the three answers here.

## The attribute pulls the next block in

```
- {.k}
# H
```

carve-rs:

```html
<ul><li><h1 class="k" id="H">H</h1></li></ul>
```

carve-js and carve-php:

```html
<ul><li></li></ul>
<section id="H"><h1>H</h1></section>
```

Same split with a table in place of the heading:

```
- {.k}
| a | b |
```

carve-rs puts a `<table class="k">` inside the item; the other two leave the list empty and the table at document level.

## The container ends and the attribute is lost

```
> {.k}
tail
```

carve-js and carve-php:

```html
<blockquote><p class="k">tail</p></blockquote>
```

carve-rs:

```html
<blockquote>

</blockquote>
<p>tail</p>
```

An empty `<blockquote>` in the output is worth noting on its own.

## The definition list

```
:: t
:  {.k}
tail
```

carve-php folds `tail` into the `dd`; carve-js and carve-rs end the list and make it a document paragraph. Neither applies the class.

## The seven that agree

`- {.k}` then a column-0 line; `- {.k}` then a blank then a column-0 line; `. {.k}` then a column-0 line; `:::` / `{.k}` / `:::` then a line after the closer; `[^f]: {.k}` then a column-0 line; `> {.k}` / `>` / `> tail`; `- {.k}` then a blank then an indented line. Those seven are the ones a corpus case would most plausibly have covered, which is why the divergences are all in the other five.

## The question

Does a pending floating attribute survive its container's end?

- **No** is the simplest rule: the attribute is scoped to the container that holds it, an unconsumed one is dropped, and carve-rs's list-item behavior and carve-php's block quote behavior both change.
- **Yes** makes the attribute a document-level pending slot, and then carve-php's escape is correct and the other two change.

The definition-list row is a separate bug under either answer, since carve-php folds the line in without applying the attribute.

---

Found by a differential sweep on 2026-08-16 against carve `8a9724d`, carve-js `a985a1ea`, carve-php `eaf39fa` and carve-rs `b345b3b5`, with all three engines rebuilt from `origin/main`. Every gate was green at those revisions: 1142/1142 corpus pairs per engine, `cross_impl_diffs=0` over five targets, 1053 documents conformant to the executable spec, 1056 unanimous AST trees, 39/39 canonical-writer claims, `combinatorial:check` 120/0 and `property:check` 800/0.

dereuromark:
Closes markup-carve/carve#1281.

Ruled: a floating attribute is scoped to the container that holds it, an
unconsumed one is dropped, and the linter says so.

## The clause the ruling contradicted

§15 A4 named exactly one way to run out of following blocks:

> **A4 DROP IF DANGLING.** pending with no following block element (**end of
> document**) produces no output.

and A2a's rationale leaned on that being the only one:

> A4 already reserves discarding for the one case where there is genuinely
> nothing left - end of document.

The ruling creates a second discard site, so both had to move. A4 now names the
two ways to run out of blocks - end of document, and end of the container that
holds the attribute - and states the scoping rule that produces the second.
A2a's sentence says "the cases" and points at both.

The failure mode of the other reading is the argument against it:

```
> q
> {.k}

tail
```

```html
<blockquote><p>q</p></blockquote>
<p>tail</p>
```

An authored class crossing out of a block quote and over a blank line onto a
document-level paragraph is not a feature anyone would ask for.

## The lint rule: `unattached-block-attribute`

Dropping silently is the one thing this must not do, so the ruling's second half
needed a name. The table has 32 entries and no numeric code scheme, so the name
is the whole contract.

**Why this one.** Every id there is kebab-case and names the CONDITION as a noun
phrase - `unresolved-reference-link`, `unused-footnote-definition`,
`broken-crossref`, `duplicate-heading-id`. "Attach" is the verb §15 itself uses
for the operation ("attaches to the next VISIBLE block"), so the id is the
negation of the rule's own verb and points at the clause without borrowing a
word from it.

**What lost.** `dangling-block-attribute` takes A4's heading word, but "dangling"
reads as the end-of-document case specifically, and this rule covers both.
`unconsumed-block-attribute` is the ruling's prose word and is vaguer to a reader
who has not read §15. `unused-block-attribute` collides with what "unused" means
in `unused-footnote-definition`, which is "never referenced" - a different
relation.

## The table could not hold a rule specified ahead of the engines

`tests/lint-rule-table-claims.test.mjs` requires every documented rule to have a
trigger, and every trigger to fire in the pinned carve-js. So a new id could not
be specified before an engine shipped it - a chicken-and-egg on a page whose
whole point is that the id is agreed BEFORE two engines pick different ones.

`NOT_IN_THE_PIN_YET` is that window, with the same two directions of failure as
`resources/engine-pin-drift.txt`: a rule listed there that the pinned build DOES
emit fails, and so does one the page does not list. Mutation-proved both ways.

## Oracle

The definition body folded a flush-left line in without asking whether anything
was open, so `:  {.k}` / `tail` gave the `dd` a line the attribute then claimed.
A `dd` now asks PART 1 S4's question the way every other container does.

## Ten cases

The four container shapes the ruling moves, the definition body, the wrapped
attribute spelling, and four controls where the attribute is INSIDE its
container and attaches exactly as before - a quote, an item, a definition body,
and the top level. The controls are what keep "scoped" from reading as
"dropped".

## Codex review found a real hole, since fixed

The fold guard classified the body's last PHYSICAL line, and §15 A5 lets one
attribute block wrap. A body ending `{.k` / `#x}` has a closing brace on its last
line, read as prose, and `tail` folded in and took the attribute - the exact
shape the ruling forbids, arriving through the multi-line spelling. The question
is now asked of the trailing RUN through `tryAttrLine`, which is the function
that already knows a block may span lines. A tenth case pins it.

## Left open, as the ruling asks

What a container that ends up EMPTY renders - carve-rs's empty `<blockquote>` for
`> {.k}` / `tail`. Every case here uses a NON-EMPTY container specifically so
that shape is not pinned as a side effect, and A4 says it is out of scope.

Corpus 1085 -> 1095. No existing document moves. Two slugs are declared in
`resources/engine-pin-drift.txt`; the pinned build is already correct on the
other eight, which matches the ruling's finding that carve-js is on the right
side of all five shapes.

No CHANGELOG entry - the maintainer is doing one pass at the end.


dereuromark:
Codex review pass on the branch found one P2, which reproduced and is fixed in the second commit.

**The fold guard classified the wrong thing.** It asked `opensParagraph` about the definition body's last PHYSICAL line, and §15 A5 lets one attribute block wrap across lines. So a body ending

```
:: t
:  d
   {.k
   #x}
tail
```

has `#x}` as its last line, which reads as prose, and `tail` folded into the `<dd>` and took the attribute:

```html
<dl>
  <dt>t</dt>
  <dd>
    <p>d</p>
    <p class="k" id="x">tail</p>
  </dd>
</dl>
```

That is the shape the ruling forbids, arriving through the multi-line spelling. The question is now asked of the trailing RUN through `tryAttrLine`, which is the function that already knows a block may span lines:

```html
<dl>
  <dt>t</dt>
  <dd>d</dd>
</dl>
<p>tail</p>
```

A tenth corpus case pins it, mutation-proved both ways. The guard's other direction is unchanged: an indented `# H` after the same wrapped block is still attributed inside the `dd`, which is the control that keeps "scoped" from becoming "dropped".

Issue #1363, opened 2026-08-18
A blank-separated footnote body at a container's content column divides four ways


dereuromark:
A footnote definition at a list item's content column may carry a BLANK-SEPARATED second block. The item's reach over that definition is the contiguous indented run, so the blank ends it and the line after reopens the item's paragraph - and the shape divides four ways.

Measured 2026-08-18 against carve-js `79b3b3b`, carve-php `d4f214a`, carve-rs `6976b60f` and the executable spec at `9c4af50`.

Input:

````
- a
  [^f]: t

    more
tail

x[^f]
````

**carve-js** gives the note two blocks and folds `tail` into the item:

```html
<ul>
  <li><p>a</p>
    <p>tail</p>
  </li>
</ul>
```

**carve-php** gives the note ONE block and folds `more` into the item as well:

```html
<ul>
  <li><p>a</p>
    <p>more
tail</p>
  </li>
</ul>
```

with the note holding only `t`.

**carve-rs** gives the note two blocks and ends the item:

```html
<ul>
  <li>a</li>
</ul>
<p>tail</p>
```

**The executable spec** gives the note two blocks and folds `tail`, which is carve-js's answer.

## Why it is not decided by the content-column rule

PART 1 S4's AT A CONTAINER'S CONTENT COLUMN, A BLOCK ENDS THE PARAGRAPH IT SITS UNDER is stated over the BLOCK, not over its first line, and a footnote definition's indented continuation is part of that block. Read straight, that implies carve-rs's answer here: the whole definition is one block however many blocks its BODY holds, so nothing of the item's is open across any of it.

Reaching that answer means the container reader modelling the footnote parser's own multi-block extent, which is more than the contiguous-run reach that clause needed. It is recorded there as explicitly not reached rather than left to be discovered.

## Adjacent, and already settled

The CONTIGUOUS spelling is unanimous after `markup-carve/carve#1362`'s sibling PR:

````
- a
  [^f]: t
    more
tail
````

ends the item in carve-rs and the executable spec, and corpus `357-...-4` pins it. This ticket is only about what a BLANK inside the body does.

Filed rather than folded in: it is a live four-way divergence needing its own ruling, and it postdates the corpus freeze.

Issue #1350, opened 2026-08-17
A comment or a definition as a definition body's last line: one divergence, and one three-way split


dereuromark:
Found while auditing `opensParagraph`'s callers for `markup-carve/carve#1345`. Neither shape involves a continuation row, and both predate that work.

Measured 2026-08-18 against carve-js `0.1.3` (pin `79b3b3b`), carve-php `ef0a7d8`, carve-rs `38675a8` and the executable spec at `6af110c`.

## A comment as a definition body's last line: the oracle is alone

````
:: t
:  a
   %% c
tail
````

carve-js, carve-php and carve-rs all keep `tail` in the definition:

````html
<dl>
  <dt>t</dt>
  <dd>
    <p>a</p>
    <p>tail</p>
  </dd>
</dl>
````

The executable spec ends the definition, because §24 C3's comment does end the open paragraph and PART 1 S4 then finds nothing to fold into:

````html
<dl>
  <dt>t</dt>
  <dd>a</dd>
</dl>
<p>tail</p>
````

Three engines against one is not the argument - what needs settling is why a comment ends the run for S4 and does not end the `dd`, since a line that renders nothing making a visible difference is the shape `carve#625` already ruled against one container over.

## A definition as a definition body's last line: three answers

````
:: t
:  a
   [r]: /u
tail
````

- carve-js keeps `tail` in the `dd` as a second paragraph and consumes the definition
- carve-php and the executable spec end the `dd` and publish `tail` at the document level
- carve-rs folds both lines in as text: `a` / `[r]: /u` / `tail` in one paragraph, definition unregistered

`[^f]: x` in the same position divides the same three ways.

The list-item spelling of the second shape divides differently again, which suggests the two containers are being read by separate rules rather than by S4:

````
- a
  [r]: /u
tail
````

carve-js and the executable spec fold `tail` into the item; carve-php and carve-rs end it.

Nothing pins any of these today.

dereuromark:
part1: The executable spec is right.
part2: carve-php and carve-rs are right

dereuromark:
## Ruling, consolidated

**Part 1: the executable spec is right.** A comment as a definition body's last line ends the open paragraph, and the flush-left line does not fold.

**Part 2: carve-php and the executable spec are right.** A definition as a definition body's last line ends the `dd`; `tail` is published at document level and the definition registers.

**Part 3, the list-item spelling: carve-php and carve-rs are right.** `tail` does not fold.

````
- a
  [r]: /u
[...]

Issue #625, opened 2026-08-04
An attribute line in a list item: does it loosen, and does it still attach?


dereuromark:
An attribute line inside a list item is unpinned in two places, and carve-js and the reference now disagree on both. Neither has a corpus case, so nothing catches either one.

Measured on carve-js `3ac6a47` against `scripts/spec` at `9d94e46`.

## A. Does a flush attribute line loosen the item?

```
- a

  {.c}
```

| | result |
|---|---|
| carve-js | `<ul><li>a</li></ul>` (tight) |
| executable spec | `<ul><li><p>a</p></li></ul>` (loose) |

carve#621 settled this for the other invisible kinds - "§17 L1 asks whether the item holds a blank-line-separated second *paragraph*, and a comment or a definition is not a paragraph - it renders nothing at all". An attribute line renders nothing either, so the same argument reaches it, and carve-js extended the rule that way in carve-js#624.

The clause names "a comment or a definition" and stops there, which is why this is a question rather than a bug. If the argument carries, the clause should say so and a corpus case should pin it; if it does not, carve-js#624's attribute arm has to come back out.

The indented form is not in doubt: §15 makes an attribute line column-strict, so one column further in it is literal paragraph text, it really does render, and it loosens. Both engines agree there.

## B. Does a floating attribute attach across a blank inside the same item?

```
- a

  {.c}

  b
```

| | second paragraph |
|---|---|
| carve-js | `<p class="c">b</p>` |
| executable spec | `<p>b</p>` |

Here §15 A2a reads for carve-js: pending floats past what renders nothing and attaches to the next VISIBLE block, and `b` is that block, inside the same item. The reference dropping it looks like the defect.

Note this is the *same-item* case. The cross-item case is settled and needs no clause: carve-js and the reference agree that

```
- a

  {.c}
- b
```

puts no class on `b` - the item boundary ends the attribute's scope - and carve-php carrying it across is the outlier (carve-js#620, carve-php's side still open).

## Why both belong together

They are the same construct in the same container, and they pull in opposite directions: A says an attribute line is invisible enough not to loosen, B says it is live enough to still attach. Both can be true - "renders nothing" and "has no pending effect" are different questions - but deciding one without the other is how a rule ends up self-inconsistent, which is what carve#621 itself was fixing.

dereuromark:
A third question in the same family, from carve-js#625 - same clause, same container, and it wants deciding with A and B rather than after them.

[...]

PR #1364, opened 2026-08-18
spec: at a container's content column, a block ends the paragraph it sits under

### LINKED ISSUE #1357: A footnote definition's body continuation at an item's content column leaves the paragraph state undecided
A footnote definition at a list item's content column may carry an indented BODY continuation. The definition line ends the item's open paragraph (markup-carve/carve#1350), but the continuation line below it does not stay closed in every implementation, so the flush-left line after it divides.

Measured 2026-08-18 against carve-php `ef0a7d8`, carve-rs `38675a8` and the executable spec.

Input:

````
- a
  [^f]: t
    more
tail

x[^f]
````

carve-rs ends the item:

```html
<ul>
  <li>a</li>
</ul>
<p>tail</p>
```

carve-php and the executable spec fold:

```html
<ul>
  <li>a
tail
  </li>
</ul>
```

The note itself is identical in all three - body `t more`, resolved from the `x[^f]` below - so the divergence is only about where `tail` lands.

## Why it is a separate question from markup-carve/carve#1350

`#1350` rules the DEFINITION LINE an interrupter at the content column. `    more` is not a definition line; it is the definition's own body, consumed by the footnote parser, which permits no lazy continuation into it. So nothing in that ruling says what paragraph state the item is left in after the body.

Read straight, the answer looks like carve-rs's: the whole definition is one invisible block, so the item holds nothing open across all of it and `tail` arrives at column 0 with nothing to fold into - exactly the derivation `#1350` gives the one-line spelling. Written that way the two spellings of the same definition would stop answering differently, which is the test `#1350` itself applies.

Not folded into that PR because it is a live divergence rather than a consequence, and because carve-js has not been measured on it.

## Adjacent, unanimous, and already recorded

The nested-item spelling divides nothing today:

````
- outer
  - inner
    [r]: /u
tail
````

carve-php, carve-rs and the executable spec all fold `tail` into the OUTER item. PART 1 S4 already records that nested half as deliberately undecided, with corpus 75-list-nesting-and-looseness-4 pinning the folding answer, so it is noted here only so the two are not confused.

dereuromark:
Closes #1357

Also rules the `dd`-spelling sub-question recorded on #1350, since both are instances of one rule.

## The enumeration had to go

Read construct by construct, a container's content column answered three ways at once and no clause said why:

| at a list item's `content_column` | ends the item? |
| --- | --- |
| an attribute block `{.k}` | yes, in **every** implementation |
| a link or footnote definition | yes, in most (`#1350` part 3) |
| a comment `%% c` | **no**, in none |
| the same comment at a `dd`'s content column | **yes** (`#1350` part 1) |

Two of those splits are by construct and one is by container, and §10 I5 makes all of them interrupters.

## The property

PART 1 S4 gains it, stated over the block rather than over a list of constructs:

> §24 C3 says the content column IS the container body's column 0, so a line there is read as a BLOCK, and a block ends the paragraph above it. WHAT THE BLOCK RENDERS IS NOT A PARAMETER. None of them closes the container; the container ends because the FOLLOWING line is at DOCUMENT column 0 with no open paragraph anywhere in the stack to continue.

So `- a` / `  %% c` / `tail` ends the item, and so does the `%%%` fence spelling, which travels with its opener.

## The `#1350` objection, settled by measurement rather than by argument

That ticket recorded the sub-question as free either way: §17 L1a refuses to let an invisible construct have a visible effect (`#625`), and ending the item would give it one.

The **attribute block** answers it. It is invisible, it ends the item, and every implementation already agrees it does - so an objection that would forbid this is an objection to something nobody does. `#621` and `#625` ruled that an invisible line is not a PARAGRAPH for looseness; whether a line IS one and whether it ENDS one are different questions, which is exactly what `#1350`'s own `dd` half turned on.

## `#1357`: the rule is over the block, not over its first line

A footnote definition's indented continuation is part of that definition - the footnote parser consumes it and permits no lazy continuation into it - so

````
- a
  [^f]: t
    more
tail
````

ends the item exactly as the one-line spelling does, and two spellings of one definition stop answering differently. `carve-rs` already answered this way; `carve-js` and `carve-php` answered neither.

A **link** reference definition has no body - it is one line in Carve - so an indented line under it is ordinary item text in every reader. That is the control saying this is about the block's EXTENT and not about indentation, and it is what stops the fix from over-reaching.

## Not reached, and recorded rather than left to be found

A BLANK-SEPARATED footnote body. The item's reach is the CONTIGUOUS indented run, so a blank ends it, and

````
- a
  [^f]: t

    more
tail
````

divides four ways today. A rule over the block's extent implies `carve-rs`'s answer, but reaching it means the container reader modelling the footnote parser's own multi-block extent - a ruling of its own, filed as #1363. Unchanged by this PR (verified against `main`).

## Corpus

1231 to 1239. **357** pins six documents, **358** the two controls. Nothing existing moved: both changes were entirely unpinned, which is what the two tickets reported.

Every change is individually pinned, and the set constrains both directions:

| mutation | documents that fail |
| --- | --- |
| whole change reverted | 357-2, 357-3, 357-4 |
| the comment's column test alone | 357-2, 357-3 |
| the footnote-body branch alone | 357-4 |
| the body branch WIDENED to link definitions | 357-6 |

The last row is the one that matters most: an over-wide reading is caught by the link-definition control, not only an under-wide one. The attribute-block control, the one-line definition and both 358 documents survive all four.

## Drift

`resources/engine-pin-drift.txt` gains the four documents the pinned `carve-js` does not reproduce. The attribute-block control reproduces, which is what makes the pair discriminating.


dereuromark:
Handing over an UNPINNED choice this ruling's implementation forces, while the corpus is still open. Raised because `codex review` argued the opposite of what carve-php now does and nothing in the corpus settles it - an unpinned choice that gets quietly implemented is how an engine acquires behavior nobody ruled.

## The shape

An invisible block AT a container's content column, then a line BELOW that column but NOT flush-left:

````
- a
  [r]: /u
 b
tail
````

`#1364` settles the block's effect on the PARAGRAPH: it ends. What it does not say is whether the CONTAINER keeps collecting, and that is a separate question because a below-column line reaches no content column but §24 C3 still reads it as the item's own territory.

Two answers:

1. **The item keeps collecting.** ` b` becomes the item's second block and `tail` folds into it.
2. **The item ends.** ` b` and `tail` are top-level, because the only way in was a lazy continuation and the paragraph is closed.

## What is pinned, and what is not

Corpus 197 and 277 pin answer 1 for the COMMENT spellings - `- a` / `  %% x` / ` b` renders two paragraphs inside one item, and a comment FENCE at the column is followed by a below-column marker opening a nested list inside the item.

Nothing pins any other kind. carve-php currently answers 1 for all seven, measured:

| block at the content column | below-column line stays in the item | pinned by |
| --- | --- | --- |
| comment line | yes | corpus 197 |
| comment fence | yes | corpus 277 |
| link reference definition | yes | nothing |
| footnote definition | yes | nothing |
| abbreviation definition | yes | nothing |
| attribute block | yes | nothing |
| thematic break | yes | nothing |

## Why carve-php answers this way

PART 9 §10 I5 lists the comment and the definitions together as the invisible blocks, so treating them alike is the reading with evidence behind it, and the comment side is the half the corpus pins. That is a reading, not a measurement, which is exactly why it wants a document rather than a code comment.

**The thematic break is the weakest row.** It is not invisible - it renders `<hr>` - and it is in the same code branch only because nothing separated it. If the ruling splits anywhere, that is the most likely seam.

## What would settle it

One document per side would do it: the link-definition spelling of corpus 197, and the thematic-break spelling. Whichever way they go, all three engines get a target instead of a coincidence.

Measured on carve-php at the 1241 freeze (`0490ae5`), where the corpus is 1241 of 1241.

Issue #1343, opened 2026-08-17
Table cells have no vertical alignment


dereuromark:
### Tier

Tier-1 (core language)

### Problem or motivation

Carve has no vertical alignment. There is no cell marker, no column spelling and
no AST field, on any target.

The design for it already exists, but it lives in a comment thread on #1244,
where it cannot be found, tracked or ruled on. This issue carries it across so
it has a home. #1244 is about horizontal alignment on a non-header cell, which
already works; the vertical half of that report was the part that is genuinely
missing, and it was explicitly deferred there.

Two things to state up front, because they bound how much this is worth:

- **The workaround is decent.** `{.top}` / `{.middle}` / `{.bottom}` on a cell
  plus a stylesheet rule covers most of it today, and vertical alignment is
  usually a document-wide typographic choice rather than a per-cell one.
- **The browser default may absorb the common case.** A UA stylesheet renders a
  `<td>` middle-aligned in practice, so a column-wide `vertical-align: middle`
  can be a no-op. That should be measured against a real rendering before any
  of this reaches spec text - it is the kind of claim that circulates as
  folklore.

So this is filed as a real gap with a real design, not as a recommendation to
build it.

### Proposed syntax / behavior

Add a vertical axis with two new markers, `^` (top) and `v` (bottom), reuse `~`
for middle, and let the alignment markers form a **run terminated by a single
space**:

```
| [=] [alignment run] [attribute block] SPACE content
```

```
cell_opener  ::= "|" kind? align_run? attrs? " "
kind         ::= "="
align_run    ::= horizontal vertical? | vertical horizontal?
horizontal   ::= "<" | "~" | ">"
vertical     ::= "^" | "~" | "v"
```

The space is mandatory once an alignment marker is present. It is what ends the
run, so a run can hold both axes, and it is what keeps a letter marker (`v`)
from being read as content.

`~` alone is horizontal center, as today. Paired with a marker from the other
axis it takes the remaining axis. Doubled (`~~`) the first is horizontal, the
second vertical. The run is a **set**, not a sequence: `>^` and `^>` mean the
same cell, and `carve fmt` normalizes to horizontal-then-vertical.

Without an alignment marker nothing changes, so a cell with no run keeps working
with no space at all.

Every combination, on a body cell. Horizontal is one of none / `<` / `~` / `>`,
vertical one of none / `^` / `~` / `v`:

| opener | text-align | vertical-align |
| --- | --- | --- |
| `\| x` | column's | column's |
[...]

Issue #1335, opened 2026-08-17
Table columns have no representation in the exchange AST, but table_cell.align promises one


dereuromark:
### Tier

Tier-1 (core language) - PART 12, the interchange AST

### Problem or motivation

`table_cell.align` is documented as:

> Explicit per-cell alignment; absent means the cell inherits its column's.

There is no column object in the schema to inherit from. `table` has `rows`,
`caption`, `shortCaption`, `rowGroups` and `attrs` - nothing per column. So the
field's own description refers to a thing PART 12 does not define, and a
consumer reading "inherits its column's" has nowhere to look.

This is not only a documentation slip. Every exchange partner with a table
model has per-column data, and it has nowhere to land:

- pandoc's `ColSpec` is exactly `(Alignment, ColWidth)` per column
- HTML has `<col>` / `<colgroup>`
- LaTeX, Typst, DOCX, ODT all carry a column spec

Measured through `markup-carve/pandoc-carve` on a pandoc table with no header
row, which is legal and still carries column alignment: the alignment is
written onto **every cell** on the way out, because a cell is the only place
that can hold it. It never hoists back, so the round trip is not exact, and the
emitted document asserts per-cell alignment the author never wrote. Column
width is worse - it has no representation at all, so it is dropped in both
directions with nothing to report it against.

### Proposed syntax / behavior

An interchange-only field on the `table` node, exactly like `rowGroups` and
`shortCaption`:

```json
{
  "type": "table",
  "columns": [
    { "align": "right" },
    { "align": "center", "width": 0.25 }
  ],
  "rows": []
}
```

- `align`: `left` | `right` | `center`, absent means unset
- `width`: number, the column's fraction of the table width, absent means unset
- The array is positional and MAY be shorter than the widest row; a column with
  no entry has neither
- Carve 0.1 source has no spelling for `width`, and spells `align` only on a
  header marker (`|=> Name |`) - so the same status note `rowGroups` carries
  applies: a canonical Carve writer loses what source cannot spell, and a
  conversion API should report it
- HTML, plain and ANSI output MAY ignore it, as they ignore `rowGroups`

Resolution order for a cell: its own `align` wins, then its column's, then
none. That makes the existing `table_cell.align` description true.

### Alternatives considered

- **Leave it and fix the wording** instead - change `table_cell.align` to stop
  promising a column to inherit from. This is cheaper but keeps the data loss:
  there would still be nowhere for a `ColSpec` to land, and the smear-onto-every-cell
[...]

PR #1413, opened 2026-08-19
Specify semantic table row partitions

dereuromark:
Coordinates the new table semantics across the spec and executable corpus:

- ListTable per-cell `align` / `valign`, with cell-over-column precedence
- pipe-table `header-rows=N` and `footer-rows=N`
- explicit head cells become column headers; native `|=` remains valid in body rows
- core corpus case 376 and expanded optional ListTable case 44

Implementation dependencies:
- markup-carve/carve-js#1223
- markup-carve/carve-php#1482
- markup-carve/carve-rs#1135

Draft until the implementation PRs merge and the reference-engine pin can be advanced to merged `carve-js` main.

### DIFF
--- a/docs/ast-json.md
+++ b/docs/ast-json.md
@@ -86,14 +86,15 @@ It holds **counts, never rows**: the counts consume `rows` in order and have to
 account for every row exactly once, so the grouping can never contradict the
 table's content. `rows` stays the one sequence every consumer reads. Absent
 means the implicit structure renderers already derive - a leading run of header
-rows as the head, everything after it as one body, no foot. Core pipe-table
-parsers do not synthesize the field; ListTable-aware converters may synthesize
-the head, header-led body groups, foot and row-header columns its attributes
-spell. Partitions without those landmarks remain interchange-only. Like `shortCaption`,
+rows as the head, everything after it as one body, no foot. Pipe-table parsers
+synthesize the simple partition spelled by `header-rows` / `footer-rows`, and
+ListTable-aware converters may additionally synthesize header-led body groups
+and row-header columns. Partitions without those landmarks remain
+interchange-only. Like `shortCaption`,
 it exists so a richer table model (several `tbody` groups, a group's own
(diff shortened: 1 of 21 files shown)


dereuromark:
Reviewed this against the table tickets it neighbours. It makes sense, and the draft state is correct.

**What it closes that was genuinely open.** `rowGroups` existed in PART 12 with no producer - no Carve source could spell a table partition, so the field was interchange-only. `header-rows=N` / `footer-rows=N` gives it one. That retires an instance of the "declared but never produced" shape rather than adding a second.

**Two neighbouring tickets are now stale rather than blocked by this**, both commented on separately:

- #1335 ("no column object in the schema to inherit from") - PART 12 section 19 defines a positional `columns` array and it is in `ast-schema.json`. Premise gone.
- #1343 ("no cell marker, no column spelling, no AST field" for vertical alignment) - all three exist in spec now; this PR adds the ListTable per-cell spelling, which was the last missing one. What remains there is implementation: measured on 2026-08-19, `@markup-carve/carve` 0.1.4 renders `|=<^ A |` as `text-align: left` with the caret surviving as visible text, and no released carve-php or carve-rs records the feature.

**One thing to check before the pin advances.** The schema has no `headerRows` / `footerRows` keys, so the new attributes must feed the existing `rowGroups` rather than adding fields. That is the right call - but a reader who sees `header-rows=2` in source will look for it in the AST and not find it, so it is worth one sentence in `ast-json.md` saying where it lands.

**And a stale line inside the area this touches:** the `columns` description still reads "Carve 0.1 source has no spelling for t...", while corpus 370 (`{aligns="right,,center" valigns="top,middle,bottom" widths="25,50,25"}`) is exactly such a spelling. Predates the attribute form; cheap to fix while this is open.

The two conformance failures being the new fixtures against the old pinned carve-js is the expected shape - same situation vscode-carve hit today, where the fix was an expiring waiver keyed to the engine pin rather than holding the corpus back.

Issue #1425, opened 2026-08-19
Rust renders a marker-line link definition as visible item text after a document definition


dereuromark:
## Problem

After fixing #1424, the deterministic seed-101 differential sweep advances to a separate Rust-only definition-placement disagreement.

Minimal source:

```carve
[f]: t
* [d]: u
```

Current HTML:

```text
carve-js / carve-php:
<ul><li></li></ul>

carve-rs:
<ul><li>[d]: u</li></ul>
```

The document-level definition is collected normally. A link definition written directly after the list marker is also non-rendering metadata in JS and PHP, but Rust turns that second definition into visible paragraph text.

## Why this matters

This is a distinct class from #1424: it does not involve a nested list or lazy paragraph resumption. It appears only after the #1424 finding is removed from the fixed fuzz seed, so keeping it separate preserves a reviewable parser change and an accurate exploratory-gate trail.

Reproduction:

```sh
npm run fuzz:impls -- --seed=101 --count=200 --max-findings=16
```

The minimized finding reports one divergence and no crashes, panics, timeouts, or malformed outputs.

## Suggested investigation and implementation

1. Add executable-spec controls for a marker-line link definition with and without a preceding document-level link definition.
2. Confirm the ruling for link, footnote, and abbreviation definition spellings so the fix does not accidentally generalize to definition-shaped ordinary text.
3. In carve-rs, inspect the definition collection prepass and marker-line item parser. Check whether consuming the first definition leaves definition state or source offsets that prevent `[d]: u` from being collected as metadata.
4. Keep definition-only list items structurally present but empty in rendered HTML, matching JS and PHP.
5. Add focused Rust regressions for the minimal source and the controls.
6. Re-run the exact seed and update the weekly workflow comment to the next remaining finding, or to clean if none remains.

## Acceptance criteria

- JS, PHP, Rust, and the executable spec agree on the minimal source.
- A marker-line link definition remains non-rendering whether or not another definition precedes the list.
- The corresponding footnote and abbreviation controls retain their adjudicated behavior.
- Focused parser regressions pass, and broad Rust coverage passes in CI.
- Seed 101 no longer reports this divergence class.

dereuromark:
Fixed by markup-carve/carve-rs#1141 (engine) and #1427 (corpus pins, case 382).

[...]

PR #1426, opened 2026-08-19
Specify nested marker-line lazy resumption

### LINKED ISSUE #1424: Nested marker-line list loses the resumed lazy line after a two-step dedent in carve-rs
## Problem

The deterministic seed-101 differential sweep used by #1422 exposes a remaining Rust-only placement disagreement after the earlier parser findings are fixed.

Minimal source:

```carve
* * u
%
 :
```

Current HTML:

```text
carve-js / carve-php:
<ul><li><ul><li>u % :</li></ul></li></ul>

carve-rs:
<ul><li><ul><li>u %</li></ul> : </li></ul>
```

The executable-spec ruling still needs to be added to the focused regression, but the structural symptom is clear: the first lazy line (`%`) remains in the inner item's open paragraph, while the following one-space line resumes at the outer item instead of the same innermost open paragraph.

## Why this matters

This is not covered by the clean 304-document curated combinatorial matrix because it combines:

- a marker-line nested list (`* * u`)
- a flush lazy continuation
- a second lazy line between the outer and inner content columns

It is reproducible in the fixed weekly seed:

```sh
npm run fuzz:impls -- --seed=101 --count=200 --max-findings=16
```

## Suggested implementation

1. Add a focused spec/corpus ruling for the minimal source plus controls that vary the second line across the outer column, inner column, and flush-left column.
2. In carve-rs, inspect the marker-line nested-list `fold_lazy_run_and_resume` path. After folding the flush `%` line, collection resumes with `content_col - 1`; the next line appears to be assigned to the outer item before the innermost open paragraph is reconsidered.
3. Reuse the active nested content-column stack rather than a single outer `content_col` when deciding which paragraph receives the resumed run.
4. Add a scaling control beside `definition_marker_line_s4_is_answered_per_entry` so the resumed walk remains linear.
5. Remove this issue reference from #1422's weekly workflow comment when seed 101 returns zero again; the warning-to-clean transition should be explicit.

## Acceptance criteria

- JS, PHP, Rust, and the executable spec agree on the minimal source and its column controls.
- The case is pinned in the mandatory corpus.
- `seed=101`, `count=200` no longer reports this class.
- Existing nested-list lazy-continuation and performance tests remain green.

Found while finishing #1288 and #1422.

dereuromark:
## Summary

- adjudicate lazy resumption against the innermost open marker-line item
- pin eight mandatory corpus cases spanning `%`, `%%`, and definition-only nested items
- update the published corpus and engine comparison counts from 721/719 to 729/727
- advance the deterministic fuzz-gate attribution from resolved #1424 to the separate actionable #1425 finding

## Ruling

A lazy line continues the innermost open paragraph even when its indentation is between the outer and inner content columns. A flush comment remains inside that item; the `%%` spelling closes the paragraph, so valid following inner content starts a second paragraph.

A definition on the nested marker line leaves no paragraph open. Following content is then placed strictly by indentation: below the outer content column it is outside both lists, at the outer column it belongs to the outer item, and at the inner column it belongs to the inner item.

## Implementation status

- JavaScript already matched the ruling.
- PHP already matched the ruling.
- Rust is fixed by markup-carve/carve-rs#1140, merged as `5445be958afb40749303380bbb17d9248e085267`.
- The exact seed-101 sweep no longer reports this class. Its one remaining minimized divergence is separately documented in #1425.

## Verification

- `node --test tests/corpus.test.mjs tests/normativity.test.mjs tests/implementation-comparison-counts.test.mjs tests/test-manifest.test.mjs`
- 1,326 focused checks passed
- `npm run docs:pages`
- generated corpus fixtures agree with the executable spec and pinned JavaScript oracle

Closes #1424.


dereuromark:
CI surfaced a separate pinned carve-js writer defect on the new outer-item control. It is fully documented as markup-carve/carve-js#1228, and this PR now records the exact affected slug in the reverse-checked `engine-fmt-drift.txt` ledger. The parser/spec ruling remains clean; Node 24 and 26 full conformance, writer-ratchet, executable-spec, corpus-sync, and docs gates are now green.

Issue #1428, opened 2026-08-19
A definition on a lazy marker line inside a container still defines a reference


dereuromark:
## Problem

A definition behind a list marker on a line that folds into an open paragraph is part of that paragraph's text, and defines nothing. #1425 fixed the flush-left case in carve-rs. The same shape inside a container is unruled by the corpus, and all four readers answer it differently.

Source:

```
> r
> - [d]: u

[go][d]
```

carve-js:

```html
<blockquote><p>r
- [d]: u</p></blockquote>
<p><a href="u">go</a></p>
```

carve-php: identical to carve-js.

carve-rs:

```html
<blockquote><p>r
-</p></blockquote>
<p><a href="u">go</a></p>
```

executable spec:

```html
<blockquote><p>r
- [d]: u</p></blockquote>
<p>[go][d]</p>
```

So three engines put the definition's text on the page AND let it define a reference - a line that renders where it was written and also acts as metadata, which PART 12's "renders nothing where written" says a definition never does. carve-rs additionally deletes the text, leaving a bare `-`.

The div spelling splits the engines differently. After #1425, carve-rs agrees with the executable spec there and carve-js and carve-php do not:

```
::: n
r
- [d]: u
:::

[go][d]
```

carve-rs and the executable spec leave `[go][d]` unresolved; carve-js and carve-php resolve it to `u`.

The footnote kind is worse in carve-php, which half-collects it - the reference renders inside the paragraph text and an endnote appears for a note nobody wrote:

```
> r
> - [^f]: t

see[^f]
```

```html
carve-php:
<blockquote><p>r
- <a id="fnref1" href="#fn1" role="doc-noteref"><sup>1</sup></a>: t</p></blockquote>
```

## The ruling this needs pinned

NO OPEN PARAGRAPH, NO LAZY LINE (PART 0) plus §10 (a list does not interrupt an open paragraph) already decide it, and the executable spec is the reader that follows them: the marker line is lazy text of the paragraph the container holds open, so the definition is text and defines nothing. The container it sits in does not enter into it - `>`, `:::` and the document level are the same question.

The corpus says none of this, which is why three engines drifted. Pins wanted for: the quote spelling, the div spelling, and the controls where a paragraph is NOT open (`> - [d]: u` alone, and a blank quoted line before the marker) - those already agree everywhere and must keep agreeing.

## Engine tickets

- markup-carve/carve-rs: the deleted text
- markup-carve/carve-js: the definition still defines
[...]

PR #1429, opened 2026-08-19
spec: a lazy marker line's definition defines nothing, in any container

dereuromark:
Pins the ruling behind #1428. Engine fixes: markup-carve/carve-rs#1143 (merged path for the deleted text), markup-carve/carve-js#1229 and markup-carve/carve-php#1486.

## What the corpus did not say

NO OPEN PARAGRAPH, NO LAZY LINE holds wherever the paragraph is, and §10 does not let a list interrupt one. So a marker line under an open paragraph is lazy text of that paragraph whether the paragraph belongs to the document, a quote or a div - the definition on it renders as the text the author typed and defines nothing.

The corpus said that only for the document level. Three engines drifted into three answers on the container spellings: carve-rs deleted the definition's text, carve-php half-collected the footnote kind into the paragraph, and carve-js and carve-php both let the definition still resolve a reference.

## Five pairs

Three that the rule decides:

- the quote spelling, `> r` / `> - [d]: u`
- the div spelling, where the container adds no per-line prefix to hide behind
- the footnote kind, so one rule covers both definition markers

Two controls where no paragraph is open and the definition IS collected:

- `> - [d]: u` with nothing above it
- a blank quoted line before the marker

## The declared drift

Two pairs are listed in `resources/engine-pin-drift.txt`: the pinned carve-js build still registers the lazy definition, so it resolves a reference the corpus says nothing defines. That is the window the file exists to declare, and the lines go away with the pin that carries the carve-js fix.


dereuromark:
Note for whoever moves the carve-js pin next.

The two `resources/engine-pin-drift.txt` lines this PR adds describe the PINNED build only. Measured against a carve-js build carrying markup-carve/carve-js#1230, all 1308 corpus documents match, including these five. So when the pin moves - `npm run bump-carve-pin`, or the 6-hourly "Bump downstream corpora" run - both lines have to be deleted in that same commit, which is what the file's header requires and what `engine:report --check` enforces in both directions.

Issue #1432, opened 2026-08-19
May a definition probe run an extension's block matcher? Three engines, three policies


dereuromark:
## Problem

Three engines now answer "does this marker line fold into an open paragraph?" before they may cut a definition out of it (#1425, #1428). They disagree about what the probe is allowed to do with EXTENSION block matchers, and the disagreement is observable.

| engine | policy |
|---|---|
| carve-rs | the probe parses with the caller's options, so an extension's `match_block` runs during it - and a test asserts that as the FEATURE: "the parser answers for its own extensions for free" |
| carve-php | same shape, ported in markup-carve/carve-php#1487; a `codex review` pass flagged that this executes public extension callbacks twice per probe |
| carve-js | the opposite: when ANY block matcher is registered the guard is disabled entirely and the definition is collected, deliberately falling back to pre-#1136 behavior |

So for a document with a registered block matcher and a definition-shaped marker line under an open paragraph:

```
@@@ x
- [d]: u

[go][d]
```

carve-rs and carve-php ask the extension and get the right answer. carve-js declines to ask and collects the definition. The three do not agree, and nothing in the corpus can pin it, because a corpus case cannot register an extension.

## The questions to settle

1. **May the probe execute an extension's block matcher?** A matcher is conceptually a pure predicate, but it is public API and nothing says so normatively. If it may not, carve-rs's "for free" property goes away and both it and carve-php need a structural check instead.
2. **If it may, is running it two extra times per candidate line acceptable**, and should the contract say a matcher must be side-effect free and idempotent?
3. **If it may not, what is the fallback** - carve-js's collect-anyway, or suppress-anyway? They fail in opposite directions: collecting can activate a definition the page shows as text, suppressing can delete an author's line. The engines chose "collect" everywhere else in this family precisely because deleting is worse.

## Why it needs a ruling rather than an engine fix

Whichever way it goes, at least two engines change. Picking per-engine is how the current split happened.

## Notes

- PART 9's extension contract is the natural home for a sentence on matcher purity.
- carve-php#1487 also carries a narrower, php-only fidelity gap found in the same review: a legacy `addBlockPattern()` closure binds to the original parser, so custom patterns registered that way are invisible to the probe. That one is a plain bug and is being fixed there, not here.

dereuromark:
Ruled: **a matcher is a pure predicate, and a probe may ask it.** Landed as grammar PART 9R R1a in #1435.

Answering the three questions as posed:

[...]

PR #1435, opened 2026-08-19
spec: a matcher is a pure predicate, and a definition probe may ask it

dereuromark:
Rules #1432.

## The split

Collection runs before block structure, so an engine deciding whether a definition line folds into an open paragraph has to ask the block reader - and that runs extension matchers. carve-rs and carve-php ask and consult the extension; carve-js switches its guard off entirely whenever a matcher is registered. No corpus case can pin it, because a corpus document cannot register an extension.

## What R1a says

**The answer, not the method.** A definition line that PART 0 S4 folds into an open paragraph yields no entry in any symbol table - linkDefs, footnoteDefs or abbrDefs - in any container. An engine MAY answer by parsing the run twice and comparing the open chains, or any other way.

**A matcher is a pure predicate.** No observable side effects, same answer for the same `(lines, position, ctx)`. So a processor MAY invoke one speculatively, more than once at a position, and discard the result. That is not new behavior: core parsing already discards a matcher answer when the reported consumption is rejected. State belongs in `afterParse` / `beforeRender`.

**It fails toward collecting.** No probe, an exhausted budget, an unrecognized construct - collect, never suppress. The errors are asymmetric: collecting from a line that renders puts a visible line in a symbol table, which a reader can see and report; suppressing deletes the author line from the page.

## Consequences

- carve-rs and carve-php are conformant as they stand, including the rs test that asserts the probe consults an extension.
- carve-js is conformant too - its escape hatch collects, which is the required direction - but imprecise. That becomes a tracked follow-up rather than a divergence.
- Breaking clarification for any extension keeping state in a matcher. Called out in the contract page with where that state belongs instead.


dereuromark:
Two things found while writing the per-engine rows for this clause, both worth knowing before it merges.

**The abbreviation kind does not discriminate**, so the clause no longer claims it does. Section 7 already restricts an abbreviation definition to a direct child of the document, which decides every marker-line and in-container spelling, and at the document level the definition interrupts an open paragraph rather than folding into one. Measured across all four readers: no document separates R1a's answer from section 7's for that kind. Pushed as a second commit here, because listing a kind the clause cannot decide invites a conformance test that pins section 7 while claiming to pin this one - which is exactly what carve-php had (markup-carve/carve-php#1488 replaces it).

**carve-js has a second probe whose test took the opposite premise.** `test/a-continuation-marker-attaches-one-block.test.ts` asserts a matcher is called exactly once, justified in a comment reading "`matchBlock` is a public callback and nothing requires it to be pure. A probe that called it would number a matcher's first authored block 2, for a parse whose result is thrown away." R1a reverses the premise without reversing the assertion: that probe still does not need to call the matcher, and the count still says so - but now deliberately rather than because calling would be illegal. Comment updated in markup-carve/carve-js#1232; the assertion stands.

PR #1446, opened 2026-08-19
spec: the doubled run is the canonical arrow, in both families

### LINKED ISSUE #1442: The double-arrow family has no left or bidirectional form: `<=>` renders `≤>`
## Problem

The single-line arrow family is complete and the double-line family is not.
PART 9 §8:

```ebnf
arrow      = "->" | "<-" | "<->" | "=>" ;
comparison = "!=" | "<=" | ">=" ;
```

So `->`, `<-` and `<->` all exist, `=>` exists, and there is no `<=` arrow and
no `<=>` - because `<=` is claimed by `comparison` for ≤.

Measured on carve-lang 0.1.1:

| input | output |
| --- | --- |
| `->` | → |
| `<-` | ← |
| `<->` | ↔ |
| `=>` | ⇒ |
| `<=` | ≤ |
| **`<=>`** | **`≤>`** |
| **`<==`** | **`≤=`** |

`<=>` is the one that will be typed by accident. It is the common spelling for
"if and only if", a writer who has just learned `=>` has every reason to expect
it, and what they get is `≤>`.

This is not the `-->` case. `-->` rendering `–>` is the en-dash rule working as
specified. `<=>` is a hole in a family the spec otherwise closes.

## Why `<=` should probably stay ≤

Worth stating up front, because it decides most of the options: in technical
prose `<=` is far more often a comparison than a leftward implication. Whatever
is done here, taking ≤ away from `<=` would be the wrong trade.

### A. Leave it, document the asymmetry

The set is normative and closed; the hole is stated. A writer who wants ⇔ writes
the character or uses math.

Cheapest, and it leaves `<=>` producing `≤>`, which is the actual complaint.

### B. Add `<=>` → ⇔, and nothing else

`<=>` has no comparison reading - there is no operator spelled `<=>` in the
languages this notation comes from except as a three-way compare, which is not
what a prose document means by it. So the sequence is free in practice, and
today it produces `≤>`, which no document can be relying on.

Requires longest-match ordering: `<=>` must be tried before `<=`, exactly as
`<->` is already tried before `<-` and `---` before `--`. That precedence is
already in the spec's toolkit, so this is one production and one ordering note.

- Closes the case that motivated the ticket.
- Leaves the family still asymmetric: `⇐` remains unspellable.

### C. B, plus `<==` → ⇐

Gives the double family both directions without touching `<=`. `<==` currently
renders `≤=`, so again nothing can depend on it.

- The family is complete and symmetric under longest-match:
  `<==`, `==>`, `<=>` alongside `<-`, `->`, `<->`.
- It costs a second spelling for the same concept: `=>` and `==>` would both be
  ⇒, or `==>` stays unmapped and the two directions are spelled with different
  widths, which is worse. So B+C really means deciding whether `==>` is also ⇒.
  (Today `==>` renders `=⇒`.)

### D. Move ≤ to another spelling and give `<=` to ⇐

Complete symmetry, and a breaking change to the most common of the three
comparisons. Recorded only to be rejected.

## Recommendation

**B, with C considered separately.**

B is close to free: one production, one longest-match note, and the sequence it
claims currently produces output nobody wants. It fixes the case a writer
actually hits.

C is a real design question rather than a fix, and it should not ride along with
B. Making the double family symmetric means ruling on `==>` at the same time,
and "both `=>` and `==>` are ⇒" is a second spelling for one mark - which is
exactly what this language spends its budget avoiding elsewhere. If the answer
is that ⇐ is rare enough to live without, that is a defensible closed set, and
worth saying explicitly in §8 so the hole reads as a decision rather than an
oversight.

## While the set is open

Two smaller things, if §8 is being touched anyway:

- **Fractions.** `1/2` stays literal. `pymdownx.smartsymbols` renders ½, ¼ and
  ¾, and it is the one substitution the comparison stack has that Carve does
  not. Whether that is wanted is a separate question - fractions in running
  text are rarer than the arrows - but it is the only remaining item on that
  list.
- **Carve is ahead on the rest.** `!=`, `<=` and `>=` produce ≠, ≤ and ≥ here,
  and `pymdownx.smartsymbols` has none of the three. Worth keeping in mind when
  the comparison gets written up: the gap runs both ways and mostly in Carve's
  favor.

dereuromark:
The single-line arrow family was complete and the double-line family was not.
`->`, `<-`, `<->` and `=>` all existed; there was no `<=` arrow, because `<=` is
the comparison. So `<=>` rendered `≤>` and `<==` rendered `≤=`.

## The set

| | left | right | bidirectional |
| --- | --- | --- | --- |
| **canonical** | `<--` ← | `-->` → | `<-->` ↔ |
| **canonical** | `<==` ⇐ | `==>` ⇒ | `<=>` ⇔ |
| deprecated, still render | `<-` ← | `->` → | `<->` ↔ |
| comparison, unchanged | `<=` ≤ | `>=` ≥ | `!=` ≠ |

`<=` keeps ≤, which is what forces the left double arrow to grow a character.
Once it does, spelling the family at one width throughout is why the doubled run
is canonical everywhere rather than only where it was forced.

Matched longest-first, as `<->` already beat `<-`.

## `=>` is removed, and that is deliberate

Not a deprecation. It is the one form that was actively harmful rather than
merely non-canonical:

```
the callback x => x + 1 is short   ->  x ⇒ x + 1
PHP arrays use key => value        ->  key ⇒ value
a Rust match arm: Some(x) => x,    ->  Some(x) ⇒ x,
```

So the two families are treated differently on purpose, and each exception has a
one-line reason: **`<=` is a comparison, and `=>` is code.**

## One interaction checked rather than assumed

The Ohm grammar carried a note that `litDelim` sits after `arrow` *because* arrow
claims `=>` first - and `=` is the highlight delimiter. Removing `=>` exposes
that `=` to the emphasis machinery, where a later `x=y` could in principle close
a span.

Measured: `=x value and y=z` and `a =b and c=d` both stay literal, because the
flanking rules already refuse it. The note is updated rather than left naming a
token the production no longer claims.

## Contents

Normative clause in `resources/grammar.ebnf`, production reordered in
`resources/carve-core.ohm`, renderer map in `scripts/spec/render.mjs`, corpus 20
moved to the canonical forms, new corpus **386** pinning all nine arrows plus
`=>` literal and `<=` comparison, pinned-engine lag declared for both, counts
regenerated.

2599 tests pass.

Closes #1442


dereuromark:
Two things this PR should state rather than leave implied, both surfaced while measuring #1447:

- The arrow set carries no flanking guard, so once this lands `a -->b` converts while `a ---b` stays literal under #1445. I think that is right, since `-->x` is not a plausible flag, but right now it falls out of `arrow` preceding `dashRun` in the alternation rather than from a stated rule.
- `-->` is also the HTML comment closer, so prose containing `<!-- note -->` will render an arrow. That is the same hazard shape #1445 just ruled on for flags, so it deserves an explicit decision either way.

PR #1499, opened 2026-08-21
Three blank lines are a hard list boundary

### LINKED ISSUE #1430: 0.1.x: Set a hard list boundary to separate lists by extensive newlines
## Proposal

Ship the hard list boundary on the **0.1 line**, with the minimum at **three** blank
lines rather than two.

carve#1088 / carve#1096 specified two on `next` for 0.2. That clause has never been in a
release: `git tag --contains 16df32e4` is empty, and the drift ledger records that the
pinned JS writer predates it. So the constant is still free, and the 0.1 line is reachable.

One blank line stays the loose-item separator. Two stays the loose-item separator. A run
of **three or more** blank lines before a compatible sibling marker opens a new list:

```
- Emphasis and strong
- Highlight



- Something different
- And more here
```

```html
<ul>
<li>Emphasis and strong</li>
<li>Highlight</li>
</ul>
<ul>
<li>Something different</li>
<li>And more here</li>
</ul>
```

## Why three, on a released line

Because two fires only on noise in real documents, and three fires on nothing.

Sampled 8000 Markdown files (of 30712 found across local checkouts), counting only the
shape that would change - a list open above the run, a compatible sibling marker after
it, fenced content excluded:

| run before the sibling marker | sites | files |
| --- | --- | --- |
| two truly empty lines | 12 | 11 |
| two lines, one whitespace-only | 18 | 1 |
| **three or more lines** | **0** | **0** |

Every one of those 30 sites is decoration or generator output, not an author separating
two lists:

- 10 of the 12 two-blank sites are our own CHANGELOGs across checkouts (spec
  `CHANGELOG.md:394`, carve-rs, carve-php, vscode-carve), plus `shopware/UPGRADE-6.3.md`.
- All 18 whitespace-run sites are one generated Shopware reference file that emits a
  stray four-space line between items.

So at two, the rule would be wrong in every case it fires on measured input. At three it
does not fire at all, which is what a released line needs.

This is a floor from one machine's checkouts, weighted toward our own repos plus a few
large vendored projects, and it is Markdown rather than Carve. It is evidence that three
blank lines before a marker is absent in practice, not a population rate.

## Verified portable to 0.1.x

carve#1096 called the paragraph-interruption work of carve#315 a prerequisite. Measured,
that was sequencing rather than a functional dependency. In a worktree of `main`
(`64a7ddec`) with the hunk applied at `>= 3`:

- all three insertion points exist on `main` with byte-identical context -
  `let pendingSeparation = false` (2853), the `if (nm && nm.indent === baseIndent)` block
  under the same `CONT_MARKER` comment (3078), and `list.items.push(item)` (3615). `j` is
  initialized the same way, so `j - i` is the run length there too
- `tests/corpus.test.mjs` 1303/1303, patched and unpatched
- `tests/examples.test.mjs` 2/2
- `corpus-fmt-roundtrip` + `optional-corpus` + `roundtrip` 100 pass / 2 fail, identical
  with the patch reverted, so those two are pre-existing on `main`

And it fires, rather than being a clause no document can reach:

| shape | lists | items | loose |
| --- | --- | --- | --- |
| no blank | 1 | 2 | no |
| one blank | 1 | 2 | yes |
| two blanks | 1 | 2 | yes |
| three blanks | 2 | 2 | no |
| four blanks | 2 | 2 | no |
| three blanks then content at the item's content column | 1 | 1 | yes |
| three blanks inside a quote, markers kept | 2 | 2 | no |

The sixth row is the one that has to hold: a run followed by indented continuation must
keep continuing the item, at any count. The run denies a following sibling marker the
right to join; it closes nothing on its own.

## Open decision: which levels the boundary applies at

Measured on the same worktree, a list nested inside a list item does **not** split, at a
threshold of two or of three:

```
- outer

  - a



  - b
```

renders one nested list, byte-identical to the one-blank spelling. Top level splits, and
a list inside a blockquote splits, but nested-in-item does not. The nested list comes out
loose, so exactly one blank survived - the outer item's continuation path appears to
collapse the run before the nested list is parsed from `itemLines`.

§11 N1 as written states no level restriction ("an item marker after that run starts a
new sibling list"), so the clause and the oracle disagree. Two ways out:

1. The rule is stated for the top level and for container depth, and the clause says so
   explicitly. This matches what jgm sketched upstream ("two blanks right before an item
   at the main list level").
2. The rule applies at every level, the collapse is a defect, and the oracle plus every
   engine has to preserve the run into `itemLines`.

Needs a ruling before the clause is written, because the corpus fixture differs.

## Work items

- [ ] §11 N1: three or more, with the level restriction the ruling picks
- [ ] §17 L1 note: two blank lines still loosen; three or more separate
- [ ] `scripts/spec/layout.mjs`: the constant, currently `j - i >= 2` on `next`
- [ ] corpus fixture (`05-lists-22` on `next` is written for two)
- [ ] `tests/spec/05-lists.test`
- [ ] canonical writers preserve three blank lines between adjacent compatible sibling
      lists, and normalize a decorative run to one
- [ ] a lint rule for the ambiguous shape - a lone multi-blank run in a list whose other
      separators are all one blank
- [ ] engines, then the pin and the drift ledger entry

## Upstream

jgm/djot#408. jgm withdrew the uniformity objection there (CommonMark's rule was
motivated by indented code blocks after a list, which djot does not have) and floated
*two* blanks at the main list level. If djot lands two and Carve holds three, a document
with two blank lines parses as two lists there and one list here - a silent divergence on
ordinary input, so the measurement above is worth taking upstream while the number is
still open.

dereuromark:
Closes markup-carve/carve#1430 for the spec half.

A run of **three or more** blank lines before a compatible sibling marker ends
the list: the marker after it opens a new sibling list. One blank line stays the
loose separator, and so does two.

```
- apples



- oranges
```

```html
<ul>
  <li>apples</li>
</ul>
<ul>
  <li>oranges</li>
</ul>
```

## Three, not two, and on the 0.1 line

carve#1088 specified two on `next` for 0.2. `git tag --contains 16df32e4` is
empty, so the constant was never released and is still free.

The sampling in the issue is why it moves to three. Of 8000 Markdown files,
counting only the shape that would change - a list open above the run, a
compatible sibling marker after it - there were 30 sites at two blank lines and
**none** at three, and every one of the 30 was generator output or changelog
spacing rather than an author separating two lists. At two the rule would be
wrong in every case it fired on. At three it fires on nothing already written,
which is the bar a released line has to clear.

## Every level, and what that cost

The clause is unrestricted. A boundary that fired at the top level but not
inside a list item would make one spelling mean two things depending on where it
sits, which Design Principle 1 exists to prevent.

The oracle needed one line beyond the obvious three for it. The branch handling
a sub-list after a blank pushed a SINGLE blank line in place of whatever run it
consumed, so the recursive parse of the item's lines could never see three -
which is exactly why the clause and the oracle disagreed when the issue was
written. The issue guessed a different site; a trace of every blank push located
the real one.

## The run closes nothing on its own

It denies a following sibling marker the right to join. It is not an item
terminator: a continuation at the item's content column still continues the item
at any run length. That is §17's content-column model unchanged, and it has its
own corpus pair so it stays that way.

## Two pairs, not three

The nested shape is deliberately **not** pinned here. The canonical writer
cannot yet spell two sibling sub-lists inside a tight item - its tight-item join
writes both at the item's marker column, which is where they merge back - so a
fixture for it would pass only because the drift ledger excuses it, and that
excuse would outlive the pin bump. The engines pin the nested case in their own
suites, and the writer work is tracked separately.

## Measurements

- Corpus 1342/1342 before and after the oracle change; the list-sensitive test
  set is identical to an unpatched worktree of `main`, including the same nine
  pre-existing failures.
- The wider gate set matches baseline exactly: the same six failures on this
  branch and on a clean `origin/main` worktree.
- The pinned JS build does not reproduce the new top-level pair, so it is
  declared in `resources/engine-pin-drift.txt`.

## Sequencing

Engines first, then this - the corpus pins what the engines must already do.
Engine PRs accompany this one.


dereuromark:
Two of the three engine halves are on `main`, per the engines-first sequencing this PR asks for:

- markup-carve/carve-js#1275 merged at 15:22Z
- markup-carve/carve-php#1544 merged at 15:22Z

`carve-rs` had no PR in the stack, so the corpus pair here would have landed against two engines of three. One is in progress now and this PR is waiting on it rather than on anything of its own - every check here is green.

Two things worth folding in before this merges:

**The three-collapse is unspecified.** Measuring carve-js#1275 at `ce7c2b0f`, a run longer than three is already collapsed to exactly three by the writer - a six-blank source round-trips to a three-blank one, still two lists. Nothing in the clause says that, and nothing pins it: both corpus pairs use a three-blank source, so a writer that faithfully preserved an author's run of six would pass them unchanged. Since the canonical run is a deliberate ruling rather than an accident, it wants a sentence and a pair with a longer input, or the next writer is free to keep the run and nothing goes red.

**The nested exclusion is correctly scoped.** Confirming the reasoning in "Two pairs, not three" from the other side: three blank lines and six blank lines produce byte-identical wrong output inside a tight item, because the writer takes the `+` sentinel path and never reaches the blank-run branch at all. So that shape is not a blank-count problem and could not be fixed by any choice of constant here. Detail on markup-carve/carve#1501.

PR #1547, opened 2026-08-23
spec: the table-of-contents nav says what it is called

### LINKED ISSUE #1509: The table-of-contents nav is an unnamed landmark
carve#1468 swept the emitted HTML for a shape carrying a role and no accessible
name and found seven. All seven are now closed. Re-running the same sweep
against carve-js `877a735` turns up one the original pass missed, and it is the
same shape as items 2 and 3 of that ticket: an unnamed landmark.

## The nav has no name

With either table-of-contents extension registered:

```html
<nav class="toc">
<ul>
<li><a href="#One">One</a>
<ul>
<li><a href="#One-A">One A</a></li>
</ul>
</li>
</ul>
</nav>
```

`<nav>` is a navigation landmark unconditionally - unlike `<section>`, which
maps to `generic` until it is named. So this is an entry in every landmark list
that says only "navigation", which is the argument §16a already accepted for
`<section role="doc-endnotes">` and Extensions §1.5 accepted for an untitled
admonition.

It is worse here than for either of those, because a page can hold **more than
one** of them and they are indistinguishable. Extensions §8b.1 documents exactly
that: registering both `TocPlacement` and `TableOfContents` renders two navs,
one injected and one in place, and a document may write `::: toc` more than
once. A site nav or a breadcrumb nav in the surrounding page template makes it
three or four.

## What has to be decided

- **The name.** A fixed English default ("Table of contents"), which under
  Extensions §1.5 means a `labels` key rather than an option-only string.
- **Which extensions.** `TocPlacement` and `TableOfContents` emit byte-identical
  navs (§8b.1), so whatever is decided binds both, or the two stop being
  byte-identical.
- **The author seam.** §8b.1 already carries the author's `{#id .class}` onto
  the `<nav>`; whether an authored `aria-label` there outranks the label follows
  the §1.5 precedence and probably needs only to be stated.
- **The `::: footnotes` sibling.** That directive emits a `<div class="footnotes">`
  rather than a landmark, so it is not in scope - noted so the next sweep does
  not re-check it.

## Not in scope

The `<div class="toc">` degradation floor. That was item 7 of carve#1468,
withdrawn there and documented in §8b.1 as the floor for an unhandled directive;
it carries no role and needs no name.

## Measurement

carve-js `877a735`, `carveToHtml` with `tocPlacement()`, with
`tableOfContents()`, and with both. No configuration puts an `aria-label` on the
nav in any of them, and the extension exposes no option that would.

dereuromark:
Closes #1509

`<nav>` is a navigation landmark unconditionally - unlike `<section>`, which
maps to `generic` until it is named - so a table-of-contents nav with no
accessible name is an entry in every landmark list reading only "navigation".

Measured on the pinned build rather than taken from the report, and it is worse
than one anonymous entry. Registering both extensions renders **two navs that
are byte-identical, including their absence of a name**:

```html
<nav class="toc">
<ul>
<li><a href="#One">One</a></li>
</ul>
</nav>
<nav class="toc">
<ul>
<li><a href="#One">One</a></li>
</ul>
</nav>
```

A document may also write `::: toc` more than once, and a surrounding site
template contributes its own nav. Indistinguishable entries are the defect; a
single anonymous one is only how it starts.

## Derived or authored: authored, so it gets a key

This is the decision the ticket asks for, and §1.5 states the test. Three
buckets, and the nav lands in the third:

- **Derived from the document** - a tab's `[label]`, an admonition title, a task
  item's own text. Not this: the directive's content is empty and nothing on the
  page names the nav, so there is no string to derive from.
- **No fixed English default** - a diagram fence defaults to its own class word,
  so it stays an option. Not this either, and the near miss is worth stating
  because `toc` *is* this extension's class word: an abbreviation a reader hears
  spelled out is a worse landmark entry than the prose it abbreviates, and
  "Table of contents" is ordinary English a host's catalog already holds.
- **A fixed English default the extension does not already expose as an
  option** - which gets a `labels` key. Measured: no configuration puts an
  `aria-label` on the nav in any engine, so §1.5's "unless" does not fire.

So: **a new `labels` key `tocNav`, default `Table of contents`** - and no option,
which is the half §1.5 says must not be doubled.

### It is not the `summary` string wearing a second hat

That objection is the reason to measure rather than reason. `collapsible` turns
the standalone extension's output into `<details class="toc">` with a
`<summary>` and **no `<nav>` at all**:

```html
<details class="toc">
<summary>Table of Contents</summary>
<ul>
<li><a href="#One">One</a></li>
</ul>
</details>
```

The two shapes are mutually exclusive, one string is a landmark's accessible
name and the other is visible text in a disclosure widget, and `TocPlacement`
has no such option and always renders the nav. So `summary` stays option-only
under carve#1510 and this key does not collide with it. §1.5 now says so
directly, because the near-identical defaults otherwise read as a contradiction.

The map is also the one seam all three engines already honor, which matters
here: `markup-carve/carve-rs` has neither `collapsible` nor `summary`
(`markup-carve/carve-rs#1243`), so routing the name through an extension option
would mean inventing an option in an engine that has none.

## Both extensions, and the author still wins

`TocPlacement` and `TableOfContents` read the same key, so the navs stay
byte-identical - the contract §8b.3 names. And §8b.1 already carries the
author's `{#id .class}` onto the `<nav>`, so an authored `aria-label` is simply
kept with no label added beside it, by §1.5's existing precedence. That needed
only to be stated - it is already the behavior, verbatim from the pinned build:

```html
<nav class="toc wide" id="nav1" aria-label="Chapters">
```

The match is on the attribute name **case-insensitively**, as §16a already rules
for the shapes carve#1468 closed. Worth pinning, because an author writing
`ARIA-LABEL` currently has it echoed through in that spelling.

## No corpus case, and why

§8b.3 states it: `::: toc` is Tier-3, its embedded output carries
per-implementation block indentation, and it is **not** corpus-pinned - the
cross-impl contract is the nav fragment itself. A corpus case here would be
three declared skips, which is the silence the repo's own adapter test exists to
prevent. The pins are instead:

- `tests/every-labels-key-reaches-the-output.test.mjs` - the key is read off the
  §1.5 table, so a row with no probe (or a probe with no row) fails.
- `resources/examples-tier3.md` - the fence IS the expectation and is compared
  live against the pinned build.

## The key is declared ahead of the pin, not asserted into silence

No engine writes this string yet, so both pins would go red. `examples-tier3`
already has `AHEAD_OF_PIN` for exactly this; the labels test had no such window,
which meant the only way to land a labels ruling was to leave the key out of the
tables until an engine caught up - and a key nobody documents is a key nobody
implements.

So the same window is added there, **inverted** the way the Tier-3 ledger is: a
declared key must still be MISSING from the output, so the commit that moves the
pin deletes the entry because leaving it fails. Its own liveness case rejects an
entry naming a key no table documents or no probe reaches, which is the
carve#755 shape this repo keeps finding.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **The table-of-contents nav carries an accessible name** (carve#1509).
+  `<nav class="toc">` takes an `aria-label` from a new `tocNav` label
+  (default `Table of contents`), read by both TOC extensions so their navs stay
+  byte-identical; an authored `aria-label` outranks it. Extensions §8b.1, §1.5.
 - **Table column metadata and two-axis cell alignment** (carve#1344).
   `table.columns[]` carries a horizontal alignment, a vertical alignment and a
   fractional width, `table_cell` gains `valign`, positional `aligns`, `valigns`
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -221,6 +221,7 @@ language.
 | `indexBackref` | `Back to` | Index (§8.2) |
 | `tabsGroup` | `Tabs` | Tabs (§13.4) |
 | `codeGroup` | `Code examples` | CodeGroup (§13.4) |
+| `tocNav` | `Table of contents` | TocPlacement / TableOfContents (§8b.1) |
 
 These are the extension-written keys, not the whole map: PART 9 §16a lists the
 ten core writes, and the two tables together are every key an engine honors.
@@ -246,12 +247,21 @@ spellings where the extension already had one. Measured on the pinned build:
 `labels: { headingPermalink: 'PERM' }` changes nothing, while
 `tableOfContents({ summary: 'INHALT' })` reaches the `<summary>`.
 
+The table-of-contents NAV's name is a different string and does get a key
+(`tocNav`, §8b.1). `summary` is VISIBLE text in a `<details>` disclosure, and
+the collapsible shape renders that `<details>` *instead of* the `<nav>` - so the
+two never appear in one document and neither can stand in for the other. Their
+near-identical defaults are a coincidence of English, not one string with two
+homes.
+
 **The rule for a new one.** An extension-written string with a fixed English
 default gets a `labels` key **unless** the extension already exposes it as an
 option, and it does not get both. Decide it once, when the string is added: the
 map is for strings that would otherwise have nowhere to be set, and an option is
-already somewhere. The three keys in the table above predate this rule and each
-carries an option as well, arbitrated by the precedence order stated here.
+already somewhere. The first three keys in the table above predate this rule and
+each carries an option as well, arbitrated by the precedence order stated here.
+`tocNav` is the first admitted under the rule, and it has no option: the
+extensions that write it expose nothing that names the nav (§8b.1).
 
 **Why this is not localization.** There is no locale name and no built-in
 translation table, for the reason §16a gives: a locale table is data every
@@ -339,7 +349,7 @@ from disagree whenever a render option reaches inline rendering. Given
 and a symbols map of `ok` to `OK`, a hook rendering with defaults produces
 
 ```html
-<nav class="toc">
+<nav class="toc" aria-label="Table of contents">
 <ul>
 <li><a href="#h">:ok: h</a></li>
 </ul>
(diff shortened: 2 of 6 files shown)


dereuromark:
Engine halves filed rather than implemented, one per engine:

- `markup-carve/carve-js#1326`
- `markup-carve/carve-php#1574`
- `markup-carve/carve-rs#1249`

Each carries the same four points (the key, both extensions reading it, the
authored name winning case-insensitively, and the `<details>` shape taking
nothing), plus what is specific to that engine. Two worth surfacing here because
they came out of reading the engines rather than the spec:

- **carve-php** degrades to an *empty* nav when its per-render byte budget is
  exhausted. That element is still a landmark, so the name has to survive the
  degradation path - easy to drop, since the entry list is what the budget is
  about.
- **carve-rs** has no `collapsible` and no `summary` at all
  (`markup-carve/carve-rs#1243`), so its nav is the barest of the three and the
  name goes straight onto it. Its `css_class` is also configurable, which makes
  a test grepping for the literal `<nav class="toc"` pass for the wrong reason.

Both ahead-of-pin entries added here - `tocNav` in the labels test and
`TableOfContents` in the Tier-3 ledger - are deleted by whichever bump moves the
pin past the first engine to ship this. Both checks are inverted, so neither can
be left behind quietly.

PR #1649, opened 2026-08-24
ci: cut AST conformance wall time with parallel shards

dereuromark:
## What changed

- deterministically partition the complete formatter corpus across four matrix jobs
- run formatter/render gates concurrently with PART 12 and the auxiliary conformance gates
- retain exact per-shard population checks and both formatter/render failure modes
- cancel superseded manual dispatches without cancelling scheduled evidence
- make weekly exploratory sweeps opt-in for manual dispatches (`depth: full`)
- add 15-second AST child-process timeouts, progress heartbeats, and 20-minute job ceilings
- wait for every formatter shard before the existing issue verdict is filed

## Validation

- `npm test` — 2,953 passed, 1 skipped
- targeted shard/harness tests — 5 passed
- `actionlint .github/workflows/ast-conformance.yml`
- four-shard smoke run: 8 fixtures covered exactly once, 2 by each shard
- live workflow run: https://github.com/markup-carve/carve/actions/runs/32681945788

The live run remained red on existing engine findings, as the unsharded workflow was: PART 12, converters, and shard 2 reported current divergences. The harness, population checks, aggregation, and the other three shards completed normally.

## Measured gain

| | Before | This PR | Gain |
| --- | ---: | ---: | ---: |
| Formatter hot step | 28m04s | 8m14s slowest shard | 70.7% |
| Workflow wall time | 38m40s | 11m58s | 69.1% |

The comparison uses completed run 32674741449 as the recent baseline and branch run 32681945788. Full corpus and every target/engine remain covered.

### DIFF
--- a/.github/workflows/ast-conformance.yml
+++ b/.github/workflows/ast-conformance.yml
@@ -22,6 +22,12 @@ on:
     - cron: '0 4 * * 1-6'
     - cron: '0 4 * * 0'
   workflow_dispatch:
+    inputs:
+      depth:
+        description: Run the daily gates or include the weekly exploratory sweeps
+        type: choice
+        options: [daily, full]
+        default: daily
 
 permissions:
   contents: read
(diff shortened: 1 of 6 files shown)


dereuromark:
Post-merge audit of the sharding, since a partition is the classic way to build a
check that cannot fail. **The population checks hold** - I mutation-tested them
rather than taking the claim. One separate regression, filed as #1656.

Measured on 61e7ca40 with carve-js and carve-php built from their mains, Node
24, corpus of 1384 documents.

### The partition covers the corpus exactly once

Driving the real `selectShard` over the real corpus listing:

```
0/4 -> 346 docs
1/4 -> 346 docs
2/4 -> 346 docs
3/4 -> 346 docs
corpus total       : 1384
union of shards    : 1384
covered exactly 1x : 1384
MISSING (no shard) : 0 []
DUPLICATED         : 0 []
```

And in the branch run itself (32681945788), each shard job's own log:

```
corpus_pairs=346 shard=0/4
corpus_pairs=346 shard=1/4
corpus_pairs=346 shard=2/4
corpus_pairs=346 shard=3/4
```

### Mutation 1: drop one document from a shard's slice

Baseline `--shard=0/4 --limit=40` is `corpus_pairs=10`, exit 0. Appending
`.slice(1)` to `selectShard`:

```
CASES: compared 9 case(s), expected exactly 10
EXIT=2
```

### Mutation 2: put documents in no shard at all

`position % total === index` widened to `position % (total + 1) === index`, so
every fifth document lands in no shard. All four fail, not just one:

```
shard 0/4 EXIT=2 :: CASES: compared 8 case(s), expected exactly 10
shard 1/4 EXIT=2 :: CASES: compared 8 case(s), expected exactly 10
shard 2/4 EXIT=2 :: CASES: compared 8 case(s), expected exactly 10
shard 3/4 EXIT=2 :: CASES: compared 8 case(s), expected exactly 10
```

This is the one that mattered. `availableCaseCount()` reads the corpus
independently of the shard and `loadPairs()` applies it, so the two cannot drift
together - which is what makes the check load-bearing rather than a restatement.

### A shard that runs over nothing

```
$ compare:impls --shard=3/4 --limit=2 --fail-on-diff
CASES: compared 0 case(s) but expected at least 1. ... Choose a populated shard,
raise --limit, or drop it to compare the whole corpus.
EXIT=2
```

So an empty shard fails rather than reporting `pass=0/0`.

### Aggregation

A cancelled shard IS counted - `bad_jobs` picks up `cancelled` and `timed_out`,
so the cancel-reads-as-pass failure mode is not repeated here. A `skipped` one is
not counted; latent today since the matrix job has no `if:` or `needs:`, filed in
#1656 along with the bigger one.

The 15-second AST subprocess timeout throws and is caught into
`<engine>Findings`, and findings exit 1 - so a timeout fails the run by name
rather than being swallowed.

The verdict step's wait loop re-tests `pending` before taking the green branch,
so a wait that gives up cannot report green.

### The extra two failures were timing, not this branch

| run | ref | failing steps |
| --- | --- | --- |
| 32681945788 (02:06) | this branch | PART 12, converter, PART 11 formatter, render |
| 32683069262 (02:26) | `main` post-merge | PART 12, converter |
| 32683876950 (02:41) | `main` post-merge | PART 12, converter |

carve-php#1659 merged at 02:25, nineteen minutes after the branch run started.
All four shards are green on both post-merge `main` runs.

### Speed

Wall time on `main` post-merge: 10m54s and 12m05s, against the 38m40s baseline
the body cites. The gain reproduces.

`npm test` on 61e7ca40: 2957 passed, 1 skipped, 0 failed. `actionlint` is not
installed on this host so I did not re-run it; the workflow parses and executes,
which the three runs above demonstrate.

PR #1687, opened 2026-08-24
chore: retire the list.tight divergence, which no longer reproduces

### LINKED ISSUE #1685: The list.tight row in ast-value-divergence.txt is stale, and the gate stays red until it goes
`resources/ast-value-divergence.txt` still carries

```
list.tight  1  carve-rs alone; tracked at markup-carve/carve-rs#1358
```

added by #1682 at 16:27:00Z. carve-rs merged the fix three minutes later
(markup-carve/carve-rs#1359, `f35e09cd`), so the row is stale and the file's own
contract makes the gate red until it goes: *"a listed field no longer diverges
-> fixed, delete the line, fails until you do"*.

`AST conformance` run 32751456715, built against carve-rs main at 16:36, reports
exactly that and nothing else in its value panel:

```
THREE-WAY VALUE COMPARISON: the engines publish the same values everywhere.
  resources/ast-value-divergence.txt declares nothing that still diverges.

  FIXED      list.tight no longer diverges - delete its line
```

Shape and span panels were clean in the same run - 1397 documents unanimous,
28662 spans identical.

Deleting the one row is the whole change. Not opened as a PR from here because
this repo is being worked concurrently and the same file was edited twenty
minutes ago; whoever is in it can take the line out in passing.

## Separately, in the same run

`formatter (shard 3/4)` failed on one unrelated ANSI difference:

```
DIFF [ansi] 411-a-lone-indented-image-is-a-paragraph-and-its-html-cannot-say-so-3 (core): rust, js, php
```

That document is `>   Apollo` - a quoted lone image, which holds no list
node at all, so it is #1677's subject rather than a consequence of the carve-rs
merge. carve-rs prints the blockquote gutter before the image placeholder and
carve-php prints none. Noting it here only so the two failures in that run are
not read as one.

dereuromark:
chore: retire the list.tight divergence, which no longer reproduces

markup-carve/carve-rs#1359 made a list item whose second block is a lone or
captioned image at the item's content column loose, matching carve-js,
carve-php and the executable spec.

AST conformance run 32751456715 reports the value panel clean and names this
row:

    THREE-WAY VALUE COMPARISON: the engines publish the same values everywhere.
      FIXED      list.tight no longer diverges - delete its line

The file fails until the line is deleted, which is the two-directional half
working: a declaration that stops reproducing is a defect, not bookkeeping.

Closes #1685

### DIFF
--- a/resources/ast-value-divergence.txt
+++ b/resources/ast-value-divergence.txt
@@ -156,4 +156,3 @@
 #
 # Measured over 1397 samples at carve-js 99b28ab, carve-rs 661d4ff and carve-php
 # 38de559, each built from a fresh clone of main.
-list.tight  1  carve-rs alone; tracked at markup-carve/carve-rs#1358


dereuromark:
Superseded by #1686, which landed both halves - the ledger row and the `LAST_MEASURED` companion - three minutes before this branch pushed. Verified on `main` at `0dfcb85`: `resources/ast-value-divergence.txt` has no live rows and `LAST_MEASURED` is `new Map()`. Nothing here to add; closing rather than rebasing a duplicate.

PR #96, opened 2026-06-10
Spec: inline footnotes ^[content]

dereuromark:
Pins the carve-js inline-footnote feature (markup-carve/carve-js#107) into the conformance corpus and the normative spec.

## Changes

- **`docs/examples.md`** — new `## Inline footnotes` section, generating corpus pair `85-inline-footnotes` (interleaved numbering with a reference footnote).
- **`resources/grammar.ebnf`** — §16 gains the inline form: pandoc-style `^[content]`, a carve extension beyond djot. Inline-only content, shared document-order numbering, attrs on the noteref, no notes-in-notes. PART 8 ranks `^[` above superscript (so `^[x]^` is a note + literal `^`, `^^[x]` is suppressed, `\^[x]` is literal).
- **`docs/native-features-analysis.md`** — inline footnotes move from Deferred to implemented; notation `[^inline content]` becomes `^[content]`. Sidenotes stay deferred.
- **`docs/divergence-from-djot.md`** — record `^[content]` under "what Carve adds on top".
- **`docs/.vitepress/carve-lib`** — re-vendored from merged carve-js `main`. This also brings in the already-merged #102 (block-renderer hook) and #106 (tag leading digit); the existing corpus stays byte-green against the new lib.
- **`.gitignore`** — ignore `docs/superpowers/` working docs.

## Lockstep

carve-js #107 is merged. carve-php conformance follows in its own PR (spec submodule bump, `^[content]` implementation, category promotion).

PR #1703, opened 2026-08-24
Specify bare-marker list content columns

### LINKED ISSUE #1698: The three engines measure an attributed marker's content column in three different units
An item's content column counts the marker's ATTRIBUTE BLOCK, and the engines
count it in different units. Nobody counts CODEPOINTS, which is the unit the AST
positions settled on.

## Measured

`-{title="😀"} ` is 13 codepoints, 14 UTF-16 code units, 16 UTF-8 bytes. A
heading is written on the continuation line at each column in turn, and the
column at which it opens INSIDE the item is the column that engine read:

```
-{title="😀"} a
              # h
```

| continuation at | carve-js | carve-rs | carve-php |
| --- | --- | --- | --- |
| 13 (codepoints) | text | text | text |
| 14 (UTF-16 units) | **heading** | text | text |
| 16 (UTF-8 bytes) | text | **heading** | **heading** |

The same split with one non-ASCII character rather than an astral one:
`-{title="é"} ` is 13 codepoints and 13 UTF-16 units but 14 bytes, so carve-js
opens the heading at 13 and the other two at 14.

## Not the task question, and not new

This is NOT markup-carve/carve#1692. It reproduces on a PLAIN item with no
checkbox, on the branch that has measured the marker's width from the start, and
it predates every change made for that ruling - carve#1692 only made a task
item's column agree with the plain item's in the SAME engine, which is what put
the two spellings side by side where the unit is visible.

## What needs deciding

Which unit a content column is measured in. Three candidates and three engines,
one each:

1. **Codepoints.** What PART 12's AST positions use. No engine reads the column
   this way today, so all three move.
2. **UTF-16 code units.** carve-js's answer, and an artifact of `String.length`
   rather than a decision - the same artifact that was ruled against for AST
   positions.
3. **UTF-8 bytes.** carve-rs's and carve-php's answer, and likewise an artifact
   of measuring `as_ptr()` offsets and `strlen()`.

Only an attribute block can put a non-ASCII character inside a marker, so the
blast radius is exactly the attributed spellings - `-{...} `, `1.{...} ` and
their task forms. Every other marker is ASCII and all three units agree on it,
which is why this has never been reported.

Found by a diff review on carve-rs#1374.

### LINKED ISSUE #1701: Should a marker attribute block move a list item's content column?
An attributed list-item marker moves the item's content column by the *length of
the attribute text*. That makes a presentational edit - renaming a class - a
structural edit, and nothing in the toolchain says so.

This is the second half of #1692. That ticket asked whether carve-rs and
carve-php should match carve-js for task items; they should, and that is landing.
This ticket asks the question underneath it: **should the attribute block move
the content column at all?**

## What the spec actually says

`resources/grammar.ebnf` C3:

> content_column(item) := the visual column where the item's content starts
> (marker width + separator: `- ` -> 2, `1. ` -> 3, `10. ` -> 4).

Every example is a bare marker. C3 never mentions attributes. A8 says a
`-{…} text` line attributes the LIST ITEM, and never says the brace run counts
toward marker width.

So the current behavior is not a ruling. It is what three engines happen to do
because each measures the whole prefix.

## What the corpus pins

Nothing. Every corpus document that carries a marker-attached attribute is a
single-line item:

- `90-list-item-attributes*.crv` - eight documents, all one-liners
- `172-attribute-braces-on-a-list-item-marker-line.crv` - the only multi-block
  case uses the *separate* `{a=b .c}` attribute-line channel, not the marker one

A sweep for an attributed marker followed by an indented line returns zero
documents. The content column of an attributed item is completely unpinned.

## What the engines do

Measured just now: carve-js (released 0.1.4), carve-php (`main`), carve-rs
(`main`) agree exactly for plain items. Content column is `marker+attrs` width
plus one, and it is exact - one column too many is not "still inside", it demotes
the block.

| marker | content column |
| --- | --- |
| `-` | 2 |
| `9.` | 3 |
| `10.` | 4 |
| `-{.x1}` | 7 |
| `-{.x22}` | 8 |
| `-{.averylongclass}` | 19 |

## What djot does

Two reference points, and they disagree with us in different ways.

**Reference djot** (`jgm/djot`) does not have the extension at all - `-{.c} x` is
a paragraph with a span, which is our `docs/divergence-from-djot.md` §17.

**djot-php** does have it, citing the upstream proposal jgm/djot#262, and its
content column is the **bare marker width**. The attribute text is not counted:

Input:

```
-{.averylongclass} bla

  # a
```

djot-php output:

```html
<ul><li class="averylongclass"><p>bla</p><h1 id="a">a</h1></li></ul>
```

Same result at columns 2, 3, 6 and 7 - djot-php is also lenient about the column,
which we deliberately are not (§15 strict column-0). But on the question this
ticket asks, the only other implementation of the same extension says the
attribute block does not move the column.

## The failure mode

An author writes a correct document:

```
-{.x1} bla

       # a
```

```html
<ul><li class="x1">bla<h1 id="a">a</h1></li></ul>
```

They rename the class. One character. They touch nothing else:

```
-{.x22} bla

       # a
```

```html
<ul><li class="x22">bla</li></ul>
<p># a</p>
```

The heading has left the item. Then:

- `carve lint` reports nothing, exit 0.
- `carve fmt` **bakes it in**: the output is `-{.x22} bla\n\n\# a\n` - an escaped
  hash at column 0. Re-indenting afterwards no longer helps, because the source
  now says `\#`. Confirmed on carve-js and carve-rs.

So an attribute rename plus a format run destroys the block, permanently, with no
diagnostic at any step.

The sibling case is the same defect seen from the side. Two items whose tags
differ in length cannot share a body column:

```
-{.x1} bla

  # a

-{.x22} bla

  # b
```

```html
<ul><li class="x1">bla</li></ul>
<p># a</p>
<ul><li class="x22">bla</li></ul>
<p># b</p>
```

One list becomes two, with stray paragraphs between them.

## Is this already inherent?

Partly, and the difference matters. Ordered markers are already ragged: `9.`
needs column 3 and `10.` needs column 4. But that raggedness is one column, it is
inherent to the numbering, and renumbering a list is a mechanical operation
nobody performs by hand. Editing a class name is routine styling work, the shift
is unbounded (`{.averylongclass}` is 19), and the two spellings of the *same*
attributes - marker-attached vs. the preceding attribute line (A1/A2) - behave
differently: the attribute line does not move anything.

## Options

| option | pro | con |
| --- | --- | --- |
| **A. status quo** - content column counts the attribute block | body lines up under the first line's text; no migration | the column is a function of attribute text; rename = silent restructure; `fmt` bakes it in; siblings cannot align |
| **B. bare marker width** - attributes are not counted | column is a function of the marker alone, so attribute edits are safe; siblings align; matches djot-php; arguably what C3 already says | body does not line up under the first line's text; A-shaped documents change meaning |
| **C. accept either column** | no migration, and B's stability | two channels for one thing; changes what a marker at the bare column means (sublist vs. lazy text) |
| **D. djot leniency** - any column past the marker | no counting at all | contradicts §15 / C3 strict column-0, which is engine-wide and corpus-pinned |

## Recommendation: B

Three reasons, in order of weight.

**The column should not depend on text the author edits for presentation.** C3's
own framing is that the content column is the item body's column 0. A document's
column 0 does not move when you restyle it. Under A it does, and the toolchain
cannot tell you - lint is silent and fmt makes it permanent.

**C3 already reads that way.** "marker width + separator", with `- ` -> 2. The
attribute block is neither the marker nor the separator; A8 calls it an
attribute *on* the item, which is metadata, not content. Reading the braces into
"marker width" is an inference nobody wrote down.

**The one other implementation of this extension agrees.** djot-php implements
jgm/djot#262 and does not count the attribute text. That is not a vote - it is
that the two independent readings of the same proposal split, and the one that
does not make the column depend on the attribute is the one that survives an
edit.

### The migration is mild and mechanically findable

A document written for A does not fragment under B. Its body sits *inside* the
item, over-indented, so the block demotes to a paragraph - the content stays in
the right container and only the block's kind is lost. Carve has no
indented-code construct, so over-indentation costs nothing else. Every instance
is findable by one rule: an attributed item whose body sits at the old
full-prefix column. That rule is worth adding to `carve lint` either way.

Measured corpus breakage: zero documents.

### What would change my mind

A real document where the first line's text and the body visually lining up is
worth more than the column surviving an attribute edit. I do not have one, and
under B every item in a list shares a left edge, which is the property lists are
read by.

### Either way

`carve lint` should flag an item whose next indented block landed outside it.
That the current break is *silent* is a separate defect from which column is
right, and it is the part that let this go unnoticed.

dereuromark:
Closes #1701.
Closes #1698.

Implements Resolution B: marker-attached attributes and task checkboxes contribute zero to `content_column(item)`. Ordered marker width remains variable.

This reverses category 413 from #1697/#1700, updates the executable oracle and normative C3/A8 text, adds sibling, legacy full-prefix, Unicode, and attributed-parent sublist cases, and documents the Djot divergence.

Validation: corpus generation plus corpus, normativity, divergence, and executable-oracle tests (1,448 checks).

Engine PRs follow with the new corpus pin.


dereuromark:
Red on both `spec-corpus conformance` shards (run 32783808347). Two failures,
one trivial and one expected-but-undeclared.

```
AssertionError: docs/index.md says "pinned by 1400 corpus examples"; the corpus holds 1404
```

Four corpus documents added, count not updated. One line.

```
AssertionError: the oracle reads a different document out of the formatted source,
and resources/engine-pin-drift.txt does not excuse it
```

Expected consequence of moving the oracle ahead of the pinned engines: the oracle
now reads the bare-marker content column (#1701) while the pinned engine builds
still read the full prefix, so their `fmt` output re-parses to a different
document under the new oracle. `resources/engine-pin-drift.txt` is the declared
mechanism for exactly this - add the rows the failure names, in the format a
previous spec-ahead-of-engines change used.

One constraint on those rows: every declaration and ignore list in this repo must
be EMPTY before the release is cut, so they are temporary by construction. Worth
saying in the PR body that they come out once the three engine PRs land, so
nobody reads them as a standing waiver.

dereuromark:
The first two are fixed by `00e8072d`. The current red on run 32784964290 is a
third thing:

```
AssertionError: the import population changed; if the engine pin moved, inspect
every changed count and bump this with the pin - if the CORPUS grew instead, bump
it in the commit that adds the documents and account for each count that did not
move with it
```

That is `resources/import-roundtrip-baseline.json`. The four new corpus documents
need the ratchet retaken. Recipe, from hitting the same gate in #1697:

```
node --test tests/import-roundtrip-ratchets.check-helper.mjs
```

Read the `actual:` object out of the failure and write those numbers in.

Worth decomposing corpus growth from pin movement rather than writing the summed
numbers straight back: once they are added together the two causes are
indistinguishable, and the assertion text is asking you to account for each count
that did NOT move. Here the pin has not moved, so every changed count should be
explained by the four new documents alone - if one moved that the new documents
cannot explain, that is a real finding rather than ratchet noise.

Two other things from this branch's neighbourhood, for whoever is carrying it:

- `declarations:check` currently reports one pre-existing finding, carve-php
  `KNOWN_GAPS`. Every declaration and ignore list has to be empty before the tag,
  so that one needs clearing regardless of who owns it.
- carve-rb embeds the engine and will need a pin bump after carve-rs#1379 lands,
  the same way #96 was needed after carve-rs#1374. Its parity gate caught a stale
  pin once tonight already, so it is worth queueing rather than discovering.

PR #1707, opened 2026-08-24
Specify lenient authored block bases in list items

### LINKED ISSUE #1702: carve lint: an item's body that landed outside the item is not reported
When a block meant for a list item's body sits at the wrong column, it silently
leaves the item and nothing reports it.

Input:

```
-{.x22} bla

       # a
```

Output:

```html
<ul><li class="x22">bla</li></ul>
<p># a</p>
```

`carve lint` exits 0. `carve fmt` then writes `-{.x22} bla\n\n\# a\n` - an escaped
hash at column 0 - so re-indenting afterwards no longer recovers the heading.

Two rules are wanted, both cheap:

1. An indented block that sits BELOW a preceding item's content column and left
   the item, where the author's indentation makes the intent unambiguous. This is
   the general rule and it is the one that matters.
2. For the #1701 migration specifically: an attributed item whose body sits at
   the old full-prefix column. Mechanical, finds every instance, and can be
   dropped once the migration is done.

This is deliberately separate from #1701. Which column is right and whether the
break is silent are different defects; the silence is what let the first one go
unnoticed, and it stays worth fixing whichever column wins.

### LINKED ISSUE #1705: Allow block openers past a list item's content column
## Proposal

Inside a list item, recognize a block opener at **any indentation at or beyond
the item's content column**. Treat the opener's authored column as that block's
temporary base while parsing it. Canonical Carve output writes the block back at
the exact content column.

For a bullet item whose content column is 2, all three headings would therefore
be the same structure:

```carve
- item

  # heading
```

```carve
- item

   # heading
```

```carve
- item

        # heading
```

`carve fmt` canonicalizes the latter two to the first.

Indentation **below** the content column is unchanged. Top-level block openers
also remain column-strict. This proposal is about containment inside list items,
not a general three-space Markdown allowance.

## Why

Today one extra space changes the AST:

```carve
- item

  # heading
```

is a nested heading, while:

```carve
- item

   # heading
```

is a second paragraph containing the literal `# heading`. The source is valid,
lint is silent, and `fmt` canonicalizes the unintended paragraph rather than
recovering the apparent block. The same failure applies to quotes, fences,
containers, tables, definition lists and thematic breaks.

This is especially sharp after #1701: making item attributes layout-transparent
removes metadata-dependent columns, but an editor, pasted block or hand-aligned
document can still gain or lose one space and silently change node type.

## Existing precedent inside Carve

Sublist markers already implement the proposed containment rule. Under `- item`,
all of these are nested lists in the current reference engine:

```carve
- item

  - child
```

```carve
- item

   - child
```

```carve
- item

      - child
```

The executable layout code uses `indent >= contentCol` for that path and the
canonical writer normalizes the result. C3's prose currently says an opener
ABOVE the column folds as text, while examples and implementation comments
already describe sublists as opening "at or past" it. The proposal makes other
block openers follow the containment rule lists already use.

## Comparison

Measured for `- item`, a blank line, and an indented `# heading`:

| opener column | Carve today | CommonMark | Djot |
| ---: | --- | --- | --- |
| 0 | top-level heading | top-level heading | top-level heading |
| 1 | outside item, literal | top-level heading | nested heading |
| 2 | nested heading | nested heading | nested heading |
| 3-5 | paragraph text in item | nested heading | nested heading |
| 6+ | paragraph text in item | indented code from 6 | nested heading |

CommonMark gets its bounded window by consuming the list indentation and then
allowing up to three spaces before a block opener. Djot accepts any positive
indentation relative to the marker. The proposal is neither rule exactly: it
keeps Carve's minimum content-column boundary, has no indented-code transition,
and canonicalizes every accepted spelling to one column.

## Required semantic definition

Checking `indent >= content_column` is not sufficient. The parser needs a
temporary **authored block base**:

1. Measure the first nonblank line once, using the existing tab-stop arithmetic.
2. If its nonspace suffix is a block opener and its column is at least the item
   content column, set `block_base` to that authored column.
3. Parse that one block as though `block_base` were the item body's column 0.
4. Preserve indentation beyond `block_base` as content.
5. Keep source positions at their authored offsets and columns.
6. The canonical writer ignores `block_base` and emits at `content_column`.

Without this rebase, merely stripping the minimum content-column indent leaves
residual spaces in front of the opener and reproduces today's paragraph reading.

The rule must apply both with and without a preceding blank. Without a blank, a
recognized visible block opener interrupts the item's open paragraph exactly as
the same opener at the exact content column does today.

## Block-specific work

Each block family needs a paired exact-column / over-indented fixture and a
canonical round trip:

- headings and thematic breaks;
- block quotes, including repeated prefixes and lazy quote lines;
- fenced code, including body indentation, closing-fence alignment and an
  unterminated fence;
- raw blocks and colon fences, including nested containers and closers;
- tables, continuation rows, captions and row alignment;
- definition-list term openers, their `:  ` definition line and wrapped terms;
- block attribute lines followed by each supported target;
- images/captionable blocks where column determines block position;
- invisible definitions and comments, preserving their existing registration
  and any-column exceptions;
- tabs that reach or cross the minimum content column;
- the same cases inside ordered, task, attributed and nested list items.

For multi-line constructs, the first opener establishes `block_base`. A closer
or structural continuation should be interpreted relative to that base, not
independently rebased line by line. Otherwise a fence could close on accidental
dedentation or table rows at different columns could be combined unpredictably.

### Intentional literal examples change meaning

This is the main compatibility blocker. Documents can currently use
over-indentation to show block-looking text literally:

```carve
- Explain this syntax:

   # not a real heading
```

Under the proposal it becomes a heading. The current corpus deliberately pins
at least one such case:
`149-fence-folds-as-lazy-inline-code-above-the-content-column`. The existing
"over-indented definition folds" documentation must also be split between an
over-indented *opener* (changed here) and indentation internal to an already
open definition list (not automatically changed).

The compatibility spelling is an escaped opener, for example `\#`, `\>`, or an
escaped fence delimiter. Migration cannot infer whether an author intended a
real block or a literal example.

### Tight and loose list rendering can change

An over-indented opener currently becomes a paragraph. After a blank this can
make the list loose. Reclassifying it as a nested non-paragraph block can keep
the list tight under L2. Tests must assert both AST structure and `<p>` wrapping,
not only visible text.

### Definitions can start registering

An over-indented link, footnote, abbreviation or definition-list opener that is
literal today may become structural. That affects references elsewhere in the
document and can turn a local indentation edit into nonlocal output changes.
These cases require explicit migration diagnostics.

### Opaque payload indentation must survive

For a fence opened at column 5 under a content column of 2, three columns are
structural over-indent. Payload indentation beyond column 5 is authored data and
must survive. Applying one blanket trim to every collected line would corrupt
code, raw content and source positions.

### Attribute and caption attachment can move

An over-indented attribute line and target must be treated as one block group or
left literal together. Re-basing either line independently can attach attributes
or captions to a different node than the author wrote.

### Parser performance must remain linear

Do not probe every possible dedent suffix. The leading run is already measured;
classify the one nonspace suffix, store one `block_base`, and carry it through
the existing collector. Deep nesting must not re-scan the same prefix once per
candidate column. The layout-work and adversarial nesting tests need a ratchet.

### Source layout metadata needs a distinction

`contentColumn` remains the item's canonical/minimum column. If editor tooling
needs to preserve authored layout, the block's actual base is different data.
Do not overload `contentColumn`; either derive the authored base from positions
or add an explicitly optional layout field in a later schema revision.

## Migration and diagnostics

Before changing the parser, add a compatibility lint that finds a block-shaped
line above an item's content column and reports both possible intentions:

- dedent to the content column to make it structural now; or
- escape the opener to preserve literal text under the new rule.

`fmt --migrate` may safely offer those as explicit fixes, but a default formatter
must not guess. For old source parsed under a compatibility mode, preserving old
semantics means escaping the opener; opting into the new structure means
dedenting it.

The release notes should call this a source-compatibility change even if the
existing corpus delta is small. A corpus measures known syntax, not private
documents that intentionally use indentation for examples.

## Implementation areas

- **Spec and oracle:** revise PART 9 §24 C3's AT/ABOVE split; make the existing
  sublist `>=` behavior explicit; introduce and carry `block_base` through the
  one-block collectors and fence trackers.
- **carve-js:** the list collector currently admits `col >= contentCol` but
  dedents only by `contentCol`; opener recognition and block collection need the
  authored-base rebase.
- **carve-rs:** update the corresponding `indent > content_col` lazy-text branch,
  collectors, fence state and source-position tests.
- **carve-php:** update the list content-column gates, prepass fence/comment
  trackers and every duplicated content-column scan; prior fixes show these
  paths can disagree if only the main block parser moves.
- **Writers/importers:** writers should already emit the minimum column, but
  round-trip and every direct-source importer must prove that rather than assume
  it.
- **Lint/LSP:** compatibility diagnostic, dedent quick fix, escape quick fix and
  tests showing the edits preserve the selected intent.
- **Docs:** parsing ambiguities, divergence from Djot, core/edge examples,
  migration guidance and AST source-layout documentation.

## Required experiment before ruling

Implement the rule first in the executable oracle behind a temporary switch and
run the full corpus without updating expected output. Report:

1. every changed document, grouped by block family;
2. AST changes separately from HTML-only tight/loose changes;
3. formatter fixed-point and source-round-trip changes;
4. reference/footnote/abbreviation registrations that become nonlocal changes;
5. work-counter and wall-clock deltas on the deep-nesting fixtures;
6. three-engine behavior for every changed shape before moving their pins.

The known `149` fence case proves the delta is not zero. The experiment decides
whether that is the only intentional literal dependency or merely the first one
we know by name.

## Acceptance criteria

- Every block opener at `column >= content_column` nests with the same AST and
  HTML as the exact-column spelling.
- Canonical Carve output uses exactly `content_column`.
- Indentation below the content column is unchanged.
- Plain paragraph indentation is unchanged; only recognized block groups gain
  the rebase.
- Opaque payload and relative interior indentation are byte-preserved.
- Closers and continuation lines are bound to the opener's `block_base`.
- Tight/loose behavior is explicitly pinned.
- Old literal intent has a diagnostic and an escape-based migration fix.
- Source positions continue to describe authored source, not canonical output.
- All three engines, the executable oracle and direct-source importers agree.
- Layout work remains linear under adversarial nesting.

## Decision requested

Approve the experiment and the minimum rule:

> Within a list item, a recognized block opener at any indentation at or beyond
> `content_column(item)` belongs to the item. Its authored indentation is the
> temporary base of that block; canonical output emits it at
> `content_column(item)`.

Do not yet rule on making top-level block openers indentation-tolerant or on
changing below-column lazy continuation. Those are separate axes with a much
larger compatibility surface.

dereuromark:
Resolves #1702 and #1705 together across the executable oracle and migration contract.

- accepts structural list-item blocks at any authored column at or past the item content column
- keeps below-minimum lines detached/literal and preserves descendant-list ownership
- uses the opener column as the base for multiline quotes, fences, tables, definition lists, attributes, images, and captions
- keeps exact-column and over-column lone-image looseness identical
- specifies list-item-body-detached and list-item-block-overindented migration diagnostics, including dedent-versus-escape guidance
- adds edge cases for absolute visual columns, tabs, nested ownership, fence payload, captions, wrapped definitions, formatter stability, and image looseness
- pins the merged carve-js implementation commit used by executable conformance

The JS, PHP, and Rust implementation PRs have merged. Claude CLI reviewed the implementation and spec changes; CI is the final gate here.

Closes #1702.
Closes #1705.


dereuromark:
@claude please review this PR, including correctness, behavioral consistency with the linked spec issue, regressions, performance, and edge cases.

PR #1736, opened 2026-08-25
test: declare the list-column clauses as running ahead of the pinned engine

dereuromark:
`main` is red on two checks. Both have the same cause and neither is a defect in the clauses themselves: the lenient list content-column rules landed in the spec and in the oracle while the pinned build still predates them, and the resulting gaps were not declared.

Reproduced on a clean checkout of `d81bd4e0` before touching anything, so this is not fallout from another branch.

## `every trigger provokes the rule it names`

carve#1707 added `list-item-body-detached` and `list-item-block-overindented` to `docs/validation.md` AND to the `TRIGGERS` map in the same commit. Neither rule exists in the pinned build - grepping the installed engine finds no such rule id - so the triggers provoke nothing.

`TRIGGERS` is an assertion that a rule fires. A rule specified ahead of the engines belongs in `NOT_IN_THE_PIN_YET`, which is what that map's own guidance says:

> If the rule is specified ahead of the engines, declare it in NOT_IN_THE_PIN_YET with the ticket that specifies it.

Both moved there. The declaration is guarded from both directions - a companion test fails if the rule stops being documented, and fails again once the pinned build starts carrying it - so this cannot quietly outlive the engine work.

## `the oracle reads the engine's formatted output as the same document`

Thirteen corpus documents, every one a list content-column shape. The oracle now reads the lenient column; the pinned writer still emits the older indentation; so the oracle reads a different document out of the formatted source than out of the original.

That is the second of the two kinds `resources/engine-fmt-drift.txt` exists to hold, named in its own header:

> the ORACLE can fail to read the pin's output back while the pin reads it fine - which is what a spec rule the pin has not shipped looks like from the writer side. Both belong here

All thirteen are declared with the clauses that moved. The first render still matches the corpus for each, which is why only the cross-read check sees them.

## After this

3022 tests, zero failures. Every line added here is deleted when the engine pin advances past the clauses - both files say so in their headers, and both have a check that forces it.

### DIFF
--- a/resources/engine-fmt-drift.txt
+++ b/resources/engine-fmt-drift.txt
@@ -15,3 +15,16 @@
 #
 # Format: <slug><space><space><reason>
 415-a-floating-attribute-does-not-widen-a-list-item-s-content-column-7  pinned writer moves an empty task marker across its detached paragraph
+05-lists-11  oracle reads the lenient list content column of carve#1705 / carve#1707 / carve#1732; the pinned writer still emits the pre-clause indentation
+143-post-blank-list-continuation-content-column-model-3  oracle reads the lenient list content column of carve#1705 / carve#1707 / carve#1732; the pinned writer still emits the pre-clause indentation
+148-colon-fence-as-a-block-opener-in-a-list-item-3  oracle reads the lenient list content column of carve#1705 / carve#1707 / carve#1732; the pinned writer still emits the pre-clause indentation
+149-fence-folds-as-lazy-inline-code-above-the-content-column  oracle reads the lenient list content column of carve#1705 / carve#1707 / carve#1732; the pinned writer still emits the pre-clause indentation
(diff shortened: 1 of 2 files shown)


dereuromark:
Closing this: the premise is wrong, and main was never broken.

Both failures came from a stale `node_modules` in my working copy, installed at the start of a long session and never refreshed while I rebased onto newer mains. The lockfile's engine pin advanced in between; my install did not. So I measured the current spec against an older engine and read the mismatch as a defect in the clauses.

A fresh `npm ci` on an unmodified branch passes both tests. The two lint rules ARE in the pinned build - `dist/lint.js` carries them - so the `TRIGGERS` entries carve#1707 added are correct, and the thirteen fmt-drift declarations here describe drift that does not exist.

The check I thought was independent was not. I verified "on clean main" in a fresh worktree, but copied the same stale `node_modules` into it rather than installing, so the second measurement inherited the fault of the first.

Nothing here should land. carve#1707 needs no follow-up.

PR #1789, opened 2026-08-26
spec: the continuation marker is one operation in every container

### LINKED ISSUE #1782: The continuation marker is one operation, and two containers implement a second one
## The ask

Describe `+` with one operation:

> `+` transfers ownership of the next flush-left block to the current container.

After the transfer, parse that block normally. No separate rules per attached-block kind, and no branches on what the attached block is.

## Where it is not one operation today

**Two extent functions for the same marker.**

- `scripts/spec/layout.mjs:3024` `takeOneBlock` -- used by list items and block quotes. Boundary set is blank / further `+` / **a quote line**, and the result is then narrowed by `firstBlockEnd`, which asks the block parser for exactly ONE block.
- `scripts/spec/layout.mjs:3033` `takePulledBlockEnd` -- used by footnote bodies and definition descriptions. Boundary set is blank / further `+`, and there is **no narrowing at all**: whatever the extent covers is attached.

So §17 L3's ONE IS A COUNT AND THE BOUNDARY IS NOT (#1290) holds in one pair of containers and not in the other. The clause's own example, moved into a footnote body:

```
- a
+
para
> q
```

```html
<ul>
  <li>a
    para
  </li>
</ul>
<blockquote><p>q</p></blockquote>
```

```
[^n]: a
+
para
> q

see[^n]
```

```html
    <li id="fn1">
      <p>a</p>
      <p>para</p>
      <blockquote><p>q</p></blockquote>
      ...
```

The footnote took both blocks. A definition description does the same. §17 L3 says the quote stays outside, and says it with that document.

**A kind branch inside the shared path.** `takeOneBlock`'s boundary set tests `QUOTE`, so a `+` written inside a block quote never attaches a following quote line -- it falls back to ordinary quote continuation:

```
> a
+
> q
```

renders `<blockquote><p>a</p><p>q</p></blockquote>` today, and `<blockquote><p>a</p><blockquote><p>q</p></blockquote></blockquote>` with that one test removed. Under pure ownership the second is the answer: `> q` is a flush-left block and the marker says to attach the next one.

**The spec contradicts itself about which containers take the marker.** §17 L4 lists LIST ITEM and BLOCK QUOTE. §17 L5 puts footnote bodies in NOT CONTINUATION CONTAINERS -- "a `+` attach form is possible future work, deliberately OUT OF SCOPE". Meanwhile:

- `resources/grammar.ebnf:3428` -- `footnote_definition = ..., {footnote_continuation | continuation_marker_block}`,
- `resources/grammar.ebnf:3432` -- "A lone `+` also attaches a flush-left block ... the same continuation marker lists, block quotes and definition bodies use",
- `resources/grammar.ebnf:1576` -- `definition_body` carries `continuation_marker_block` too,
- corpus `66-footnote-with-multiple-blocks-2` pins the footnote `+` and has for a long time,
- both readers implement it.

L5's sentence is the only thing in the language that says otherwise, and everything else already moved.

## Measured cost of unifying

Two probes on the oracle: narrow the footnote/description pull through `firstBlockEnd` like the item path, and drop the `QUOTE` test from the boundary set.

| probe | corpus documents changed (of 1450) | tests changed (of 3112) |
| --- | --- | --- |
| pull narrowed to ONE block | 0 | 0 |
| quote kind-branch removed | 0 | 0 |

Both unifications are invisible to everything the repo currently pins. That is the finding, not a reassurance: the divergence between the two paths is unpinned, so either reading can drift without a red test.

(Measured against `origin/main` at `869d4624`, which is itself red on 5 unrelated definition-separator pin-lag tests; the counts above are deltas against that same baseline.)

## Work

1. One clause for the marker: ownership transfer of the NEXT flush-left block, one block, boundary as §17 L3 already states it. Delete the per-container repetition.
2. Fix §17 L4/L5: footnote bodies and definition descriptions TAKE the marker. Remove the out-of-scope sentence.
3. One extent function. Every container narrows to one block.
4. Decide the `> a` / `+` / `> q` case explicitly (nest, under ownership) and pin it.
5. New corpus rows for what none of the above currently pins: `+` in a footnote body and in a description with a second block after it, and the quote-after-`+`-inside-a-quote case.
6. Re-measure the engines afterwards -- neither divergence has an engine row today.

dereuromark:
Closes #1782.

## Do the spellings differ behaviorally?

**Yes.** This is a real defect, not a prose cleanup. `+` had two extent
functions, and they answered §17 L3's own example differently:

```
[^n]: note
+
attached
> outside
```

attached BOTH blocks to the note, where the same document under `- note`
attaches one and leaves the quote at document level. A definition description
behaved like the footnote body. Separately, the block-quote path carried a
`QUOTE` test in its boundary set - a branch on the attached block's KIND, which
no clause states - so

```
> quoted
+
> attached
```

fell through to ordinary lazy continuation instead of nesting.

Both divergences were **unreachable from every corpus document**: unifying the
paths moved zero goldens and left the suite at the same pass count. That is the
finding rather than a reassurance - either reading could drift without a red
test.

## The authoritative spelling

The list-item path, because it is the one that matches the written clause.
§17 L3 already states the boundary set (blank line, sibling marker, further `+`)
and `ONE IS A COUNT AND THE BOUNDARY IS NOT`; and
`continuation_marker_block = continuation_marker, block` has always been a
single `block`. The footnote/`dd` path was missing the narrowing and the quote
path had an extra boundary neither the production nor the clause mentions.

## What changed

- One extent function, `attachedBlockEnd`, at all four call sites. `blank` and
  `sibling` are the container's view of §17 L3's boundary set, not extra rules.
- §17 L3 gains `ONE OPERATION, EVERY CONTAINER` (NORMATIVE): the marker
  transfers ownership of the next flush-left block and never branches on that
  block's kind.
- §17 L4 gains FOOTNOTE BODY and DEFINITION DESCRIPTION. §17 L5's "a `+` attach
  form is possible future work, deliberately OUT OF SCOPE" goes: it contradicted
  the footnote and definition productions, corpus
  `66-footnote-with-multiple-blocks-2`, and both readers.
- Category 427 pins the three shapes plus a two-marker control.

## Goldens

None moved. The only new fixtures are category 427's four.

The fixtures discriminate: reverting `scripts/spec/layout.mjs` to its previous
state fails three of the four (`-2`, the two-marker control, already passed).

## Ledger

One row ADDED to `resources/engine-pin-drift.txt`: the pinned carve-js build
still folds the quote-after-a-marker case in as lazy continuation, so
`427-...-in-every-container-4` is declared. The pin reproduces the other three
new documents byte for byte - the footnote and description fixes move the oracle
ONTO the engines, not away from them.


dereuromark:
Closing as a duplicate: #1787 landed the same unification (same `attachedBlockEnd`, same corpus 427 slugs) while this was in flight.

Issue #1802, opened 2026-08-26
ast-conformance: two formatter shards red on footnote ordering and the canonical content column


dereuromark:
The daily `ast-conformance` workflow on `markup-carve/carve` is red on two
formatter shards. `conformance` passes; these two do not:

- `formatter (shard 1/4)`
- `formatter (shard 3/4)`

Reproduced against spec `1774be24`, carve-js `983b4f24`, carve-php `90a5cbc1`,
carve-rs `0c54d2ea`.

## Diff 1 - carve-js reorders hoisted footnote definitions in the non-HTML writers

Document `417-an-authored-base-carries-opaque-payload-captions-and-nested-metadata-4`,
targets `markdown`, `plain`, `ansi`. Source:

```
[^outer]: intro

     [^inner]: note

     see[^inner]

see[^outer]
```

All three engines agree on the **canonical** (`carve`) form, and the round-trip
invariant holds for each:

```
see[^outer]

[^outer]: intro

  see[^inner]

[^inner]: note
```

carve-rs and carve-php carry that order into the `markdown` target:

```
see[^outer]

[^outer]: intro

see[^inner]
[^inner]: note
```

carve-js does not:

```
see[^outer]

[^inner]: note
[^outer]: intro

see[^inner]
```

`inner` is emitted before `outer`, the two definition lines are glued into one
block, and `outer`'s second body paragraph is detached below them where it reads
as document text. **carve-js is wrong**: it contradicts the HTML numbering
(`outer` is `fn1`, `inner` is `fn2`) and its own canonical writer, which gets the
order right. Same defect in `plain` and `ansi`.

## Diff 2 - carve-php writes the authored content column instead of the minimum

Document `423-one-authored-base-rule-reaches-a-definition-nested-in-a-list-item`,
target `carve`. Source:

```
- intro

   :: term
   :  definition

      > quote
```

carve-rs and carve-js:

```
- intro
  :: term
  : definition

    > quote
```

carve-php:

```
- intro
   :: term
   : definition
     > quote
```

carve-php keeps the authored 3-space column and drops the blank line before the
quote. The round-trip invariant holds for all three, so PART 11 §1 does not
discriminate - but canonicality does. **carve-php is wrong.**

## Clauses

- `docs/experiments/definition-footnote-authored-bases.md`, the Decision clause:
  "A recognized block opener at or beyond the innermost open container's minimum
  content column belongs to that container. Its authored column is the local base
[...]

PR #1805, opened 2026-08-26
fix: the roundtrip summary drops the fixture claim it zeroed

dereuromark:
The residue of the reporting defect in markup-carve/carve#1802, after
markup-carve/carve#1804 closed the larger half of it.

`compare:impls` computes once and gates twice. `#1804` made the roundtrip run
GATE its cross-implementation count rather than defer it, which is the right fix
and is why that condition stays in the success sentence.

What is left: `mismatches` is still set to 0 outright in roundtrip mode - the
fixture count belongs to the default run and to `check-compare-report.mjs`
reading the written report - and the sentence still claims it. So the roundtrip
run continues to assert a condition it deliberately did not evaluate.

That is the same defect as the one just fixed, and the same class
`markup-carve/carve#755` catalogs: a clean bill for a check that did not run,
reached from the reporting side rather than the gating side.

After:

```
No round-trip or cross-implementation differences, no PART 11 §1 invariant failures, and no cross-read failures.
NOT checked here: fixture mismatches. This run counted them and wrote them to
/tmp/compare-impls.json, and 'npm run compare:report -- <report>' is the gate that
decides them.
```

Named rather than merely dropped: a reader told the condition is decided elsewhere
goes and reads it, where a reader of a sentence that silently omits it learns
nothing about who owns it. The default run still claims the fixture mismatches it
does gate, and that is the control in the new test.

The test drives `compare-impls.mjs` with stub engines, reusing the scaffolding
`tests/a-failed-engine-run-is-named.test.mjs` established: the property under test
is what the runner prints on a run where nothing is wrong, which is exactly the
run whose sentence was false.

Rebased onto `#1804`; the original version of this branch fixed both halves and
has been reduced to the part that is still open.


dereuromark:
Blocked on markup-carve/carve#1806 only. The one red check here is `declarations and engine pin are current`, which fails on `main` too: merging markup-carve/carve-js#1546 left this repo's carve-js pin behind, and markup-carve/carve#1798's staleness check fails every commit until the pin moves. #1806 moves it. I will rebase this branch on top once that lands - both touch `package.json`, so it needs a rebase rather than a merge.

Issue #1803, opened 2026-08-26
The oracle renders a comment-then-blank list item tight where L1b and every engine make it loose


dereuromark:
L1b AN INVISIBLE LINE DOES NOT CANCEL THE SEPARATION is pinned in one line order and wrong in the other. The oracle at `e3b0333f` renders the unpinned spelling TIGHT where all three engines render it LOOSE.

## Pinned, and correct

corpus 186-an-invisible-line-does-not-cancel-a-blank-line-separation, blank line THEN comment:

```
- a

  %% n
  text
```

```html
<ul>
  <li><p>a</p>
    <p>text</p>
  </li>
</ul>
```

## Unpinned, and wrong

The same two paragraphs with the comment BEFORE the blank line:

```
- para
  %% c

  more

x
```

oracle:

```html
<ul>
  <li>para
    more
  </li>
</ul>
<p>x</p>
```

carve-js `983b4f24`, carve-php `90a5cbc1` and carve-rs `0c54d2ea` all render:

```html
<ul>
  <li><p>para</p>
    <p>more</p>
  </li>
</ul>
<p>x</p>
```

## Why loose is right

L1b's own argument does not mention line order: a line that renders nothing "is not a separator, so the separation it appears to interrupt is intact". `para` and `more` are two paragraphs with a blank line between them, and the comment is not a paragraph and not a separator. The item is loose either way the two lines are ordered, and a rule that changed with the order would need a reason there is none of.

This is the one-rule-two-spellings shape: the corpus pinned the order somebody happened to write, the mirror went unmeasured, and the oracle answers them differently.

## Work

Fix the oracle, then write the mirror row. Found while measuring the band for #1800; kept out of that branch because it is a different clause and the two fixes are independent.

Issue #1800, opened 2026-08-26
Below a definition body's column the five invisible-line kinds give four different answers


dereuromark:
## What was measured

The oracle at `origin/main` (`1774be24`), five invisible-line kinds written one or two columns BELOW a definition description's content column, with a plain line as the control. Host in every case:

```
:: t
:  d
```

so the body's content column is 3, and the probe line sits at column 2. `tail` follows at column 0.

## Four different answers

Control, a plain line:

```
:: t
:  d
  x
tail
```

```html
<dl>
  <dt>t</dt>
  <dd>d
x
tail</dd>
</dl>
```

The body survives and the line folds in. A comment:

```
:: t
:  d
  %% c
tail
```

```html
<dl>
  <dt>t</dt>
  <dd>d</dd>
</dl>
<p>tail</p>
```

Invisible, the body ends, `tail` is a document sibling. A link reference definition:

```
:: t
:  d
  [r]: /u
tail
```

```html
<dl>
  <dt>t</dt>
  <dd>d</dd>
</dl>
<p>[r]: /u
tail</p>
```

The body ends and the line is published as text at DOCUMENT level, not lazily into the body. The footnote definition spelling behaves the same. An attribute line:

```
:: t
:  d
  {.k}
tail
```

```html
<dl>
  <dt>t</dt>
  <dd>d</dd>
</dl>
<p>tail</p>
```

The line is gone: no text anywhere, no attribute attached to `tail`. That is filed as #1801, since it is a defect on its own rather than a missing ruling.

An abbreviation definition:

```
:: t
:  d
  *[A]: a
tail
```

```html
<dl>
  <dt>t</dt>
  <dd>d
*[A]: a
tail</dd>
</dl>
```

Folds into the body like the plain control, which follows from AN ABBREVIATION DEFINITION IS RECOGNIZED ONLY AT DOCUMENT LEVEL: in a container it was never an invisible line, so it is ordinary text and lazy continuation applies.

## What the prose currently claims

`resources/grammar.ebnf` at PART 9 section 10 I5 says:

> At a nonzero column BELOW `content_column` it is lazy paragraph text and does not register.

[...]

PR #1807, opened 2026-08-26
spec: an attribute line below a container's column keeps its characters

### LINKED ISSUE #1801: An attribute line at column 0 under a definition description is discarded, and a blank line flips it
## The defect

An attribute line written at COLUMN 0 directly under a definition description is discarded. The same line under a list item or a block quote attaches, and inserting a blank line above it makes the definition-list case attach too.

Measured on the oracle at `origin/main` (`1774be24`).

A list item, which is the reference answer:

```
- d
{.k}
tail
```

```html
<ul>
  <li>d</li>
</ul>
<p class="k">tail</p>
```

A block quote agrees:

```
> d
{.k}
tail
```

```html
<blockquote><p>d</p></blockquote>
<p class="k">tail</p>
```

A definition description does not:

```
:: t
:  d
{.k}
tail
```

```html
<dl>
  <dt>t</dt>
  <dd>d</dd>
</dl>
<p>tail</p>
```

The `dl` ends at the attribute line either way, so by the time `tail` is classified the attribute is at document level with a visible block after it. It should attach. Add one blank line and it does:

```
:: t
:  d

{.k}
tail
```

```html
<dl>
  <dt>t</dt>
  <dd>d</dd>
</dl>
<p class="k">tail</p>
```

A blank line is not supposed to decide this. A2 FLOAT FORWARD says pending survives blank lines; it does not say a blank line is what makes pending exist.

## Not the same as the pinned drop

Corpus 329-a-floating-attribute-is-scoped-to-the-container-that-holds-it-5 pins the drop AT the description's content column:

```
:: t
:  d
   {.k}
tail
```

```html
<dl>
  <dt>t</dt>
  <dd>d</dd>
</dl>
<p>tail</p>
```

That one is correct and stays correct: the attribute is inside the description, the description closes with nothing left to attach to, and section 15 A4 discards it. Column 0 is outside the description, which is exactly why the list-item and block-quote spellings attach. Two different columns, one behavior, and only one of them is ruled.

## Also unpinned, same kind of loss

Below the content column, at column 1 or 2, the attribute line disappears as well:

```
:: t
:  d
  {.k}
tail
```

```html
<dl>
  <dt>t</dt>
  <dd>d</dd>
</dl>
<p>tail</p>
```

A link reference definition at that same column publishes its text, and a plain line folds into the description, so the attribute is the only kind whose authored characters reach neither the page nor a symbol table. In a list item the same spelling folds as literal text:

```
- d
 {.k}
tail
```

```html
<ul>
  <li>d
{.k}
tail</li>
</ul>
```

which is what AN INVISIBLE LINE FOLDS LIKE ANY OTHER asks for.

## The ask

1. At column 0 under a description, the attribute is a document-level floating attribute and attaches to the next visible block. Make the no-blank spelling agree with the blank spelling and with the other two hosts.
2. Below the content column, decide between literal text (the list-item answer) and discard, and state it. Whichever wins, A4 requires the discard be reported as `unattached-block-attribute` rather than silent.
3. Corpus rows for both columns, since neither is pinned today.

Related: the band itself gives four different answers across the five invisible-line kinds, filed as #1800. Both are premises for the consolidation in #1783.

dereuromark:
Closes #1801. Rules and pins the band #1800 asked about, minus two rows that wait on the engine.

## The defect

`foldablePlain` decides whether a below-column line is PULLED INTO a definition body, and the pull dedents the line to the body's column 0. It knew about the two definition kinds and not about the attribute line, so an attribute line was pulled in and handed to the one column where it stops being literal text and becomes a real attribute block. Inside the body nothing was left to attach to, the block dangled, and the line was dropped: neither the class nor the characters reached the output.

```
:: t
:  d
  {.k}
tail
```

before

```html
<dl>
  <dt>t</dt>
  <dd>d</dd>
</dl>
<p>tail</p>
```

after, which is what carve-js, carve-php and carve-rs already produce

```html
<dl>
  <dt>t</dt>
  <dd>d</dd>
</dl>
<p>{.k}
tail</p>
```

Excluding the attribute line from the predicate ends the body instead, and the surviving document context reads the line where it was authored. That recovers the column-0 spelling in the same move, so the two symptoms in #1801 are one fix:

```
:: t
:  d
{.k}
tail
```

now attaches, matching a list item, a block quote and this host's own blank-line spelling. This is #1786's defect one construct over: that pass moved the question to the dedented line so a below-column quote marker could not open a quote it opens nowhere else, and it reached the openers only.

The drop AT the content column is untouched and still correct, pinned by corpus 329-a-floating-attribute-is-scoped-to-the-container-that-holds-it-5.

## The band, ruled rather than counted

#1800 asked whether the band needed a ruling. It does not: a definition line is block-shaped, so below the body's column it ends the body, and at a nonzero column in the document context it is no longer a definition either, so it is text that registers nothing. The two kinds that look like exceptions fall out of the same chain. An abbreviation definition in a container was never a definition, so it never enters the chain and folds as prose. An attribute line enters it and lands as an indented attribute line, which corpus 157-indented-attribute-line-stays-literal already reads as literal.

Nine rows in three categories: the band per kind, the column-zero attribute with its blank-separated control, and four content-column cells no document reached.

I5's below-column sentence moves with them. It said the line is "lazy paragraph text", which names the list-item case as if it were the rule - a list item has a paragraph open and the line folds in, a definition body ends at any line below its own column and the text is the document's. The registration half was right and is unchanged.

## Two rows held back, and four engine defects

`npm run compare:counts` over all 1487 documents finds exactly two mismatches, and both are filed:

- markup-carve/carve-rs#1438, a link definition folds where the same build's footnote spelling ends the body
- markup-carve/carve-php#1794, the column-0 attribute is discarded

Two more rows are not in this PR at all, because the pinned reference build does not reproduce them and declaring the drift would leave the declaration audit red:

- the abbreviation definition in the band, markup-carve/carve-js#1544
- the comment at a footnote body's content column, markup-carve/carve-js#1545

Both land with the pin bump that clears them.

## Notes for the reviewer

No committed golden moves. The oracle reproduced 1480 documents before the change and 1487 after, which is the evidence that only unpinned cells were wrong.

The comparison page's denominators are counted live, so nine new documents move them past the run they describe. The three categories are declared under "added since this run" rather than retaking a snapshot that would publish new engine SHAs and timings for a one-predicate change.

Two things found on the way that are NOT in this branch: the oracle renders a comment-then-blank list item tight where L1b and every engine make it loose (#1803), and `npm run declarations:check` is red on main because the carve-js pin is one commit stale, which a pin bump clears.


dereuromark:
Closing this as superseded, not as wrong. The analysis in it is sound; the ground moved underneath it.

**Its ticket is already closed.** `markup-carve/carve#1801` was closed at 10:38Z by `markup-carve/carve#1809`, which landed the same fix - the below-column fold and the column-zero attach as one change, for the same reason stated here.

**The categories collide, with more coverage on the other side.** All three numbers are taken on `main` by different slugs covering the same ground:

| this PR | on `main` |
| --- | --- |
| `429-...-each-invisible-line-keeps-its-own-answer` (4 docs) | `430-below-a-definition-body-s-column-an-invisible-line-folds-as-text` (6 docs) |
| `430-...-attaches-from-column-zero` (2 docs) | `431-a-floating-attribute-under-a-definition-attaches-at-column-zero` (4 docs) |
| `431-every-invisible-line-at-a-container-s-content-column` (3 docs) | `432-an-abbreviation-definition-outside-document-level-is-not-an-invisible-line` covers two of the three rows |

`429` on `main` is a different subject entirely (`an-invisible-line-before-the-blank-does-not-cancel-the-separation`, from carve#1808).

## One row is worth salvaging

The base document of this PR's `431` does not appear to be pinned anywhere:

```
:: t
:  d
   [^f]: n

see[^f]
```

A **footnote definition at a description's content column**. `main` pins below the column (430), at column zero (431), the abbreviation exception (432), and an abbreviation at a *list item's* content column (194) - but not this. That is a sampled search rather than an exhaustive one, so it is worth confirming before adding.

If it is genuinely unpinned, it belongs in a follow-up as a single row rather than a rebase of this branch: the rest of this PR is now duplicate, and rebasing it would mean renumbering three categories to re-add coverage that already exists in better form.

PR #1817, opened 2026-08-26
spec: the continuation marker's column gate reaches every container

### LINKED ISSUE #1814: The continuation marker's column gate reaches one container out of four
`AND FLUSH-LEFT MEANS COLUMN 0` (§17 L3, NORMATIVE, carve#1436) says:

> the marker attaches a block that begins at column 0 and NOTHING else. A line at any other column is not attached at all - it falls through to the ordinary column rules, which give it to whichever container its own column names, exactly as if the `+` line had been a comment.

That holds in the LIST ITEM and nowhere else. Measured on `origin/main` (`3aa62a71`) against the oracle and the pinned carve-js (`4a2f8389`); the two agree with each other on every row below and both contradict the clause in three of the four L4 containers.

The clause names its own control - "exactly as if the `+` line had been a comment" - so each row is the same document twice, once with the marker and once with a comment in its place.

## Definition description

The description's content column is 3, so column 2 is below it and the ordinary rules place the line outside the `dd`. The comment spelling agrees:

```
:: t
:  a
%% c
  para
```

```html
<dl>
  <dt>t</dt>
  <dd>a</dd>
</dl>
<p>para</p>
```

The marker pulls the same line in:

```
:: t
:  a
+
  para
```

```html
<dl>
  <dt>t</dt>
  <dd>
    <p>a</p>
    <p>para</p>
  </dd>
</dl>
```

## Footnote body

Body minimum column 2, so a line at column 1 is below it. Comment spelling - the body ends and the paragraph lands at document level:

```
[^n]: a
%% c
 para

see[^n]
```

```html
<p>para</p>
<p>see<a id="fnref1" href="#fn1" role="doc-noteref"><sup>1</sup></a></p>
<section role="doc-endnotes" aria-label="Footnotes">
  <hr>
  <ol>
    <li id="fn1">
      <p>a<a href="#fnref1" role="doc-backlink" aria-label="Back to reference">↩</a></p>
    </li>
  </ol>
</section>
```

Marker spelling - the paragraph is inside the note:

```
[^n]: a
+
 para

see[^n]
```

```html
      <p>a</p>
      <p>para<a href="#fnref1" role="doc-backlink" aria-label="Back to reference">↩</a></p>
```

The same happens at column 2, the body's own minimum: the comment spelling ejects the line, the marker keeps it.

## Block quote

Column 2 with no `>` is in no quote at all - `A QUOTE IS REACHED BY ITS MARKER, AND A COLUMN NEVER REACHES INTO ONE` (§10 I5, carve#1384) - and the blank-line control agrees:

```
> a

  para
```

```html
<blockquote><p>a</p></blockquote>
<p>para</p>
```

The marker reaches into the quote for it:

```
> a
+
  para
```

```html
<blockquote>
  <p>a</p>
  <p>para</p>
</blockquote>
```

## Why it is exactly the shape carve#1782 named

carve#1782 found the marker implemented as two operations: one extent function for list items and block quotes, another for footnote bodies and descriptions. #1787 unified the ONE BLOCK count and removed the quote kind-branch. The COLUMN gate was not part of that unification and is still in one place only - `scripts/spec/layout.mjs:3490`, in the list-item `attachFlushLeft` path:

```js
if (!first || (first.L ? first.L.col : first.col) !== 0) return false
```

The other three attach sites (`layout.mjs:1974` footnote body, `:2249` description, and the quote path) have no equivalent test. Removing that one line moves exactly two goldens - `384-a-continuation-marker-attaches-only-a-flush-left-block-2` and `-6` - and category 384 is entirely list items. So the clause is pinned for one container out of four, which is why three readers could violate it without a red test.

`THE MARKER IS ONE OPERATION IN EVERY CONTAINER` (§17 L3, carve#1782) forbids exactly this: "A container may not answer differently because of WHAT it is attaching", and a fourth container answering differently about WHICH LINE it attaches is the same defect one step earlier.

## Why this is filed instead of pinned

It came out of carve#1792 item 4, which asked for corpus rows for the continuation-attached band. The band cannot be pinned as it stands: a golden recording the current answer would contradict a NORMATIVE clause, and re-snapshotting a wrong reading is worse than leaving the rule undocumented. So the rows wait on the ruling.

Note also that the two readers agreeing here is not evidence: the clause's own history paragraph records that all four readers attached a line at any indentation before carve#1436, and only the list-item path was ever fixed.

## Work

1. Apply the column gate at the other three attach sites, or state per container why the reach differs - implementation convenience does not count (carve#1782's own decision gate).
2. Corpus rows for the band in a footnote body, a description and a block quote, with the comment or blank-line control alongside each, which is what carve#1792 item 4 wanted.
3. A naive `ind(i).col === 0` guard at `layout.mjs:1974` and `:2249` does NOT work - it moves nothing, because `ind` there is relative to the active strip rather than to document column 0. The fix needs the document column, the way `attachFlushLeft` reads `first.L.col`.

dereuromark:
Fixes markup-carve/carve#1814.

`AND FLUSH-LEFT MEANS COLUMN 0` (§17 L3, carve#1436) says the `+` marker attaches a block that begins at document column 0 **and nothing else** - a line at any other column is not attached at all and falls through to the ordinary column rules, "exactly as if the `+` line had been a comment". That question was asked in the list item and nowhere else.

## What was wrong

The gate had two spellings, both in list code: `attachFlushLeft`'s `first.L.col !== 0`, and the item collector's `sourceCol === 0`. The footnote body, the description and the block quote had no equivalent, so each reached out for a line the clause leaves where the author wrote it. Each row is the same document twice, once with the marker and once with the invisible line the clause names as its control.

A footnote body, below its minimum column:

```
[^a]: intro
+
 more

see[^a]
```

put `more` inside the note, where the comment spelling ejects it to document level. At column 2, the body's own minimum, the marker split the body's paragraph while the comment spelling ended the body.

A description whose content column is 3:

```
:: term
:  intro
+
  more
```

pulled a column-2 line into the `<dd>`.

A block quote, with the quoted paragraph closed so the row asks about the column and nothing else:

```
> intro
>
+
  more
```

put `more` in the quote, where both the blank-line control and the comment control leave it outside - and `A QUOTE IS REACHED BY ITS MARKER, AND A COLUMN NEVER REACHES INTO ONE` (§10 I5, carve#1384) says a column-2 line is in no quote at all.

## The fix, and it is one site

The gate is now `attachesFlushLeft(meas)`, defined once next to `attachedBlockEnd`, which applies it before measuring any extent. Both list spellings were deleted in favor of it. It reads the SOURCE column (`meas.L.col`), not the frame column: inside a container the frame column is what the strip left, not what the author wrote, which is why both surviving spellings already read `.L.col`.

A refused marker is consumed and contributes nothing, and the line is then handled by the container's ordinary rules - which is the clause's own comment analogy. Two of the four containers need the predicate one line earlier than the extent, at the `+` itself: a footnote body and a `<dd>` END at an invisible line at document column 0, so a `+` that turns out not to be a marker has to end them at the same place a comment does, and the branch that would otherwise have claimed the line runs before the extent is ever measured. That is the container's invisible-line rule, not a second column rule - the predicate is still the one definition.

## Two rows in the ticket were not gate defects

The ticket reads the footnote's column-2 row as a violation on the strength of a column-0 comment control. The band at the body's own minimum column is claimed by the continuation branch before the pull branch is reached, so the marker never decided it; an INDENTED comment at that column agrees with the marker exactly (pinned in "A comment in a footnote body is invisible in both spellings"). What actually differs is that a column-0 invisible line ends the body. The fix makes the refused marker do that too, so the row lands where the ticket wanted it, but by the container's rule rather than by the gate.

The quote's open-paragraph spelling is left folding lazily. A column-0 comment under an open quoted paragraph is folded into it as lazy TEXT rather than skipped, which is a defect of the quote's invisible-line handling (the carve#1813 family) and not of the marker; the corpus row closes the paragraph with a bare `>` first so it asks about the column alone, and there blank, comment and marker all agree.

## Pinned

Category 435, thirteen documents: five marker rows, five controls (four comment spellings and one blank-line spelling), three column-0 positives, across all four containers. `tests/the-continuation-marker-s-column-gate-reaches-every-container.test.mjs` pins the RELATION the clause states - marker spelling equals control spelling - which no single golden can express.

Engines follow: carve-js, carve-php and carve-rs all read the pre-ruling answers, so the new rows are ahead of them until each engine ships the gate.


dereuromark:
One site this PR deliberately does NOT gate, from a review pass.

The FIRST-BLOCK form `:  +` sets `pullPending` on the entry line, before the loop and before the shared predicate is asked, so it is a fifth attach site that never asks the column question. Adding the guard there is measurable dead code: with it and without it,

```
:: term
:  +
  more
```

renders `<dl><dt>term</dt><dd>more</dd></dl>` at columns 0, 1 and 2 alike, because the below-the-column fold claims a plain line into an empty description body whether or not a marker put it there. Under the "always refuses" mutation it does not move either, so the guard cannot fail in either direction, and this repo has enough history with checks that cannot fail.

What is worth a ruling separately: the marker and the comment spellings of that form disagree, since `:  %% c` over `  more` pushes the comment as body text, closes the paragraph and ejects the line, while `:  +` folds it in. That is a question about the FIRST-BLOCK form's control, not about the column gate, so it is not decided here.

Issue #1821, opened 2026-08-26
An empty description body claims a following line where the list item and the comment control do not


dereuromark:
`AND FLUSH-LEFT MEANS COLUMN 0` (§17 L3, NORMATIVE) gives the marker its own control: a refused `+` behaves "exactly as if the `+` line had been a comment". Measured against the oracle on `main`, two shapes break that control. Found while landing markup-carve/carve#1817, which fixed the gate itself; these are next to it, not it.

## 1. The first-block form: an empty description body claims the following line

Payload one column in, so the marker must be refused:

```
:: t
:  +
 flush
```

```html
<dl>
  <dt>t</dt>
  <dd>flush</dd>
</dl>
```

The comment control, same position:

```
:: t
:  %% c
 flush
```

```html
<dl>
  <dt>t</dt>
  <dd></dd>
</dl>
<p>flush</p>
```

**The marker gate is not what causes this.** Adding the gate guard to the first-block form is measurable dead code - output is identical at columns 0, 1 and 2 with or without it, because the below-column fold claims a plain line into an empty `dd` **regardless**. The comment never reaches that fold, because it is invisible and ends the body first.

So the question is not about `+` at all. It is whether an **empty description body** may claim a following line.

**RULING: it may not. The body ends, matching the comment.** An empty `dd` never claims a following line. The fix belongs in the below-column fold, not in the marker gate.

## 2. A marker at the description's own content column renders as literal text

```
:: t
:  d
   +
flush
```

```html
<dl>
  <dt>t</dt>
  <dd>d
+
flush</dd>
</dl>
```

The `+` is emitted as literal text and drags the following line into the `dd`. The comment control ends the body:

```
:: t
:  d
   %% c
flush
```

```html
<dl>
  <dt>t</dt>
  <dd>d</dd>
</dl>
<p>flush</p>
```

**RULING: the marker ends the body, as its control does.** A literal `+` in the output is a fall-through rather than a decision. `flush` belongs at document level.

## Scope

Both are the same clause's control failing in the same host, so they are likely one fix in the description's below-column handling rather than two.

Pin each pair - marker spelling and comment spelling - as a **relation**, not two independent goldens. That is what markup-carve/carve#1817 did for the gate, and it is the only shape that catches a change repairing one spelling and drifting the other.

Check the other three L4 containers before assuming the description is special. If a footnote body, a list item and a block quote answer either shape differently, that is part of the same ruling and belongs in the same pass.

Engine ports will be needed once this lands.

[...]

PR #1823, opened 2026-08-26
spec: a resolution result about a wrapper is not content of it

dereuromark:
Two things, and the second is why the first cannot be a separate PR.

1. **The pin.** Moves the spec's own carve-js devDependency from `bcb5cf4f` to `e0119621`, and deletes the `paragraph.blockImage` row the bump makes stale.
2. **The clause.** Widens PART 11 §1c's bare-wrapper bound to admit `paragraph.blockImage`, per the ruling on this PR. New normative sub-clause `CARVE-P11-039`, *a resolution result about the wrapper is not content of it*.

## Why the clause moves

`isBareWrapper` decided bareness by a KEY COUNT - exactly `type` and `children`, no third key. `paragraph.blockImage` makes three, so the day carve-js began publishing the field §1c stopped recognising the wrapper it was written to forgive, and a wrapper loss the clause already tolerated was recounted as a **field** loss. Three `411-a-lone-indented-image-...` documents went red for that reason and no other.

carve-js did not regress. Only one side of the round trip moved: flush-left `parse(fmt(x))` is byte-identical on both pins, the sole delta being `blockImage: true` on the indented paragraph. All three documents still render a bare `` with no `<p>`, and `engine:report --check` reproduces every corpus pair. The indented image was already a block image before the field existed; the field only made the existing fact visible to a key count.

PART 12 §23 and `resources/ast-schema.json` bless both spellings - an `image` in block position, and a `paragraph` carrying `blockImage` around one `image` - so dissolving either into the other loses nothing, which is what §1c is for. Leaving the bound unwidened would instead stop §1c reaching **every** block image the moment the field is published: a wider and quieter change than admitting one key.

**It is one key, not a policy.** The clause says so and the checker enforces it: a key no clause has ruled on is content, `blockImage: false` is a spelling PART 12 §23 forbids, and an attribute block beside the field is exactly as disposable as one alone.

## The §1c guards, both directions

The acceptance test is that the widened branch is REACHED on the new pin and still refuses everything §1c does not forgive. Both halves measured by reverting the change:

| predicate | `node --test tests/corpus-fmt-roundtrip.test.mjs` |
| --- | --- |
| as shipped here | **exit 0**, `pass 12 / fail 0` |
| reverted to a key count | **exit 1**, `fail 4` - the sweep, the new carve-out guard, the bound test and the strong property |
| widened to "any key that is not `attrs`" | **exit 1**, `fail 1` - the bound test |
| key admitted, value unchecked | **exit 1**, `fail 1` - the bound test |

A new guard asks the corpus directly whether the widened branch is reached, separately from the ceiling guard beside it - the ceiling is satisfied by a paragraph holding one comment, so the metadata carve-out could stop being exercised by anything and that test would stay green.

Re-measured over all 1527 corpus documents and their 12027 canonical nodes: admitting the field decides **3 nodes** differently, one in each `411` document, and moves those three from failing the sweep to passing it. Unlike the three widenings already recorded above that test, this one is not a no-op, so it is asserted rather than swept.

## What the rebase surfaced

`html-import:check` is not part of `npm test`, and the pin move left `resources/import-roundtrip-baseline.json` behind. Only `renderImportRoundTrips.markdown` moves, and it is **four** documents rather than two:

- `435-...-2`, `-4` and `-13` now survive the markdown render/import round trip, because the column-gate work in the new pin reads them correctly
- `436-...-7` leaves it: the pin now reads `- +` over a flush-left line as a refused marker, and the markdown target has no spelling for that

Every other count in the ratchet is unchanged. `engineCommit` moves to the new pin, since that is what the numbers were measured against.

## Ledgers and gates

`engine:report --check` empties `resources/engine-pin-drift.txt`. Two of its rows **move rather than vanish** - the pin now READS them correctly and still WRITES them wrongly, which is what `resources/engine-fmt-drift.txt` is for. Both ledgers are ratcheted in both directions, so neither move can silence anything.

`declarations:check` release mode, measured with carve-js, carve-php and carve-rs as siblings, both readings against the same base:

| | findings | detail |
| --- | --- | --- |
| `origin/main` | 2 | `PIN STALENESS BEHIND` (`bcb5cf4f`, 3 behind) + 9 owed rows in `engine-pin-drift.txt` |
| this branch | 2 | `PIN STALENESS BEHIND` (`e0119621`, **1** behind) + 2 owed rows in `engine-fmt-drift.txt` |

The remaining staleness is carve-js `5c8c3dc`, a test-only commit that landed after this bump was measured; chasing it is a treadmill and it changes no behavior. The remaining owed rows are honest: the release is genuinely not clear to tag while carve-js's *writer* lags the #1821 ruling, whose reader half already ships at `e0119621`. `declarations:pr` - the mode CI gates on - passes.

## Not touched

- No `CHANGELOG.md` entry. Nothing here is consumer-visible: a devDependency pin, a spec clause and the checker that encodes it.
- No version field, no tag, no publish.


dereuromark:
`declarations:check` release mode, measured with carve-js / carve-php / carve-rs checked out as siblings:

**before** - exit 1, 2 findings:

```
PIN STALENESS  BEHIND - @markup-carve/carve bcb5cf4f vs carve-js origin/main e0119621 (pin-only 0, main-only 2)
spec  resources/engine-pin-drift.txt  -  6  owed  two-way  <== OWED
DECLARATION AUDIT FAILED (mode: release) - 2 finding(s). Not clear to tag.
```

**after** - exit 0, 0 findings:

```
PIN STALENESS  current at carve-js e0119621
DECLARATION AUDIT PASSED (mode: release) - every owed list is empty, the pin is current, and every guard is two-directional.
```

So this does not just keep the per-PR gate quiet - it flips the release verdict on a tree that four prepared drafts are waiting on.

Full suite is running locally alongside CI; will report if it disagrees.

dereuromark:
## Reading: this is candidate 2 - the PART 11 §1c bound is now too narrow. carve-js has not regressed.

Measured both pins against the same three documents. The decisive fact is that **only one side of the round-trip changed**.

### The two trees, old pin vs new

| document | `parse(src)` (indented) | `parse(fmt(src))` (flush-left) |
| --- | --- | --- |
| old `bcb5cf4f` | `paragraph[image]` | `image` |
| new `e0119621` | `paragraph{blockImage:true}[image]` | `image` |

`fmt` canonicalizes the indent away on both pins - ` Apollo` becomes `Apollo` - and the flush-left parse is **byte-identical** across the two pins. The single delta is `blockImage: true` appearing on the indented paragraph.

That is exactly why the gate flipped. `isBareWrapper` requires the wrapper to have **exactly two keys**, `type` and `children` (tests/corpus-fmt-roundtrip.test.mjs:271). Adding `blockImage` makes it three, the wrapper stops being "bare", §1c stops dissolving it, and a difference that was forgiven as a wrapper loss is now counted as a field loss. The old pin passes this file 11/0; the new one fails the two round-trip assertions.

### Why candidate 1 is ruled out

The suspicion would be that an indented top-level image should not promote. The corpus says it does, and always did:

```
 ![Apollo](a.jpg)
```

renders as

```html
<img src="a.jpg" alt="Apollo">
```

No `<p>`. An unpromoted lone-image paragraph would render one. All three documents render a bare ``, `engine:report --check` reproduces 1514/1514 pairs with 0 mismatched, and those HTML files are unchanged by this PR. So the indented image was already a block image before `blockImage` existed - the field did not promote it, it only started *saying* it was promoted.

### Why both shapes are conformant

PART 12 §23 and the schema bless both spellings, and say so in as many words. From `resources/ast-schema.json` on the `image` node:

> Also valid in BLOCK position: a lone image paragraph is a block-level image. [...] Read the PUBLISHED tree when checking this: an engine may promote during reference resolution rather than in its block pass, so a parse-only tree can still show the paragraph (carve#1663). Where it does, `paragraph.blockImage` carries the answer.

So `document > image` and `document > paragraph{blockImage:true} > image` are two spellings of one fact. carve-js#1556's container-content-column gate selects which one you get: flush-left matches its container's content column and promotes in the block pass (the wrapper goes away), indented does not and promotes during reference resolution (the wrapper stays and carries the field). Both are legal; the engine is not wrong either way.

Which makes the loss `fmt` incurs here **the same loss wearing a new name**, which is the coordinator's own phrasing for candidate 2. Dissolving `paragraph{blockImage:true}[image]` into `image` in block position destroys no information, because an image in block position *is* the block image. §1c already licenses that dissolution; it just describes the wrapper by its key count, and the key count moved.

### Candidate 3 ruled out

`fmt` should not preserve the indent. The indent is not semantic - both spellings are block images, and the HTML is identical - so canonicalizing it away loses nothing that §1c does not already forgive.

### What I did not do

I did not add the three slugs to `resources/engine-fmt-drift.txt`, and did not relax the assertion. The sweep says plainly that no allowlist belongs there:

> There is no allowlist here and there is not to be one: a slug would silence a whole document, where this states the one difference §1c licenses and leaves every other difference failing.

### Recommendation

Widen §1c so the wrapper it dissolves may carry `blockImage: true` and nothing else - a paragraph whose only own field is `blockImage`, wrapping a single `image`. Narrow, principled, and it keeps every other field difference failing.

**My call: it belongs in this PR, not a follow-up.** On the old pin the widened branch is unreachable - no document produces a `blockImage`-carrying wrapper - so a standalone clause PR would land an exemption that nothing exercises, and `a corpus document still reaches the PART 11 §1c ceiling` plus `the PART 11 §1c bound reaches a bare single-child wrapper and no other shape` exist precisely to stop that. The bump is what makes the clause testable.

That turns this from a pure `chore` into a normative change, so it wants the label and the review to match. Flagging rather than writing the clause myself, since it is a PART 11 ruling and not pin plumbing. Holding `#1823` red and unmerged until that is settled.

Note `tests/corpus-fmt-roundtrip.test.mjs:479` (`the strong property still fails on every difference PART 11 §1c does not forgive`) also fails, for the same reason: its hardcoded expectation is the pre-promotion tree. It needs its fixture moved in whatever change settles this.

dereuromark:
## Rebased onto `#1822`, which surfaced a second, unrelated thing worth naming

`#1822` landed while this was open and added three rows to `resources/engine-pin-drift.txt` for its own `436-*` documents, written against the old pin. Under the new pin all three **render** correctly - `engine:report --check` reports 1521/1521 reproduced, 0 mismatched, and calls all three `STALE`.

But two of them still fail `formatting never changes what a corpus document says`. The pin now READS them correctly and still WRITES them wrongly, flattening the refused marker back to `: flush`.

That is the exact distinction the two ledgers encode, per `tests/fmt-drift.mjs` and the files' own headers:

- `engine-pin-drift.txt` - the FIRST render. Ratcheted in both directions by `engine:report --check`, so a row cannot outlive the drift it names.
- `engine-fmt-drift.txt` - "documents the pinned build READS correctly and WRITES wrongly." Ratcheted by `every writer-drift line still names a document the pin writes wrongly`.

So the two writer-drifting slugs **moved** rather than vanished, and `436-...-7` left outright - it now reads and writes correctly. The writer-side window itself is unchanged by this PR; only its correct home moved, because the bump fixed the render half and not the writer half. Nothing was silenced: both ledgers are ratcheted in both directions, and a row in either one fails the moment it stops being true.

Worth flagging for whoever owns the `#1821` engine ports: **carve-js already ships the reader half** of that ruling at `e0119621`, so the remaining port there is the writer.

## Current state

`tests/corpus-fmt-roundtrip.test.mjs` is down to the two failures that are the real question, both the 411 documents and both the §1c bound:

```
not ok 4 - formatting a corpus document re-parses to the same tree (PART 11 §1, bounded by §1c)
not ok 7 - the strong property still fails on every difference PART 11 §1c does not forgive
# pass 9
# fail 2
```

Everything adjacent is green - `corpus-fmt-cross-read`, `drift-ledger-rejects-a-duplicate`, `schema-fields-are-produced`, `engine-pin-matches-the-lock` and `corpus` run 1539 pass / 0 fail together.

Still not merging. The §1c call from the previous comment is the open item, and it is a PART 11 ruling rather than pin plumbing.

dereuromark:
## Second finding: the writer half of a spec-rule window has no home a gate tolerates

Separate from the §1c question, and worth a decision of its own.

Under the new pin the three `#1822` documents split: the pin now **reads** them right and still **writes** two of them wrong. There are only three places that fact can live, and all three are red:

| where the two slugs go | what fails |
| --- | --- |
| stay in `engine-pin-drift.txt` | `engine:report -- --check` - it measures the first render only, sees them reproduce, and reports `STALE ... drop the line` |
| move to `engine-fmt-drift.txt` | `declarations:check` **per-PR** - 1 finding, `resources/engine-fmt-drift.txt ... 2 owed` |
| delete outright | `formatting never changes what a corpus document says` - the writer really does change the document |

I took the middle one, because it is what that file's own header asks for:

> The pin can fail to read its own output back, and the ORACLE can fail to read the pin's output back while the pin reads it fine - **which is what a spec rule the pin has not shipped looks like from the writer side.** Both belong here.

That is this situation exactly. The placement is right; the gate just does not tolerate it.

### Why it is red anyway

`scripts/declaration-audit.mjs` gives the two ledgers different per-PR policies:

```js
{ path: 'resources/engine-fmt-drift.txt', policy: 'owed', guard: 'two-way', ... }
{ path: 'resources/engine-pin-drift.txt', policy: 'owed', prPolicy: 'declared', guard: 'two-way', ... }
```

carve#1811 added `prPolicy: 'declared'` to the render ledger on the reasoning that "a release cannot ship with that window open; a PULL REQUEST is how the window opens in the first place", and that gating both on emptiness "made every leading spec ruling red by construction". The writer ledger encodes the **same kind of window** - its header says so in as many words - and did not get the same treatment. Nothing protected it from the asymmetry before, because the file was empty.

So a spec ruling that lands ahead of the engines is tolerable per-PR on the render side and red on the writer side, for no reason I can find in either file.

### Two ways out, and I am not picking unilaterally

1. **carve-js ships the writer half of #1821.** Then `engine-fmt-drift.txt` empties on its own and the asymmetry stays theoretical. This is the natural fix and it is the port another agent already has - carve-js is half-shipped at `e0119621`, reader done, writer not.
2. **Give `engine-fmt-drift.txt` the same `prPolicy: 'declared'`**, on carve#1811's own reasoning. That is a policy change to the audit manifest and wants its own ticket.

Option 1 makes this PR green without any policy change, so it is worth trying first.

I have **not** relaxed the assertion, added a slug to silence a document, or touched the audit manifest. Both findings this bump turned up - the §1c bound and this one - are left standing and red on purpose.

dereuromark:
Ruling on the finding in this PR: **widen §1c's bare-wrapper test to admit `blockImage`, and land it here.**

## Why this is forced rather than a judgment call

`isBareWrapper` requires exactly two keys, `type` and `children`. `paragraph.blockImage` makes three, so §1c stops recognising the wrapper it was written to forgive, and a wrapper loss it already tolerated is recounted as a field loss. The three `411` documents fail for that reason and no other.

The alternative reading - that carve-js regressed - is ruled out by the evidence in this PR:

- only one side of the round trip moved: flush-left `parse(fmt(src))` is byte-identical on both pins, and the sole delta is `blockImage: true` appearing on the indented paragraph
- all three documents still render a bare `` with **no** `<p>`, unchanged by the bump, and `engine:report --check` reproduces 1521/1521
- so the indented image was already a block image before the field existed; the field only made the existing fact visible to the key count

PART 12 §23 and `resources/ast-schema.json` bless both spellings - `document > image` and `paragraph{blockImage:true} > image` - so dissolving one into the other loses nothing. That is exactly what §1c is for.

Leaving it unwidened would mean §1c silently stops applying to **every** block image the moment the field is published, which is a larger and quieter change than widening it.

## Land it in this PR

The clause change is unreachable on the old pin - the field does not exist there - so it cannot be landed separately and verified. The two §1c "carve-out is reached" guards exist to catch exactly that dead branch, and they would fire.

That does make this PR a normative change rather than pin plumbing. Relabel it accordingly, describe the clause move in the body, and treat the §1c guards as the acceptance test: the widened branch must be **reached** on the new pin, and must still refuse every difference §1c does not forgive.

Do not widen further than the one key. `blockImage` is metadata about the wrapper; anything else appearing on a paragraph is content and must keep failing.

dereuromark:
Picked this up. Rebased onto current `main` (`b597d6fc`, so past #1822, #1825 and #1826) in a fresh clone; the rebase was clean and the branch diff is unchanged.

Reproduced the red on the rebased tree, `node --test tests/corpus-fmt-roundtrip.test.mjs` -> **exit 1**, `pass 9 / fail 2`:

```
not ok 4 - formatting a corpus document re-parses to the same tree (PART 11 §1, bounded by §1c)
      411-a-lone-indented-image-is-a-paragraph-and-its-html-cannot-say-so
      411-...-2
      411-...-3
not ok 7 - the strong property still fails on every difference PART 11 §1c does not forgive
+ '{"children":[{"blockImage":true,"children":[{"alt":"Apollo","src":"a.jpg","type":"image"}],"type":"paragraph"}],"type":"document"}'
- '{"children":[{"alt":"Apollo","src":"a.jpg","type":"image"}],"type":"document"}'
```

Both are the one cause the ruling names: three keys instead of two, so `isBareWrapper` stops recognising the wrapper.

Implementing the ruling now - widening §1c to admit `blockImage` and nothing else, in the normative clause as well as the checker, since PART 11 §1c states the bound in prose ("a wrapper carrying an attribute block is not BARE ... and neither is a node that owns a label, a destination, a level or a kind") and a checker-only change would leave the two disagreeing.

dereuromark:
Ruling implemented and pushed. `6c7f7e53`, rebased onto `c1ad63ec` (past #1828).

**The clause.** New normative sub-clause under PART 11 §1c, `CARVE-P11-039` - *a resolution result about the wrapper is not content of it*. It names `paragraph.blockImage` as the one field a wrapper may carry and still be BARE, and it refuses the generalization explicitly: not "ignore unknown keys". A key no clause has ruled on is content; `blockImage: false` is a spelling PART 12 §23 forbids; an attribute block beside the field is as disposable as one alone, which is to say not at all. `resources/spec/rules.json` and `resources/normative-clauses.txt` follow.

**Both §1c guards confirm the widened branch is reached.** Measured by reverting the predicate three ways rather than asserted:

| predicate | `node --test tests/corpus-fmt-roundtrip.test.mjs` |
| --- | --- |
| as shipped | **exit 0**, `pass 12 / fail 0` |
| reverted to a key count | **exit 1**, `fail 4` - sweep, carve-out guard, bound test, strong property |
| widened to "any key that is not `attrs`" | **exit 1**, `fail 1` - bound test |
| key admitted, value unchecked | **exit 1**, `fail 1` - bound test |

The carve-out guard is new and separate from the ceiling guard on purpose: the ceiling is satisfied by a paragraph holding one comment, so the metadata branch could stop being exercised by anything and that test would stay green. Over all 1527 documents and their 12027 canonical nodes, admitting the field decides 3 nodes differently, one in each `411` document, and moves those three from failing to passing - so unlike the three widenings already recorded above the bound test, this one is not a no-op.

**The rebase surfaced one thing.** `html-import:check` is not in `npm test`, and the pin move had left `resources/import-roundtrip-baseline.json` behind. Only `renderImportRoundTrips.markdown` moves, 397 -> 399, and it is four documents rather than two: `435-...-2`, `-4` and `-13` join, because the new pin reads them correctly, and `436-...-7` leaves, because the pin now reads `- +` over a flush-left line as a refused marker and the markdown target has no spelling for it. Every other count in the ratchet is unchanged.

**Gates**, all against the pushed commit:

```
npm test                    exit 0    3239 tests, 3238 pass, 0 fail
npm run html-import:check   exit 0    pass 7 / fail 0
npm run escape:check        exit 0    pass 5 / fail 0
npm run spec:check          exit 0
npm run spec:rules:check    exit 0    242 stable normative rule ids
npm run declarations:pr     exit 0    PASSED (mode: per-pr)
```

No corpus document, fixture or golden moved apart from the three `411` documents this fixes.

**Release mode is still 2 findings**, and one of them changed identity. `origin/main` is `PIN STALENESS BEHIND` (`bcb5cf4f`, 3 behind) plus 9 owed rows in `engine-pin-drift.txt`; this branch is `PIN STALENESS BEHIND` (`e0119621`, **1** behind) plus the 2 owed rows in `engine-fmt-drift.txt`. The one commit is carve-js `5c8c3dc`, test-only, which landed after the bump was measured - so the release verdict does not flip here after all, and the honest remaining blocker is unchanged: carve-js's writer still lags the #1821 ruling whose reader half ships at `e0119621`.

Relabeled `bug` + `area:spec` + `area:tooling`, since this is a normative change rather than pin plumbing. Merging once CI is green.

dereuromark:
Merged as `7a19ac63` while a review pass on the checker was still running, so two findings on it land in #1831 rather than here.

Both are about the predicate rather than the clause's substance. `blockImage` was admitted without naming the child it describes, so a paragraph carrying it around a *comment* - §1c's other shape - counted as bare; and the reachability guard asked whether a document PRODUCES such a wrapper rather than whether one NEEDS the carve-out to agree, which stays green on a dead carve-out. #1831 tightens both and re-pins the width with a fifth mutation row.

PR #1831, opened 2026-08-26
spec: the block-image carve-out describes an image, not any bare wrapper

dereuromark:
Follow-up to #1823, which landed while review feedback on it was still being worked. Two findings, both on the checker that encodes PART 11 §1c's new `CARVE-P11-039`.

## The carve-out did not name the child it describes

§1c reaches two shapes: a `paragraph` holding one `image`, and a `paragraph` holding one `comment`. The clause states the equivalence as *an image in block position* versus *a paragraph carrying `blockImage` around one `image`* - and the predicate checked the key and its value, then let the pre-existing child test stand. So

```
{ type: 'paragraph', blockImage: true, children: [{ type: 'comment', value: 'c' }] }
```

was BARE, and the checker would have forgiven losing both that wrapper and its field, for a shape no clause has interpreted. Nothing produces it today, which is why it belongs in the predicate rather than in a promise about producers.

All three halves are checked now, and each is a way the field could mean something no clause has ruled on:

- the **key**, because an unruled field is content
- the **value**, because `blockImage: false` is a spelling PART 12 §23 forbids
- the **child**, because a comment has no block-image fact to state

The clause says the same in prose.

## The reachability guard did not ask whether the carve-out was needed

`a corpus document reaches the PART 11 §1c metadata carve-out` asked only whether some document PRODUCES a wrapper carrying the field. That stays green in a world where the writer starts preserving those wrappers and an unrelated comment wrapper keeps the general ceiling reached - it would report a dead carve-out as live, which is the exact failure the guard exists to prevent.

It now runs the sweep's own comparison twice, once with the carve-out and once without, and names the documents that agree only WITH it. Renamed to match the question it asks: `a corpus document needs the PART 11 §1c metadata carve-out to agree`.

## The width, re-pinned

Each row measured by reverting that one thing, `node --test tests/corpus-fmt-roundtrip.test.mjs`:

| predicate | result |
| --- | --- |
| as shipped here | **exit 0**, `pass 12 / fail 0` |
| reverted to a key count | **exit 1**, `fail 4` - sweep, load-bearing guard, bound test, strong property |
| widened to "any key that is not `attrs`" | **exit 1**, `fail 1` - bound test |
| key admitted, value unchecked | **exit 1**, `fail 1` - bound test |
| key admitted, child unchecked | **exit 1**, `fail 1` - bound test |

The fourth row is what this PR adds; on `main` it passes, which is the defect.

No behavior change for any engine, and no corpus document, fixture or golden moves.

## The clause is condensed with it

It states what a bare wrapper IS - the key, the value and the child all holding - and no longer carries the paragraph explaining why the bound had to move. The rule belongs in the clause; how it arrived belongs here and in the commit message.

### DIFF
--- a/resources/grammar.ebnf
+++ b/resources/grammar.ebnf
@@ -8755,33 +8755,22 @@ EOF = (* end of file *) ;
 
       A RESOLUTION RESULT ABOUT THE WRAPPER IS NOT CONTENT OF IT -- NORMATIVE
       (carve#1823). `paragraph.blockImage` (PART 12 §23) is the ONE field a
-      wrapper may carry and still be BARE, and it is admitted BY NAME. It is
-      set by PART 9R R7's promotion phase and by nothing else, it is present
-      only as `true`, and what it records is that this paragraph IS the block
-      image -- which is exactly what the dissolved spelling states outright. So
-      the two spellings PART 12 §23 blesses for one fact, `image` in block
-      position and a `paragraph` carrying `blockImage` around one `image`,
-      differ in no information, and dissolving either into the other takes
-      nothing with it. Note that this is a rule about the CHECKER's comparison
-      and not about the writer: `fmt` still does not read the field, per
-      DECIDED ON THE SPELLING above.
+      wrapper may carry and still be BARE, and it is admitted BY NAME. §23
+      bounds it to `true` and PART 9R R7's promotion phase is its only
+      producer, so what it records is that this paragraph IS the block image --
+      which is what the dissolved spelling states outright. The two spellings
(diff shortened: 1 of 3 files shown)


dereuromark:
Pushed `cfb75d73`. CI is queued rather than running - the org's runner queue is backed up, four other branches are sitting in it too - so this is waiting on the matrix rather than on anything here.

What was verified locally is the part CI cannot do, the mutation battery in the body: each of the four ways to loosen the predicate turns a test red, and the fifth row (child unchecked) is the one that passes on `main` today.

PR #1832, opened 2026-08-26
spec: a colon followed by only whitespace is not a description

### LINKED ISSUE #1830: parser: four empty-description spellings diverge across the engines
While checking the six spellings `markup-carve/carve#1627` rejected for an empty description body, four of them turned out to disagree across engines. None of them yields an empty `<dd>` in any engine, so this does not affect the `markup-carve/carve#1827` ruling - but it is live cross-engine drift that nothing appears to declare.

Measured at carve-js `5c8c3dcb`, carve-php `67f7df9f`, carve-rs `1cfab2d6`, on this document:

```
:: t
<the spelling>

flush
```

The trailing `<p>flush</p>` is identical everywhere and is elided from the table.

| spelling | carve-js | carve-php | carve-rs |
| --- | --- | --- | --- |
| colon + 1 space | `<dl><dt>t</dt></dl><p>:</p>` | `<dl><dt>t :</dt></dl>` | `<dl><dt>t</dt></dl><p>:</p>` |
| colon + 2 spaces | `<dl><dt>t</dt></dl><p>:</p>` | `<dl><dt>t :</dt></dl>` | `<dl><dt>t</dt></dl><p>:</p>` |
| colon + 3 spaces | `<dl><dt>t</dt></dl><p>:</p>` | `<dl><dt>t :</dt></dl>` | `<dl><dt>t</dt></dl><p>:</p>` |
| colon + tab | `<dl><dt>t</dt></dl><p>:</p>` | `<dl><dt>t :</dt></dl>` | `<dl><dt>t :</dt></dl>` |
| colon + `{}` | `<dd>{}</dd>` | `<dd>{}</dd>` | `<dd>{}</dd>` |
| bare colon | `<dt>t :</dt>` | `<dt>t :</dt>` | `<dt>t :</dt>` |

So the three trailing-whitespace spellings split carve-php against the other two, and the tab spelling splits carve-js against the other two. The last two rows agree everywhere.

The executable spec at `c1ad63ec` answers the whitespace rows as carve-php does (`<dt>t : </dt>`) and the tab row as carve-php and carve-rs do.

This also corrects an inherited fact: `markup-carve/carve#1627` records colon-plus-three-spaces as rendering `<dd>&nbsp;</dd>`, and no engine does that today. That ticket's conclusion (none of the six spells an empty description) is unaffected.

I did NOT check whether any drift ledger already declares these, and that should be the first step - check `resources/engine-pin-drift.txt`, `resources/oracle-divergence.txt` and the AST divergence ledgers before treating this as new.

dereuromark:
Closes markup-carve/carve#1830.

## The rule

A description marker takes a separator space and non-empty content, so a `:` line carrying nothing but whitespace opens no description. It is a plain line under the open term, which folds it as a soft break, and the line's trailing whitespace run is dropped. `:`, `: `, `:  `, `:   ` and a tab-separated `:` are one document.

## What decides it

Three clauses, all already normative, none moved:

- **MARKER REQUIRES CONTENT** (PART 2) - *"A marker that takes a separator space opens its block only when followed by that space AND non-empty content [...] The rule ignores trailing whitespace"*, and *"EVERY SUCH MARKER, not only the two named here [...] applies wherever one appears"*. `definition_separator` is a separator space.
- **MARKER SEPARATORS AND PADDING SLOTS** (PART 1) - *"A MARKER SEPARATOR [...] is spelled `space`, and a tab never satisfies it."* The tab spelling is refused here, a step before the content test, and lands in the same place.
- **NO TRAILING WHITESPACE** (PART 2) - *"A `whitespace` run at the END of a CONTENT LINE is DROPPED [...] `abc<SP>` followed by `def`, where the run sits before a SOFT BREAK; and so do a heading, a list item, a block quote line, **a definition term or description**"*, over the terminal `' ' | '\t'`.

The one clause edit widens MARKER REQUIRES CONTENT's enumeration from the term marker `::` to the description marker `:` as well. It named `::` because that is the marker it was widened for; `:` is the marker two of the three engines read differently.

## Which reader was wrong

carve-php is correct on all six spellings. The executable spec was not: it classified the four rows correctly and then leaked the trailing run into the `<dt>`, so it was a fourth reading rather than agreement with php. Its term fold appended the folded line raw. A sweep of the ten fold sites in the oracle found this one alone, and it is not confined to the degenerate spellings - `:: a` / `b<SP>` / `c` leaked too, on an ordinary wrapped term. All three engines already drop it there.

## Fixtures

Corpus 438, six documents: the four whitespace spellings plus two controls - `: {}`, where the marker has content and the description forms, and the bare marker, whose empty trailing run leaves the fold nothing to drop. The controls are why the four read as one rule.

## Ports

carve-js emits a stray `<p>:</p>` on all four spellings (markup-carve/carve-js#1561); carve-rs does on the three space spellings and reads the tab correctly (markup-carve/carve-rs#1452), which is `markup-carve/carve-rs#518` already working.

Measured against the six fixtures: carve-php reproduces **6/6 byte-for-byte**, carve-js fails four, carve-rs fails three. The pin's window on the four is declared in `resources/engine-pin-drift.txt` and retires when the pin moves past the carve-js fix; the new category joins the declared-lag list in `docs/implementation-comparison.md`, so the published denominator stays at 768 rather than quoting a run nobody took.


dereuromark:
Rebased onto `2e71e711` now that #1833 and #1831 have landed, and took the two collisions this PR owns as the second one in.

**Category 439.** #1833's five `438-an-empty-description-body-is-written-with-the-empty-sentinel` documents are on `main`, so these six moved up. The section now sits after the sentinel's in `resources/examples/edge-cases.md` and the renumber ran through `CORPUS_RENUMBER=1 npm run corpus:build`, so the sidecar lock and the `.test` mirrors were regenerated rather than renamed by hand. The four `engine-pin-drift.txt` rows, `fixture-bytes.test.mjs` and the comparison page's declared-lag line follow the new name.

**`docs/html-import.md`.** All five whitespace-only and bare colon spellings fold the line into the term and reach no `<dd>` at all - fixtures 439 through 439-4 and 439-6 are byte-identical in output - so the passage states that as one row. `: {}` stays separate, because it does reach a `<dd>` holding literal `{}`.

Also picked up the counting gates for #1833's five documents, which were not routed: `resources/example-pages.txt`, the comparison page's declaration line, the landing-page count and `resources/import-roundtrip-baseline.json`. Those four gates are red on `main` as it stands, and are green here.

Actions is still in a major outage, so the evidence is local, on the pushed commit `6814a069` with a clean tree and a fresh `npm ci` under Node 24: `npm test`, `html-import:check`, `escape:check`, `spec:check`, `spec:rules:check`, `core:check` and `engine:report --check` all exit 0.

Issue #1886, opened 2026-08-31
chore: a shared fixture's declared mode is asserted but never applied


dereuromark:
A shared HTML-import fixture declares a `mode` in its `expected.report.json`, and every
fixture declares `safe`. That field looks like configuration and is not: no runner passes
it to the importer.

| runner | how it imports |
| --- | --- |
| spec `tests/html-import-contract.check.mjs` | `htmlToCarve(html)` |
| carve-js `test/html-import-conformance.test.ts` | `htmlToCarve(html)` |
| carve-rs `tests/html_import.rs` | `html_to_carve(&html, &HtmlImportOptions::default())` |
| carve-php shared-fixture path | `new HtmlToCarve()` |

Each imports with the default. The declared mode is asserted against the report the import
produced, and never used to produce it. So a fixture declaring `roundtrip` would be
imported in `safe` and fail in all four runners at once, which is why every fixture reads
`safe` and why the field has never been exercised.

Nothing is broken today. The cost is that the fixture format offers a knob that silently
does nothing, and the first person to reach for it gets four simultaneous failures with no
hint of the cause.

## What closing this looks like

Engines first, because the reverse order breaks every runner at once:

1. carve-js, carve-rs and carve-php each teach their shared-fixture runner to pass the
   fixture's declared mode. Harmless while every fixture says `safe`, and it changes no
   result.
2. The spec runner does the same.
3. Only then may a fixture declare a mode other than `safe`.

## What this is NOT blocking

It was written up as the blocker for `attribute-preserved` coverage in #1878. That was
wrong, and the reason is worth keeping: the gate has TWO oracles, and only the fixture one
ignores the mode. The corpus oracle already sweeps all three modes, so a corpus document
reaches the code with no harness change at all - which is how #1878 was actually closed.

So this is a fixture-format wart to tidy, not a coverage gap. If the answer turns out to be
that shared fixtures may only ever be `safe`, the right close is to drop the field from the
format rather than teach four runners to read it.

PR #1887, opened 2026-08-31
test: cover attribute-preserved with a corpus row, not a harness change

### LINKED ISSUE #1878: Two diagnostic codes have a producer and no shared case
Two diagnostic codes have producers and no shared case, so the new
two-directional gate in `tests/html-import-contract.check.mjs` (carve#1835)
carries them as exemptions rather than coverage.

Both are real, and both are narrow:

- `attribute-preserved` needs `roundtrip` mode AND an element kept whole as raw
  HTML that also carried a refused attribute. Every fixture under
  `tests/html-import/` declares `"mode": "safe"`, and the corpus sweep the gate
  runs reaches the code in no mode.
- `table-degraded` needs a table the importer cannot represent structurally.
  No corpus document and no fixture holds one.

Measured with the pinned build over all corpus documents in all three modes,
the vocabulary that actually appears is:

```
attribute-dropped: 222
element-dropped: 18
element-unwrapped: 373
raw-preserved: 14
structure-unspellable: 9
style-unmapped: 65
```

`encoding-assumed` comes from a fixture; `diagnostics-truncated` is unreachable
by construction here and stays exempt for that reason.

## What closing this looks like

A fixture per code, which is the coverage the exemption stands in for:

- For `table-degraded`, an input whose table the importer degrades, with the
  three expectation files beside it.
- For `attribute-preserved`, the open question first: whether a shared fixture
  may declare a mode other than `safe`. The engine-side runners read that field,
  so it is a contract question rather than a fixture-authoring one. If the
  answer is no, the exemption is permanent and its reason should say so instead
  of pointing here.

Neither is urgent. What matters is that the reasons stay accurate: an exemption
whose premise expires is a hole with a comment on it.

dereuromark:
Closes #1878. The `mode` wart that ticket described is filed separately as #1886 and blocks
nothing.

## The ticket asked for a four-repo sequencing. It needed one corpus row.

The last exemption read: `attribute-preserved` needs `roundtrip` mode, no fixture runner
passes one, so the harness must honor a declared mode before the coverage can exist.

That was measured from the wrong oracle. The gate has two, and only the FIXTURE oracle
imports with the default:

```
for (const mode of ['safe', 'semantic', 'roundtrip']) {
  for (const d of htmlToCarve(html, { mode }).report.diagnostics) codes.add(d.code)
}
```

The corpus oracle already sweeps all three. Coverage needs an input, not a capability -
the same thing that turned out to be true of `table-degraded` in #1882.

## The row

A raw block passes its payload through untouched, including an attribute the importer
refuses. Through the gate's own path:

| mode | codes |
| --- | --- |
| safe | `attribute-dropped` `element-unwrapped` |
| semantic | `attribute-dropped` `element-unwrapped` |
| roundtrip | **`attribute-preserved`** `raw-preserved` |

The example documents that as a fact about raw blocks rather than as a fixture
contrivance: the same attribute is reported dropped in one mode and preserved in the other,
which is the pair the code exists to name.

`NOT_COVERED_HERE` is now the single entry that is unreachable by construction, and the
comment above it no longer describes two codes with producers that nothing exercises.

## Counts

One document, so `corpusDocuments`, `completed`, `canonicalFixedPoints` and
`renderedTextPreserved` each move by one, and the landing page and comparison page follow.

`renderImportRoundTrips.html` is deliberately NOT touched. It is drifting on main
independently of this change - `663` measured against an expected `662` - and the same two
gate failures reproduce with this branch stashed. Someone is on that; this branch leaves
the line to them, so the ratchet diff here shows only theirs.

That also means this PR cannot go green until that lands. Everything it does own is green:
the fixture-bytes inventory, the corpus census, the docs-page routing and the comparison
counts.

The published `1541 / 1541` was bumped after checking the new row renders byte-identically
in carve-rs, carve-js and carve-php, rather than assuming a new document passes everywhere.
`npm run compare:counts` could not regenerate them here - it finds no runnable engine in
this environment and skips all three.

### DIFF
--- a/docs/implementation-comparison.md
+++ b/docs/implementation-comparison.md
@@ -21,15 +21,15 @@ implementation exposes.
 
 <div class="impl-summary-grid">
   <div class="impl-summary-card">
-    <strong>1540 / 1540</strong>
+    <strong>1541 / 1541</strong>
     <span>Rust corpus pass</span>
   </div>
   <div class="impl-summary-card">
-    <strong>1540 / 1540</strong>
+    <strong>1541 / 1541</strong>
     <span>JS corpus pass</span>
   </div>
   <div class="impl-summary-card">
-    <strong>1540 / 1540</strong>
+    <strong>1541 / 1541</strong>
     <span>PHP corpus pass</span>
   </div>
   <div class="impl-summary-card">
@@ -40,9 +40,9 @@ implementation exposes.
 
 | Implementation | Commit | Corpus | Mismatches | Errors | Avg CLI ms/file |
 |----------------|--------|--------|------------|--------|-----------------|
-| Rust | `da45f9d2` | `1540 / 1540` | `0` | `0` | `3.71` |
-| JS | `f0abfc66` | `1540 / 1540` | `0` | `0` | `107.12` |
-| PHP | `3a39d658` | `1540 / 1540` | `0` | `0` | `75.37` |
+| Rust | `da45f9d2` | `1541 / 1541` | `0` | `0` | `3.71` |
+| JS | `f0abfc66` | `1541 / 1541` | `0` | `0` | `107.12` |
+| PHP | `3a39d658` | `1541 / 1541` | `0` | `0` | `75.37` |
 
 Spec commit: `3eae6be`. The run covers the whole corpus, so nothing is declared
 after it.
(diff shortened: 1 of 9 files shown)


dereuromark:
Correction to this PR's own description, for anyone reading it later.

It said `renderImportRoundTrips.html` was drifting on main, that the same two gate failures
reproduced with the branch stashed, and that this could not go green until someone else
fixed main. All three are wrong. Main was never red.

The failures came from my checkout, not the repo: its `node_modules` were two days old, so
the gate ran against an older pinned engine build, which measured `662` where the pinned
build measures `663`. A clean clone of main passes 10/10, and so does this branch on the
same clone.

The bump in `resources/import-roundtrip-baseline.json` is unaffected and was right for the
opposite reason to the one given: `html` stays at `663` because the correct engine measures
`663` and this row adds no HTML round trip, not because the line belonged to someone else.

The verification method holds - the row was checked against carve-rs, carve-js and
carve-php before the published counts moved. What failed was reading a stale environment as
a fact about the repo.

Issue #1965, opened 2026-09-07
AST span/value divergences after the 1b27b68 corpus bump


dereuromark:
The 1b27b68 corpus bump exposed cross-engine AST **span/value** divergences that no ledger covered. `npm run ast:check` (PART 12 conformance) fails on them until each is fixed or declared in `resources/ast-span-divergence.txt` / `resources/ast-value-divergence.txt`. This issue tracks the residual set — the ones that are legitimate cross-engine differences rather than a single engine's §4 violation (no engine is flagged by the source-side `checkOpeningMarkup` / `checkStopsAtChildren`).

Measured over the three engine mains plus the two comment-span fixes (carve-rs#1568, carve-php#1921) and assuming carve-js's separate `endOffset past end of source` fix on the `455` family lands. The `455` rows (`list`, `list_item`, `code_block`, and the 455 half of `definition_list` / `definition_description`) are that carve-js bug and clear with it — they are not declared here.

### Span residuals (`ast-span-divergence.txt`)

- `definition_description (extent)` and `definition_list (extent)` — `447-...-13`. The `definition_description` has NO placed children; its footnote and `link_reference_definition` are hoisted document-level siblings. carve-js and carve-rs span the description over them (offset 27); carve-php stops earlier (offset 14). Per carve#1522 a hoisted definition is a sibling, so the wider span is the one in question — carve-js/carve-rs to narrow.
- `footnote (extent)` — `456-...`. carve-js starts the hoisted footnote definition one codepoint into the indentation (offset 30, column 4); carve-rs and carve-php start on the `[` marker (offset 31, column 5). `footnote` is absent from `OPENING_MARKUP`, so the source-side rule does not name a side — carve-js to start on the marker.
- `block_quote (extent)` — `458-...`. carve-php has not ported carve#1962 (a wrapped attribute block does not reach past a quote's closing boundary): it spans the quote to offset 26 over the trailing lines; carve-js and carve-rs end it at offset 5. carve-php to port carve#1962.

### Value residuals (`ast-value-divergence.txt`)

- `paragraph.attrs.classes`, `paragraph.attrs.id`, `paragraph.attrs.order` — `458-...`, the same carve-php carve#1962 lag: carve-php attaches the wrapped attribute block to a paragraph the other two engines never build.

Each row is one document; the counts are the residual after the `455` fix.

dereuromark:
[...]

PR #1966, opened 2026-09-07
Declare residual AST span/value divergences after the 1b27b68 bump

dereuromark:
## What

Declares the residual cross-engine AST span/value divergences the `1b27b68` corpus bump exposed, so `npm run ast:check` (PART 12 conformance, the carve#1963 gate) can reconcile them instead of failing on undeclared rows.

The `comment` findings #1963 names are **fixed in the engines**, not declared here:
- carve-rs: markup-carve/carve-rs#1568
- carve-php: markup-carve/carve-php#1921

## What is declared

Four span rows and three value fields, across three documents. None is flagged by the source-side PART 12 §4 checker (`checkOpeningMarkup` / `checkStopsAtChildren`), so none is provably one engine's violation — they are legitimate cross-engine differences. Per-document detail and suspected owners: markup-carve/carve#1965.

- `definition_description` / `definition_list (extent)` — `447-…-13`: js and rs span a description over its hoisted footnote and link-ref-def siblings; php stops earlier (carve#1522).
- `footnote (extent)` — `456-…`: js starts the hoisted note one codepoint into the indent; rs and php on the `[` marker. `footnote` is absent from `OPENING_MARKUP`, so no side is source-named.
- `block_quote (extent)` + `paragraph.attrs.{classes,id,order}` — `458-…`: carve-php has not ported carve#1962, so it spans the quote past its closing boundary and attaches an attribute block to a paragraph the other two engines never build.

## Dependency (read before merging)

The counts are the **residual after the separate carve-js `endOffset past end of source` fix on the `455` family**. That carve-js bug currently inflates `definition_description`/`definition_list` and adds `list`/`list_item`/`code_block` rows; those clear with the carve-js fix and are deliberately **not** declared here. Reconciled with the checker's own `reconcileSpans` / `reconcileDeclared`:

- against today's engine mains: `definition_description`/`definition_list` report COUNT and `list`/`list_item`/`code_block` report NEW (all the carve-js `455` bug); `footnote`, `block_quote` and the three value fields already reconcile clean.
- against the engines with the carve-js `455` fix applied: **both files reconcile clean.**

So this merges green only once carve-rs#1568, carve-php#1921 and the carve-js `455` fix have landed. It is the convergence piece of #1963.

Refs markup-carve/carve#1963, markup-carve/carve#1965.

### DIFF
--- a/resources/ast-span-divergence.txt
+++ b/resources/ast-span-divergence.txt
@@ -687,3 +687,36 @@
 # carve-rs#1308, and the same run reports BINDING PARITY differing from carve-rs
 # on the four `410` documents. That is a stale `ext/carve/Cargo.toml` pin in
 # carve-rb, not a span this file could declare.
+
+# RE-MEASURED 2026-09-07 for the carve#1963 release gate, at carve-js 9e5319f2,
+# carve-rs (markup-carve/carve-rs#1568) and carve-php (markup-carve/carve-php#1921)
+# over 1685 corpus documents plus 3 synthetic samples. The comment-span findings
+# #1963 names are FIXED in those two engine PRs, not declared. What remains is
+# four rows across three documents, each a legitimate cross-engine difference the
+# bump exposed - none flagged by the source-side PART 12 SS4 checker
+# (checkOpeningMarkup / checkStopsAtChildren), so none is provably one engine's
+# violation. Named per document and suspected owner in markup-carve/carve#1965.
+#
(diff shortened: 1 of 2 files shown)


dereuromark:
Closing as obsolete. The residual span/value divergences this PR would declare were instead fixed in the engines (carve-rs#1568/#1570, carve-php#1921, carve-js#1661) and the carve-rb pin bumped to match, so all three engines are byte-identical on the current corpus. Both ledgers now carry zero declaration rows, and #1968 trims their stale commentary. There is nothing left to declare.

Issue #1974, opened 2026-09-11
Where does the trailing line go once a description-hosted note takes the block below it?


dereuromark:
A note written into a description body, with a block opener at or past the note's own floor, is read three different ways. Found while porting carve#1971 to carve-php; it is a separate family and no engine agrees with another.

Input:

```
:: t
:  [^f]: note
     - nested
tail
```

The `dd`'s content column is 3, so `[^f]` stands at column 3 and PART 9 §16 puts its body floor at 5. The list opener sits at 5, so it reaches the note body. `[^f]` is never referenced, so a note that takes the list swallows it.

carve-js `6b050a68`:

```html
<dl>
  <dt>t</dt>
  <dd>tail</dd>
</dl>
```

carve-rs `ac7383d3`:

```html
<dl>
  <dt>t</dt>
  <dd>
    <ul>
      <li>nested
tail</li>
    </ul>
  </dd>
</dl>
```

carve-php `23dddc92`:

```html
<dl>
  <dt>t</dt>
  <dd>
    <ul>
      <li>nested</li>
    </ul>
  </dd>
</dl>
<p>tail</p>
```

Two questions are tangled here and they should be ruled separately.

**Does the list opener reach the note body?** js says yes, rs and php say no. The floor rule reads as yes: the opener is at the note's marker column plus two. rs applies the per-marker floor for a note inside another note (carve-rs#1575) but not for one inside a `dd`; php's own suite already asserts the floor for plain continuation text in a `dd` (`ADefinitionDescriptionDedentsByItsOwnContentColumnTest`), so it is inconsistent with itself between a continuation line and an opener.

**Where does `tail` go?** At column 0 it is at or below the `dd`'s base column, so PART 0's owner-selection table gives it to the nearest surviving ancestor - the document, which is php's answer. But at column 3 all three engines lazily continue the open paragraph into the `li`, so lazy continuation does cross the boundary, which is rs's reading. js puts it in the `dd`, matching neither.

For the record, the same shape with ordinary continuation text instead of a list opener is settled and agreed, and the nested-in-a-note band from carve#1971 is now byte-identical across all three engines over 797 grid cells. This ticket is only the opener-in-a-description-body corner.

dereuromark:
Ruled, both bands, matching djot and markdown.

[...]

PR #1981, opened 2026-09-12
spec: a description-hosted note's floor is its own marker plus two

dereuromark:
Pins the goldens markup-carve/carve#1974's ruling names, and moves the oracle's Q2 answer with the engines.

Section 461 holds the four bands. The description's content column is 3, so `[^f]` stands at 3 and its floor is 5:

- a block opener at 5 is the note's, and the column-0 trailing line is a top-level sibling;
- plain continuation at 5 is the same line - the floor is a column, and nothing turns on whether the line opens a block;
- one column shy, at 4, the note reaches nothing and the list is an ordinary description child;
- the trailing line at the description's own content column stays in the description, which is what makes the column-0 row a statement about the column rather than about the note.

Every row is byte-identical across carve-js `df49fafa`, carve-php `d3269e08` and carve-rs with markup-carve/carve-rs#1580, each built from source.

### The oracle

It answered the trailing line wrong in the two rows where the note took a line, and right everywhere else - including a description holding nothing but the definition. The reason is the question it asks: whether the description body leaves a paragraph open. The paragraph belonged to the NOTE, which is extracted before the description renders.

So the lines the note's floor claims are cut before the question is asked. The definition LINE stays: an empty body deliberately answers the other way (the `:  ` plus pulled-block shape), so cutting it too would fold the line right back in. Columns 1 and 2 are untouched - that band is above the description's base column and PART 9 section 10 I5 calls it lazy paragraph text wherever the paragraph came from.

Both new rows are declared in the pin-drift ledger. The pinned build writes the trailing line back into the description, so the formatter's output reads as a different document until the pin moves past the carve-js fix.


dereuromark:
Superseded by #1979, which landed the same oracle fix and the same four bands while this was being built. Closing rather than rebasing - the merged version covers everything here, including both pin-drift declarations.

One thing worth carrying over: #1980 (conformance red on main) is caused by these rows, not by them being wrong. carve-rs main reproduces three of the four and reads `:: t` / `:  [^f]: note` / five-space `cont` / `tail` the other way, because carve-rs#1577 gathered a block opener at the hosted note's floor but kept a carve-out for plain continuation at the same column. markup-carve/carve-rs#1580 removes the carve-out, and with it built from source all four rows match byte for byte.

Issue #2027, opened 2026-09-15
carve-js and carve-php disagree about where a bare tilde strike closes inside a braced inline


dereuromark:
carve-js and carve-php disagree about where a BARE tilde strike closes when its
content contains a braced inline. Found while building the seam matrix for
markup-carve/carve-js#1706 and markup-carve/carve-php#1974; it is a parser
question, not a writer one, and I have no reading of the spec that settles it.

Input

````
~{/x/}{/y~/}
````

carve-js HTML

````html
<p><s><em>x</em>{/y</s>/}</p>
````

carve-php HTML

````html
<p>~<em>x</em><em>y~</em></p>
````

carve-js opens a bare strike at the leading `~` and closes it at the `~` INSIDE
the second braced inline's content, which leaves `{/y` inside the strike and `/}`
as literal text after it. carve-php does not open a bare strike there at all, so
the leading `~` is text and both braced inlines parse as written.

The two engines agree on every simpler bare-set form I tried, so this is narrow:

````
~a~
~a b~
a ~b~ c
````

all give `<s>...</s>` in both.

## Which is right

I do not know, and that is the ticket. Either a bare set marker may close inside
a braced inline's content (carve-js) or it may not (carve-php), and the spec
clause for the bare set forms is what has to say so. A ruling either way makes
one engine's behavior a bug with a corpus row to pin it.

## Measured

1265 of 27024 rows of an otherwise byte-identical renderer sweep diverged, and
every one of them was a context that opened with a bare `~`. Dropping that
context brought the two engines back to byte-identical on all 25024 remaining
rows, before and after the run-merge fix. So this is the only shape in that
matrix where the two parsers disagree.

dereuromark:
Third reading, from carve-rs at `30c60633`. **It agrees with carve-js, byte for byte.**

````html
<p><s><em>x</em>{/y</s>/}</p>
````

Same on the simpler bare-set forms both engines already agree on (`~a~`, `~a b~`, `a ~b~ c`), so carve-rs is not diverging somewhere else and landing here by accident.

Stating the org rule before anyone reads a majority into this: two-against-one settles nothing. What a third reading is good for is showing which RULE each engine is implementing, so here is what carve-rs's is, measured:

| input | carve-rs |
| --- | --- |
| `~{/x~/}` | `<p><s>{/x</s>/}</p>` |
| `~{=x~=}` | `<p><s>{=x</s>=}</p>` |
[...]

PR #2031, opened 2026-09-15
spec: a bare closer does not reach inside a braced inline

dereuromark:
Addresses the spec side of #2027. The engine fixes in markup-carve/carve-js and markup-carve/carve-rs are separate work, so this PR does not close the ticket.

Following the ruling on #2027, PART 9 §9 gains **E2a**: code spans and braced inlines are both opaque to a bare delimiter. A bare closer inside braces can only close an entry opened inside those same braces. The clause names both kinds as one rule instead of adding braces to the code-span exclusion E1 already had. A construct added to either kind later is then opaque because of what it is.

Input:

```carve
~{/x/}{/y~/}
```

Output:

```html
<p>~<em>x</em><em>y~</em></p>
```

**How far the rule goes, measured.** I rendered through carve-php `origin/main` 5ff2c6ec. It applies the rule to all five bare delimiters and to every member of the braced family: forced spans, `{+ +}`, `{- -}`, `{~ ~> ~}`, `{# #}` and `{% %}`. Link labels and bracketed spans are **not** opaque in carve-php or carve-js (`~[a~](b)` gives `<s>[a</s>](b)` in both). E2a does not claim them, because that would go beyond what the ruling decided.

**The executable Core grammar already agrees.** `core:check` reproduces all three new documents. Braced spans are single PEG tokens in `carve-core.ohm`, and their content is resolved on its own stack.

**Corpus section 463** has three documents:

- the forced-span case from the ticket
- an addition, where the outer strike still closes after the braces
- `*` with a forced underline, for a second bare delimiter

The pinned carve-js 0.1.6 reaches inside the braces on all three, so they are declared in `resources/engine-pin-drift.txt`. They are also listed as added since the published comparison run. The import baseline moves by three on every axis except the Markdown round trip. That stays at 413 because the pinned build mis-parses all three, so its Markdown writer spells a strike or strong that crosses the braces and the text does not survive the round trip. When the pin moves past the fix, the three drift lines go away and that count should be looked at again.

**Mutation checks.** Each new row was tested on its own:

- Set the fixture to carve-js's output: `core:check` reports a defect and `engine:report` reports the drift line as stale.
- Broke the Ohm grammar three times, once per row, so the braced production refuses the relevant bare delimiter. Each break turned exactly its target row red. The file was restored and the md5 checked after each.

**Found on the way, not ruled, not pinned:** `~a{#x~}b~` differs between the engines. carve-php gives `<s>a{#x~}b</s>` (with `#x` as a tag), and carve-js closes at the `~` before `}`. `{#x~}` is not a braced inline, so E2a does not decide it.


dereuromark:
Expected is `<p>~<em>x</em><em>y~</em></p>`

PR #2063, opened 2026-09-16
chore: move the carve-js pin past the empty code span fix and drop its drift line

### LINKED ISSUE #2062: The carve-js pin is a squash-merged branch head, and it keeps four stale writer-drift lines alive
`package.json` pins `@markup-carve/carve` to `d81c027873ad90fb592534774e1be109ff47f6ec`. That is the head of markup-carve/carve-js#1671, which was squash-merged on 2026-09-14, so the commit is reachable from no carve-js ref: the compare API reports it 6 commits ahead of carve-js `main` and 65 behind. The lockfile resolves it by SHA, so CI stays green on a build that no branch contains.

It also holds `resources/engine-fmt-drift.txt` in place. All four lines there describe writer defects carve-js has since fixed on `main`, and the staleness ratchet in `tests/corpus-fmt-roundtrip.test.mjs` cannot say so while the pin sits behind the fixes:

- `12-inline-code-7`, fixed by markup-carve/carve-js#1778 (the fix for the closed markup-carve/carve-js#1773)
- the three `464-a-block-that-opens-a-tight-item-is-written-on-the-marker-line` rows, whose reason names the marker-column sentinel leak that markup-carve/carve-js#1686 strips

Moving the pin to carve-js `main` and deleting the lines the ratchet then names fixes both.

dereuromark:
Closes #2062

`resources/engine-fmt-drift.txt` still listed `12-inline-code-7` against markup-carve/carve-js#1773, which is closed. The line was still load-bearing: the pin #2055 set, `59ae4a60`, predates markup-carve/carve-js#1778, the fix.

The pin now names carve-js `main` at `75734e8`, which contains #1778 and which the compare API reports `identical` to `main`. At that pin the staleness ratchet in `tests/corpus-fmt-roundtrip.test.mjs` reports `12-inline-code-7` as round-tripping clean, and putting the line back turns it red. The line goes. It was the last one in the file, and nothing else there named an engine ticket.

Nothing else moved in the pin-consuming tests run before pushing: the corpus and its pin-drift ledger, both writer round-trip tests, importer fidelity, the converter corpus, the examples, both import exits, the lint rule table, and the optional corpus with its feature adapters.

This PR first also fixed the dangling `d81c0278` pin #2062 was filed about and the three `464` drift lines. #2055 merged the same fixes while this was in CI, so the branch was rebuilt on top of it with only what was left.

The pin also moves the import baseline by one on each axis, measured per document against a build of `59ae4a60` that reproduces the committed counts:

| axis | change | document | cause |
| --- | --- | --- | --- |
| HTML | 702 to 703 | `12-inline-code-7` gains | markup-carve/carve-js#1778 |
| Markdown | 419 to 418 | `84-single-line-headings-10` loses | the writer now escapes the heading's trailing hash run as #2056 requires, matching the `.md` golden; the importer keeps that backslash in Carve, so bytes differ while both render `b ###` |

A writer-side repair: the old Markdown pass came from output a CommonMark reader takes as the heading `b`.


dereuromark:
Merging at carve-js `75734e8`. markup-carve/carve-js#1786 is still open and its fix has not landed, so this pin will move again once it does; that bump re-runs the same ratchets.

PR #2071, opened 2026-09-16
spec: an inline wrapper reaches the §1c ceiling, and a hard break flattens under §1b

### LINKED ISSUE #2066: A same-kind emphasis nested in a braced one has no spelling when both levels need braces
An emphasis nested directly inside an emphasis of the same kind has no Carve spelling when both levels need the braced form. All three engines write the same bytes for it, and those bytes read back as a different tree.

Found by the 1560-shape sweep in markup-carve/carve-php#2054, which left this one row red.

## The shape

HTML:

```html
<p>a<strong><b>x<br></b></strong>b</p>
```

carve-rs (`efbf04f`), carve-php (`6d65667`) and carve-js (`7d26c8a`) all import it as:

```
a{*{*x\
*}*}b
```

All three read that back as:

```html
<p>a<strong>{*x<br>
</strong>*}b</p>
```

Each engine's writer, handed the tree through its AST ingest, writes the same bytes, so this is not only an importer bug.

The cause is PART 9 §9 E3 (NO SAME-TYPE NESTING): while a strong is open, a further `{*` does not push and stays literal. Measured, a same-kind nesting survives only when one of the two levels is bare. Here neither level can be bare:

- the outer strong sits between word characters, so it needs `{*`;
- the inner strong's content ends in a hard break, so its closer starts a line and needs `*}` (markup-carve/carve-php#2047).

The break is one way to reach that. The always-braced kinds reach it with no break at all:

| HTML | all three engines write | reads back as |
| --- | --- | --- |
| `<p><sup><sup>x</sup></sup></p>` | `{^{^x^}^}` | `<p><sup>{^x</sup>^}</p>` |
| `<p><ins><ins>x</ins></ins></p>` | `{+{+x+}+}` | `<p><ins>{+x</ins>+}</p>` |
| `<p>a<em><i>x<br></i></em>b</p>` | `a{/{/x\` + newline + `/}/}b` | `<p>a<em>{/x<br>` + newline + `</em>/}b</p>` |
| `<p>a<strong><b>x<br></b>y</strong>b</p>` | `a{*{*x\` + newline + `*}y*}b` | `<p>a<strong>{*x<br>` + newline + `</strong>y*}b</p>` |

Where one level can be bare, the engines already write a spelling that round-trips: `a{**x**}b` without the break, and `*{*x\` + newline + `*}*` without the word characters. `<s><del>` is two kinds and is unaffected.

## Spellings tried

Each read back with all three engines on the commits above. None gives `a<strong><strong>x<br></strong></strong>b`:

| source | read back |
| --- | --- |
| `a{*{*x\` + newline + `*}*}b` | inner `{*` is text; the first `*}` closes the outer strong |
| `a{**x\` + newline + `**}b` | inner `*` opens nothing: its closer starts the line |
| `a{*{*x\` + newline + `**}b` | inner `{*` is text, and so is the `*` before the outer closer |
| `a{**x\` + newline + `*}*}b` | inner `*` is text, the trailing `*}` is text |
| `a{*{*x\` + newline + `*}{%%}*}b` | a comment between the closers changes nothing |
| `a{*{*x\` + newline + ` *}*}b` | an indented closer changes nothing |
| `a{* *x\` + newline + `* *}b` | the padded inner pair is plain text |
| `a *{*x\` + newline + `*}* b` | round-trips, but adds two spaces the tree does not have |
| `a{*{*x*}*}b` (control, no break) | fails the same way, so the braced pair never nests |

So the failure is the braced pair `{X{X`, not the break.

### A. Declare it unspellable

No grammar change. The writer throws its unspellable error for the tree, the way markup-carve/carve-php#2044 ruled for an empty code span with content after it. The HTML importer, which has a diagnostic channel, unwraps the inner span and reports the row (`structure-unspellable` or `element-unwrapped`), as markup-carve/carve-php#2048 did for that span. For the five bare kinds the unwrap changes nothing a reader sees. For `<sup><sup>` it loses the second raise, and the row says so.

### B. Add a spelling

Let a forced opener directly inside a forced span of the same kind nest, so `{*{*x*}*}` is two strongs. This changes E3, which states the opposite in normative text. No corpus document spells `{X{X` today, so nothing pinned would move, but every engine's parser changes and the reading of existing documents changes with it.

### C. Normalize same-kind nesting

Add an N3 to PART 11 §10k so a same-kind nesting compares equal to one span, and let writers and importers merge it. §10k does not fit this well:

- it is a list for the round-trip COMPARISON, not a license to change the tree, and §1a forbids "dropping content";
- it already says "EQUAL strengths are not on the list";
- N2 only covers a child that spans the whole parent, and the last table row has `y` beside the inner strong;
- `<sup><sup>x</sup></sup>` renders a second raise, so merging it is not the same document.

## Recommendation

A. The tree reaches a writer only from an importer or a hand-built AST, which is the case markup-carve/carve-php#2044 ruled on: throw rather than write bytes that read back as something else. The importer has a channel, so under PART 11 §1c's approach it drops the inner wrapper and declares the loss instead of emitting a broken document. B would reverse a normative rule for a shape no author has written in source, and C would stretch a closed comparison list into a writer license and still leave sup, sub and the non-spanning child unsolved.

If A is chosen, the clause to touch is §1c's scope: it declares a ceiling only for a block wrapper today, and this is an inline one.

### LINKED ISSUE #2067: A hard break in a table cell has no spelling, and two engines lose the table
A `<br>` inside a table cell has no Carve spelling, and the engines handle it three different ways. Two of them lose the whole table.

HTML:

```html
<table><tr><td>x<br>y</td></tr></table>
```

carve-rs (`8ee97fc`) and carve-js (`7d26c8a`) import it as:

```
| x\
y |
```

That reads back as a paragraph, not a table:

```html
<p>| x<br>
y |</p>
```

carve-php (`6d65667`) imports it as `| x\ y |`, which keeps the table but turns the break into a no-break space (`<td>x&nbsp;y</td>`). Neither the break nor the change is reported.

All three writers, handed the tree through AST ingest, write the table-breaking form, carve-php's included. So only carve-php's importer avoids it. No clause in PART 11 or the HTML import contract says what a hard break in a cell becomes. PART 11 §1b (a flatten preserves the boundary it dissolves) looks like the closest rule.

Found while measuring markup-carve/carve-rs#1706.

carve-js tracks its side as markup-carve/carve-js#1797.

## No spelling exists

A cell is one line (`cell_content = inline_content`), and a `+` continuation row joins its column with a single space (PART 9 §5 T6), so neither route carries a break. Every answer is a loss, and the engines should take the same one.

## Options

A. Flatten the break to one space (U+0020). The HTML importer reports the loss (`structure-unspellable`); `render_carve`, with no channel, writes the space.

B. Keep carve-php's `\ `. The table survives, but the break becomes a no-break space, a character neither the tree nor the HTML had.

C. Refuse the tree in the writer, the way markup-carve/carve-php#2044 ruled for an empty code span, and let the importer drop the break with a report.

## Recommendation

A. PART 11 §1b already covers an inline-only slot, naming the table cell, that is handed something it cannot carry: the producer flattens, the canonical separator is one space, it "MUST NOT" invent a character, and a producer with diagnostics SHOULD report the loss. A hard break is not a block, so the clause does not reach it as written, but its reasoning does. B invents the no-break space §1b rules out. C drops a boundary §1b says a flatten must keep, and it would refuse any hand-built table that holds a break.

dereuromark:
Closes #2066
Closes #2067

Normative text, so this waits for maintainer approval. Do not merge.

## Clause text

Added to PART 11 §1b (CARVE-P11-003), after "WHAT IS NOT REQUIRED":

```
A HARD BREAK IN SUCH A SLOT IS FLATTENED THE SAME WAY. No spelling
carries a hard break in a single-line slot, so the producer writes one
space in its place, or nothing where the slot holds no token on one side
of it: `<td>x<br>y</td>` writes `| x y |` and `<td>x<br></td>` writes
`| x |`. A producer with a diagnostic channel reports each break as
`structure-unspellable`.
```

Added to PART 11 §1c (CARVE-P11-004), after "STATED OVER WHAT THE SHAPE SPELLS". It now covers any depth, per the follow-up ruling on #2066:

```
AN INLINE WRAPPER REACHES THE SAME CEILING. A span nested at any depth
in a span of the same kind has no spelling where both levels need the
braced form, because PART 9 §9 E3 leaves the inner opener literal while
the outer span is open: the inner `{/` in `{/a *b {/c/}*/}` is text. A
writer refuses that tree; an importer unwraps the inner level and
reports `structure-unspellable`, so `<sup><sup>x</sup></sup>` imports as
`{^x^}`.
```

No rule id is added or renumbered. `docs/rules/canonical-writing.md` moves only by line numbers.

The executable grammar does not follow E3 for forced spans: it reads `{/a *b {/c/}*/}` with the inner emphasis nested. carve-js reads it as the clause says, and the carve-php lane measured the same for its own indirect shapes. That is filed as #2078 and does not touch the fixtures here, which only the engines consume.

## Fixtures

`same-kind-strong-nesting`:

```html
<p>a<strong><b>x<br></b></strong>b</p>
```

```
a{*x\
*}b
```

`same-kind-superscript-nesting`:

```html
<p><sup><sup>x</sup></sup></p>
```

```
{^x^}
```

`same-kind-indirect-nesting`, an emphasis inside a strong inside an emphasis, where every level sits between word characters:

```html
<p>a<em>b<strong>c<em>d</em></strong></em>e</p>
```

```
a{/b{*cd*}/}e
```

`table-cell-hard-break`:

```html
<table><tr><td>x<br>y</td></tr><tr><td>x<br></td></tr></table>
```

```
| x y |
| x |
```

Each same-kind fixture reports one `structure-unspellable` row at the inner level, and the AST exit keeps every level and both breaks.

## Engine drift

None left. The engine fixes have merged: markup-carve/carve-js#1805 and markup-carve/carve-js#1806, markup-carve/carve-rs#1724 and markup-carve/carve-rs#1727, markup-carve/carve-php#2078 and markup-carve/carve-php#2080. #2076 moved the carve-js pin past them, so the `PIN_LAG` entries and the converter-drift lines are gone. The new indirect fixture passes against the pinned carve-js; carve-rs and carve-php main are checked by a dispatched AST conformance run on this branch.


dereuromark:
carve-php tickets for these rulings: markup-carve/carve-php#2070 (a hard break in a table cell, carve#2067) and markup-carve/carve-php#2071 (same-kind nesting, carve#2066).

dereuromark:
carve-rs: the #2066 half landed in markup-carve/carve-rs#1727 (`72092c3`), and the #2067 half in markup-carve/carve-rs#1724. On carve-rs main both `same-kind-strong-nesting` and `same-kind-superscript-nesting` write the expected `.crv` bytes and the same `structure-unspellable` rows, and `table-cell-hard-break` writes `| x y |` and `| x |` with both rows.

dereuromark:
carve-js main at `d3312de` (after markup-carve/carve-js#1805 and markup-carve/carve-js#1806) produces all three fixtures from this branch: `table-cell-hard-break`, `same-kind-strong-nesting` and `same-kind-superscript-nesting` match on the written Carve, the report codes and paths, and the AST.

dereuromark:
carve-php: markup-carve/carve-php#2078 merged (`080bf14`), fixing markup-carve/carve-php#2070. It writes the `table-cell-hard-break` bytes, report rows and tree, checked by running the shared fixture test with this branch's fixture directory in place. The `php/html-import--table-cell-hard-break` drift line can go once the pin moves.

dereuromark:
carve-php: markup-carve/carve-php#2080 implements the §1c inline-wrapper ceiling and passes `same-kind-strong-nesting` and `same-kind-superscript-nesting` locally. It is not merged yet.

One point for the clause: it says "nested directly", but E3 keeps the inner opener literal while any span of that kind is open. carve-php's parser reads `a{*{/{*x*}/}*}b` as `a<strong>{/{*x</strong>/}*}b` and `{^/{^b^}/^}` as `<sup>/{^b</sup>/^}`, so markup-carve/carve-php#2080 applies the rule at any depth. No fixture covers that case yet.

dereuromark:
carve-php: markup-carve/carve-php#2080 merged (`2500828`), fixing markup-carve/carve-php#2071. With `2500828` and `080bf14`, carve-php now writes all three fixtures on this branch (`same-kind-strong-nesting`, `same-kind-superscript-nesting`, `table-cell-hard-break`): bytes, report and tree. The three `php/` drift lines can go once the pin moves.

PR #2072, opened 2026-09-16
corpus: pin the escaped colon before a bracket-opening node

### LINKED ISSUE #2068: A text node ending in :name joins the next link or span into an inline extension, in all three engines
Text that ends in `:name` before a link or a span is written bare, and the two join into an inline extension. All three engines write the same bytes.

HTML:

```html
<p>x :name<span class="k">n</span></p>
```

carve-rs (`8ee97fc`), carve-php (`6d65667`) and carve-js (`7d26c8a`) import it as

```
x :name[n]{.k}
```

which reads back as

```html
<p>x <span class="ext-name k">n</span></p>
```

A link does the same: `<p>x :name<a href="u">n</a></p>` imports as `x :name[n](u)`, which reads back as an `ext-name` span followed by the text `(u)`.

This is the colon-sigil form of markup-carve/carve-rs#1710, where a caret ending a text node joined the next node's `[` into an inline note. It turned up while checking that fix for other trailing characters that combine with a node's first character.

dereuromark:
Closes #2068

A new HTML import fixture, `extension-sigil-escape`, pins the bytes settled on the ticket.

```html
<p>x :name<span class="k">n</span> and :link<a href="u">n</a></p>
```

```
x \:name[n]{.k} and \:link[n](u)
```

The written form reads back as the input and is a fixed point of the pinned writer. With the escapes removed, `the-two-import-exits-agree` goes red on this fixture: five nodes parsed against four recorded.

No clause changes. PART 11 §2 already requires the escape, because omitting it changes the tree.

## Engine drift

None left. markup-carve/carve-js#1809, markup-carve/carve-rs#1728 and markup-carve/carve-php#2081 have merged, and #2076 moved the carve-js pin past the carve-js fix, so the `PIN_LAG` entry and the three converter-drift lines are gone.


dereuromark:
carve-rs: landed in markup-carve/carve-rs#1728 (`38b8a9b`). On that branch `extension-sigil-escape` writes the expected `.crv` bytes, `x \:name[n]{.k} and \:link[n](u)`, with an empty report.

dereuromark:
carve-js main now produces `extension-sigil-escape` from this branch (markup-carve/carve-js#1809): the written Carve, the empty report and the AST all match.

dereuromark:
carve-php: markup-carve/carve-php#2081 merged (`0699fd7`), fixing markup-carve/carve-php#2069. It writes the `extension-sigil-escape` bytes, checked by running the shared fixture test with this branch's fixture directory in place.

PR #2074, opened 2026-09-16
corpus: a backtick an earlier construct used up does not stop a later link

dereuromark:
Pins a parser split that was not in the corpus. Engine tickets: markup-carve/carve-js#1815 and markup-carve/carve-rs#1733.

**This PR is expected to stay red until both engine fixes land.** Engine PRs merge first; this one waits for them and for the maintainer's approval of the two new goldens.

## The split

```carve
Run {*`make*}[the docs](u).
```

Executable grammar and carve-php:

```html
<p>Run <strong><code>make</code></strong><a href="u">the docs</a>.</p>
```

carve-js and carve-rs:

```html
<p>Run <strong><code>make</code></strong>[the docs](u).</p>
```

PART 3 UNCLOSED RUN (#2056) ends the run at the enclosing `X}`, and the link text scan (CARVE-P3-001) starts at the link's own `[`. Both engines build one bracket table over the whole paragraph and stop it at the first backtick with no closer, so no later `[` gets a match.

## Rows added

- `12-inline-code-8`: a link, image, span and spaced link after `{* *}`, `{_ _}`, `{= =}` and `{+ +}` closers that end an unclosed run.
- `12-inline-code-9`: a link after a backtick in an attribute value, in a `{% %}` comment and in two link destinations. carve-js gets the comment and destination paragraphs right on their own, but reads the destination one literally once the document holds a comment.

Each paragraph holds one run. Two single-backtick runs in one paragraph pair up as a closed code span in the grammar but not in any engine; that is a separate split, left out of these rows.

## Bookkeeping

- `engine-pin-drift.txt` declares both rows against the pinned carve-js, so `engine:report --check` passes. `declarations:check` in release mode lists that ledger as owed, which is the expected red.
- The import baseline moves by the two documents. `-8` round-trips through HTML as the literal text the pin reads; `-9` does not, because its comment renders nothing. Neither round-trips through Markdown, where the writer escapes the brackets.
- Both rows are on the methodology page's "added since this run" list and routed to a docs page.


dereuromark:
carve-js main now renders `12-inline-code-8` and `12-inline-code-9` from this branch as expected (markup-carve/carve-js#1816).

